PR #5, opened 2018-10-25
fix_bug_rgb_input

lm-jira:
Change the code to deal with RGB input instead of BGR input.

### DIFF
--- a/output_template/python/usb_camera_demo.py
+++ b/output_template/python/usb_camera_demo.py
@@ -38,8 +38,6 @@ from lmnet.utils.demo import (
     add_fps,
 )
 
-import matplotlib.pyplot as plt
-
 nn = None
 pre_process = None
 post_process = None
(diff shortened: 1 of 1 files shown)


hadusam:
ok to test

ruimashita:
@lm-jira I think the bug doesn't be fixed by this PR.
We should change input of `run_inference()` to RGB.

lm-jira:
@ruimashita Thank you for your comment
I changed the input image from BGR to RGB in run_inference in this pull request.
However, the result is not improved as I expected because it misclassifies my faces as background.
The accuracy is still good in airplanes and motorbikes classification.

ruimashita:
@lm-jira Thanks, LGTM. I will verify to work.

ruimashita:
I verified that the script work well with my face detection model,
Also when the camera capture blue image, the `img[:, :, 0]` values of after [this line](https://github.com/blue-oil/blueoil/pull/5/files#diff-9ca1efa1ec38e2e985c8ae39464b3182R66) are close `0`.

lm-jira:
Thank you for the verification.

PR #8, opened 2018-10-30
add max_steps option in blueoil config file

lm-jira:
Allow config file to specify the number training (max steps) instead of the number of training epochs (max epochs). 
* Config file must contain either of max steps or max epochs
* max steps option should be written as follow. (insider trainer items of config file, where it is set to 100 in the example)
  `steps: 100`


ruimashita:
run test

hadusam:
ok to test

PR #9, opened 2018-10-31
runtime image resize implementation

yoya:
blueoil::data_processor::Resize implementation.
and Tensor enhancement as class, utility for image pixel processing.


runtime/include/blueoil.hpp
```diff
@@ -23,9 +23,17 @@ extern "C" {
 
 
 namespace blueoil {
-struct Tensor {
+class Tensor {
+public:
   std::vector<float> data;
   std::vector<int> shape;
+  int elems();
+  static Tensor zeros(std::vector<int> shape);
```
n-nez: why it is not a constructor?

runtime/include/blueoil.hpp
```diff
@@ -23,9 +23,17 @@ extern "C" {
...
+  static Tensor array(std::vector<int> shape, float *data);
```
n-nez: why it is not a constructor?

runtime/include/blueoil_image.hpp
```diff
@@ -0,0 +1,25 @@
+/* Copyright 2018 Leapmind Inc. */
+#ifndef RUNTIME_INCLUDE_BLUEOIL_IMAGE_HPP_
+#define RUNTIME_INCLUDE_BLUEOIL_IMAGE_HPP_
+
+#include "blueoil.hpp"
+
+namespace blueoil {
+namespace image {
+
+float *Tensor_at(Tensor &tensor, const int x, const int y); // return RGB
```
n-nez: Why it is not part of Tensor class itself?

runtime/include/blueoil_image.hpp
```diff
@@ -0,0 +1,25 @@
...
+Tensor Tensor_CHW_to_HWC(Tensor &tensor);
+Tensor Tensor_HWC_to_CHW(Tensor &tensor);
```
n-nez: These too seems to be used only in tests...
- Should them be a part of public API?
- If them have to be a part of public API why them are not part of the Tensor API itself?

runtime/include/blueoil_image.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+enum ResizeFilter {
+		   RESIZE_FILTER_NEAREST_NEIGHBOR = 0,
+		   RESIZE_FILTER_BI_LINEAR
```
n-nez: `RESIZE_FILTER_BI_LINEAR` is unused, do we need it?
do we need this `enum` at all, if we support only one type of resizing at the moment?

runtime/include/blueoil_image.hpp
```diff
@@ -0,0 +1,25 @@
...
+};
+
+Tensor Resize(const Tensor& image, const int width, const int height,
+	      const enum ResizeFilter filter);
```
n-nez: This function seems to be just a part of implementation of
```cpp
Tensor Resize(const Tensor& image, const std::pair<int, int>& size) {
```
which is in another file...

What are the benefits of splitting these functions across two files, when we could have just one file with only `￼Tensor Resize(const Tensor& image, const std::pair<int, int>& size)` exported as public interface instead...

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+	throw "unknown ResizeFilter:"; // XXX
+    }
+    const int srcHeight = tensor.shape[0];
+    const int srcWidth  = tensor.shape[1];
+    const int channels  = tensor.shape[2];
+    const int height = srcHeight;
+    Tensor dstTensor = Tensor::zeros({height, width, channels});
+    float xScale = (float) width / (float) srcWidth;
+    int xSrcWindow = floorf(1/xScale);
+    xSrcWindow = 1; // Nearest
```
n-nez: As we unconditionally assigning 1 here initializing it with `floor(1/xScale)` makes no sense...
How about just initializing the value with 1?
```cpp
int xSrcWindow = 1;
```

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+    //xSrcWindow = (xSrcWindow < 2)? 2 :xSrcWindow; // Bi-Liner
```
n-nez: Let's *remove* unused code, not just comment it out

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+	for (int dstX = 0 ; dstX < width ; dstX++) {
+	    int srcX = (int) floorf(dstX/xScale);
+	    int srcY = dstY;
+	    for (int c = 0 ; c < channels ; c++) {
+		float v = 0.0;
+		float totalW = 0.0;
+		for (int x = -xSrcWindow ; x < xSrcWindow; x++){
+		    float *srcRGB = blueoil::image::Tensor_at(tensor,
+							      srcX + x, srcY);
+		    float d = fabs((float)x / (float)xSrcWindow);
```
n-nez: `std::fabs()` ?

And let's use `C++` casts not `C` ones.

`static_cast<float>` for this case.

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+		    float w = (d<0.5)?1.0:0.0; // NearestNeighbor
+		    // float w = 1.0 - d; // Bi-Linear
```
n-nez: If we do not need it let's remove it, if we need it why it is commented?

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+	    int srcY = (int) floorf(dstY/yScale);
+	    for (int c = 0 ; c < channels ; c++) {
+		float v = 0.0;
+		float totalW = 0.0;
+		for (int y = -ySrcWindow ; y < ySrcWindow ; y++) {
+		    float *srcRGB = blueoil::image::Tensor_at(tensor,
+							      srcX, srcY + y);
+		    float d = fabs((float)y / (float)ySrcWindow);
+		    float w = (d<0.5)?1.0:0.0; // NearestNeighbor
+		    // float w = 1.0 - d; // Bi-Linear
```
n-nez: If we do not need it let's remove it, if we need it why it is commented?

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+    }    
+    const int srcHeight = tensor.shape[0];
+    const int srcWidth  = tensor.shape[1];
+    const int channels  = tensor.shape[2];
+    const int width = srcWidth;
+    Tensor dstTensor = Tensor::zeros({height, width, channels});
+    float yScale = (float) height / (float) srcHeight;
+    int ySrcWindow = floorf(1/yScale);
+    ySrcWindow = 1; // Nearest
+    // ySrcWindow = (ySrcWindow < 2)? 2 :ySrcWindow; // Bi-Linear
```
n-nez: If we do not need it let's remove it, if we need it why it is commented?

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+	    }
+	}
+    }
+    return dstTensor;
+} 
+
+Tensor Resize(const Tensor& image, const int width, const int height,
+	      const enum ResizeFilter filter) {
+    // int height = image.shape[0];
+    // int width  = image.shape[1];
```
n-nez: If we do not need it let's remove it, if we need it why it is commented?

runtime/include/blueoil.hpp
```diff
@@ -23,9 +23,17 @@ extern "C" {
 
 
 namespace blueoil {
-struct Tensor {
+class Tensor {
+public:
   std::vector<float> data;
   std::vector<int> shape;
```
n-nez: Should really `data` and `shape` remain to be directly accessible from outside, after we made Tensor to be a class?

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,111 @@
...
+    if (std::memcmp(&(data[0]), &(tensor.data[0]),
+               this->elems() * sizeof(float))) {
+        return false;
+    }
+    return true;
+}
+
+// all elements nealy equals check.
+bool Tensor::allclose(const Tensor &tensor) {
+    float rtol=1.e-5, atol=1.e-8; // same as numpy isclose
```
n-nez: what are `rtol` and `atol`?

runtime/tests/test_tensor.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "blueoil.hpp"
+
+int test_tensor() {
+    blueoil::Tensor tensor= blueoil::Tensor::zeros({2, 3});
+    tensor.data[0] = 1;
+    tensor.data[1] = 2;
+    tensor.data[2] = 3;
+    tensor.data[3] = 7;
+    tensor.data[4] = 8;
+    tensor.data[5] = 9;
```
n-nez: Why not use `blueoil::Tensor::array()` here as well? :thinking:

runtime/tests/test_tensor.cpp
```diff
@@ -0,0 +1,42 @@
...
+    float expected_data[][3] = {
+				{1, 2, 3},
+				{7, 8, 9}
+    };
+    blueoil::Tensor expect = blueoil::Tensor::array({2, 3},
+						    (float *) expected_data);
+    if ((tensor.allequal(expect) == false) ||
+	(tensor.allclose(expect) == false)) {
```
n-nez: The one thing I see relatively frequently but could not get is comparing floating values against `true` and `false` constants... :thinking: 

Why not just use logical not (`!`) here?

runtime/include/blueoil_image.hpp
yoya: This "at" is a function for an image, as x, y args indicates.
If we include it in the Tensor class itself , It well be difficult interface a bit.

runtime/src/blueoil.cpp
yoya: relative and absolute tolerance  threshhold.
same to NumPy functions API.
https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.isclose.html

runtime/include/blueoil.hpp
yoya: zeros, array imitates the NumPy API.
But, i agree it shall be constructor.

https://github.com/blue-oil/blueoil/pull/9/commits/8f5da4e01b17a27549d1e55ecd4faedc259a7b91
https://github.com/blue-oil/blueoil/pull/9/commits/37fcc6e4ea99cea025187e020fad1040a8fb458d

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+    if (filter != RESIZE_FILTER_NEAREST_NEIGHBOR) {
+	throw "unknown ResizeFilter:"; // XXX
+    }
+    const int srcHeight = tensor.shape[0];
+    const int srcWidth  = tensor.shape[1];
+    const int channels  = tensor.shape[2];
+    const int height = srcHeight;
+    Tensor dstTensor = Tensor::zeros({height, width, channels});
+    float xScale = (float) width / (float) srcWidth;
+    int xSrcWindow = floorf(1/xScale);
```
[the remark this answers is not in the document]
yoya: it's my mistake.
https://github.com/blue-oil/blueoil/pull/9/commits/ab936669ede5f3927f622044c6c29baa23e82360

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,194 @@
...
+    float *pixelPtr = scanlinePtr + clamped_x * channels;
+    return pixelPtr;
+}
+
+Tensor Tensor_CHW_to_HWC(Tensor &tensor) {
+    const int channels  = tensor.shape[0];
+    const int height = tensor.shape[1];
+    const int width  = tensor.shape[2];
+    if ((channels != 1) && (channels != 3)) {
+	throw "wrong channles != 1,3"; // XXX
```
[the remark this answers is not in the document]
yoya: I think so too. assert using, good to this case.
https://github.com/blue-oil/blueoil/pull/9/commits/d7597efdffa8d553ebd8fddd2db230b1d778ea2f

runtime/include/blueoil.hpp
yoya: I've done apply private to shape, data.
https://github.com/blue-oil/blueoil/pull/9/commits/34d72c45774eb357663cb5e4c2e32869e7d1e1ed

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,107 @@
...
+    Tensor_data_dump(&(data[0]), shape);
+}
+
+// all elements exact equals check.
+bool blueoil::Tensor::allequal(const Tensor &tensor) {
+    if (shape != tensor.shape) {
+        return false;
+    }
+    if (std::memcmp(&(data[0]), &(tensor.data[0]),
+               this->elems() * sizeof(float))) {
```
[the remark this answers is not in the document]
yoya: I did not notice. thank you.
https://github.com/blue-oil/blueoil/pull/9/commits/3a75100322688ff804062ae9e89df139736e3e43

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,107 @@
...
+    return std::accumulate(shape.begin(), shape.end(),
+			   1, std::multiplies<int>());
+}
+
+Tensor::Tensor(std::vector<int> shape) {
+    this->shape = shape;
+    std::vector<float> data(this->elems(), 0);
+    this->data = data;
+}
+Tensor::Tensor(std::vector<int> shape, std::vector<float> data) {
```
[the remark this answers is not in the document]
yoya: oh.. I forgot c+11 move semantics.
https://github.com/blue-oil/blueoil/pull/9/commits/741f9705277d8ed54921fb5742c34f2412ab2d97
https://github.com/blue-oil/blueoil/pull/9/commits/37fcc6e4ea99cea025187e020fad1040a8fb458d

runtime/src/blueoil.cpp
```diff
@@ -19,6 +19,9 @@
 namespace blueoil {
 
 int Tensor::shapeVolume() {
+    return this->shapeVolume(shape);
+}
+int Tensor::shapeVolume(std::vector<int> shape) {
```
n-nez: Should it still remain to be a member if it doesn't access tensor state?

Free standing helper could be better for the purpose.

Do we really need pass `shape` by value?

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,107 @@
 
 
 namespace blueoil {
+
+int Tensor::elems() {
+    return std::accumulate(shape.begin(), shape.end(),
+			   1, std::multiplies<int>());
+}
```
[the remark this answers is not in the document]
n-nez: I believe that use cases are different. But `data.size() == shapeVolume()` still seems to be invariant of the class...
So I wonder if it is not better to implement one operation in terms of another one... :thinking:

runtime/include/blueoil_image.hpp
n-nez: ok :+1:

runtime/include/blueoil_image.hpp
n-nez: So it is basically helper function. Wonder if here is better name, as `Tensor_at` sounds very generic. While the function expects a tensor of very concrete shape... :thinking:

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,130 @@
...
+      m_data(tensor.m_data) {
+}
+std::vector<int> Tensor::shape() {
+    return m_shape;
+}
+std::vector<int> Tensor::shape() const {
+    return m_shape;
+}
+float* Tensor::data() {
+    return &(m_data[0]);
```
n-nez: `m_data.data()` is more idiomatic way to do it.

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,130 @@
...
+}
+
+static void Tensor_shape_dump(std::vector<int> shape) {
+    std::cout << "shape:";
+    for (auto itr = shape.begin(); itr != shape.end(); ++itr) {
+        std::cout << *itr << " ";
+    }
+    std::cout << std::endl;
+}
+static void Tensor_data_dump(float *data, std::vector<int> shape) {
```
n-nez: Why do you always pass shape by value? :thinking:

runtime/src/blueoil.cpp
yoya: thank you.
https://github.com/blue-oil/blueoil/pull/9/commits/2a66a0f524f6f21f26edf654ba61b72f4d826dfc

yoya: add >> blueoil::dataAt(std::vector<int> indices, bool clamp)
    del << blueoil::image::Tensor_at(blueoil::Tensor, int x, int y)
more generic function. dataAt(indices);
https://github.com/blue-oil/blueoil/pull/9/commits/790cfbabf07fbd2d9ca816df481f06865025dc7a

runtime/src/blueoil.cpp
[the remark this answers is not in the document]
yoya: The constructor process has no m_data, so it's useful to calculate size as its volume by passing an argument "shape".

runtime/src/blueoil.cpp
yoya: By the way, I'm familiar with "AT" in OpenCV, so I used this name. It 's not good.
https://docs.opencv.org/3.4/d3/d63/classcv_1_1Mat.html#ac40425b84f60b39fd35e03814122661f

runtime/include/blueoil_image.hpp
yoya: Currently, public API header is blueoil.hpp only.
Defined by runtime/CMakeLists.txt:file(GLOB PUBLIC_HEADERS "include/blueoil.hpp")
Tensor_CHW_to_HWC, Tensor_HWC_to_CHW in blueoil_image.hpp, so internal API at this point.

runtime/include/blueoil_image.hpp
yoya: I think that it's better to implement image processing like Resize in a general-purpose, and into blueold_image.cpp. After this we will also want to process grayscale, cropping, flipping, equalize etc.
In addition, it's better to gathering the acceptance of processing corresponding to the YAML rule in blueoil_data_processor.cpp.

runtime/src/blueoil_image.cpp
yoya: I removed Its.
https://github.com/blue-oil/blueoil/pull/9/commits/3c9215681f325ce680e20949907ab409337d9d1c#diff-37230609db7212ca72dd52076761c190L172

runtime/src/blueoil.cpp
yoya: It recursively drops one dimension and calls Tensor_data_dump for dump arbitrary dimension data.
argument "shape" dimension diminishes as the recursion level of the call gets deeper.

runtime/src/blueoil.cpp
yoya:
```
just
std::vector<int> shape() const;
should be fine, as here is no way to change state of vector itself
```
yes.
https://github.com/blue-oil/blueoil/pull/9/commits/64066ac3917fb6bcd1b6fe8090f7d81246992cba

runtime/include/blueoil.hpp
yoya: because imitate numpy API. But I changed it to constructor.
https://github.com/blue-oil/blueoil/pull/9/commits/8f5da4e01b17a27549d1e55ecd4faedc259a7b91

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,175 @@
...
+std::vector<int> Tensor::shape() const {
+    return m_shape;
+}
+
+
+float* Tensor::data() {
+    return m_data.data();
+}
+
+float *Tensor::dataAt(std::vector<int> indices, bool clamp) {
```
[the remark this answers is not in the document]
yoya: https://github.com/blue-oil/blueoil/pull/9/commits/760a5e8d47f44ddd8f440c3fcec1eb21aba0c361
I deleted clamp(Indices) argument.
It was intended to the outside frame accessing by the convolution of the image.
But not good idea.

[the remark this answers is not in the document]
yoya: I've just added CTest rules and README description.
https://github.com/blue-oil/blueoil/pull/9/commits/3dafe3666d9cb9a4bdca0381d350f44db5fcdd3d

tkng:
@n-nez most of the problems you mentioned looks fixed now, could you review again and resolve the conversations that you think OK?

runtime/examples/run.cpp
```diff
@@ -23,19 +23,15 @@ limitations under the License.
...
 
 // TODO: delete this func. it is for debug. 
-blueoil::Tensor RandomImage(int num, int height, int width, int channel) {
-    blueoil::Tensor t;
-    const int data_size = channel * width * height * num;
-    vector<float> data(data_size, 0);
-    vector<int> shape = {num, height, width, channel};
+blueoil::Tensor RandomImage(int height, int width, int channel) {
+    blueoil::Tensor t({height, width, channel});
+    vector<float> data = t.data();
```
n-nez: This will create a copy of the vector.
So below loop will update it.
And as you return `t` after all you will still return it with unchanged `data`...

Maybe you were going to do something like:
```cpp
vector<float>& data = t.data();
```
instead...

runtime/include/blueoil_image.hpp
n-nez: The point was if we really need them here if they used only for testing. We could put them into some kind of helper library used for testing only except.

runtime/src/blueoil.cpp
n-nez: 1. What is wrong with `const std::vector<int>&` here?
2. What is wrong with making it just a standalone helper in private namespace like:

```cpp
namespace {

int shapeVolume(const std::vector<int>& shape) {
  return std::accumulate(...);
}

} // namespace

namespace blueoil {

int Tensor::shapeVolume() const {
  return shapeVolume(m_shape);
}

} // namespace blueoil
```

So we will not attach anything which does not access class state as a method.

runtime/src/blueoil.cpp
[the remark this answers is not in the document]
n-nez: Of course, it is useful. It does not mean that the calculation itself have to be a member of the class.

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,172 @@
...
+
+Tensor::Tensor(std::vector<int> shape, std::vector<float> data)
+    : m_shape(std::move(shape)),
+      m_data(std::move(data)) {
+}
+
+Tensor::Tensor(std::vector<int> shape, float *arr)
+    : m_shape(shape),
+      m_data(std::vector<float>(arr,
+				arr + this->shapeVolume(std::move(shape)))) {
```
n-nez: This line seems to be indented while all other code uses spaces...

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,172 @@
...
+std::vector<int> Tensor::shape() const {
+    return m_shape;
+}
+
+std::vector<float> &Tensor::data() {
+  return m_data;
+}
+
+float *Tensor::dataAsArray() {
+  if (this->m_shape.size() == 0) {
```
n-nez: Btw, writing everything as `this->m_xxx` is not C++ style, but rather python one, where you have to reference `self` explicitly.
You already prefixed all members with `m_` so it is already clear which values are members and which are not.

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,172 @@
...
+    throw std::invalid_argument("Tensor have no shape");
+  }
+  return &(m_data[0]);
```
n-nez:
```cpp
return m_data.data();
```
is more idiomatic and clear way to do it.

runtime/src/blueoil.cpp
n-nez: ok

runtime/src/blueoil.cpp
```diff
@@ -15,6 +17,172 @@
...
+        for (int i = 0 ; i < n ; i++) {
+            Tensor_data_dump(data + i*stride, shape2);
+        }
+    }
+}
+
+// dump N-dimentional array
+void Tensor::dump() {
+    Tensor_shape_dump(m_shape);
+    Tensor_data_dump(&(m_data[0]), m_shape);
```
n-nez: `Tensor_data_dump(m_data.data());`

runtime/src/blueoil.cpp
n-nez: looks like obscure abbreviations for me, but if it is widely accepted convention I'm ok with it...

runtime/src/blueoil_image.cpp
```diff
@@ -0,0 +1,182 @@
...
+	}
+    }
+    return dstTensor;
+}
+
+
+/*
+ * Resize Image
+ */
+Tensor ResizeHorizontal(Tensor &tensor, const int width,
```
n-nez: why not `const Tensor& tensor` if you do not change it internally? :thinking:

runtime/test/network_dummy.cpp
```diff
@@ -0,0 +1,11 @@
+#include "blueoil.hpp"
+
+extern "C" { // dummy functions
+    Network *network_create() { return NULL; }
+    int network_get_input_rank(Network *nn) { (void) nn ; return 0; }
+    void network_get_input_shape(Network *nn, int *shape) { (void)nn ; (void)shape; }
+    int network_get_output_rank(Network *nn) { (void)nn ; return 0; }
+    void network_get_output_shape(Network *nn, int *shape) { (void)nn; (void)shape; }
+    bool network_init(Network *nn) { (void) nn ; return true; }
```
n-nez: you can write it as
```cpp
bool network_init(Network *) { return true; }
```

runtime/test/test_tensor.cpp
```diff
@@ -0,0 +1,59 @@
...
+			       {8, 9}
+    };
+    float tensor_data3[][3] = {  // equals shape, different data.
+			       {1, 2, 3},
+			       {7, 0, 9}
+    };
+    blueoil::Tensor tensor0({2, 3}, (float *) tensor_data);
+    blueoil::Tensor tensor1({2, 3}, (float *) tensor_data);
+    blueoil::Tensor tensor2({3, 2}, (float *) tensor_data2);  // shape diff
+    blueoil::Tensor tensor3({2, 3}, (float *) tensor_data3);  // data diff
```
n-nez: could we use C++ casts instead of C ones.
- `reinterpret_cast` for pointers
- `static_cast` for float -> int, int -> float, etc

n-nez:
@tkng wrote some minor new comments. In general looks good to me. :+1:

tkng:
@yoya I really wanted to use C++ pre/post processor, so I made some changes for accelerating developement. Sorry for bothering you, could you merge my commits into your branch? My branch is here: https://github.com/tkng/blueoil/tree/runtime_imageresize

All of my commits don't change the behavior, I just followed the advices above. I believe I could fix all of the review comments above.

@n-nez Thanks, your review comments are really helpful, greate contribution to improve the code quality.

runtime/include/blueoil_image.hpp
yoya: I have moved CHW<=>HWC convert functions to the test folder.
https://github.com/blue-oil/blueoil/pull/9/commits/64cb5beac5a447e8803a6300ed0f4089f165750f

ruimashita:
I verified that unit tests pass 
```
$ mkdir build/
$ cd build/
$ DLK_LIB_DIR=`pwd`/../examples/dlk_lib/ cmake -DBUILD_SHARED_LIBS=ON ../
$ make
$ make test
Running tests...
Test project /home/wakisaka/blueoil/runtime/build
    Start 1: blueoil-test-tensor
1/4 Test #1: blueoil-test-tensor ..............   Passed    0.02 sec
    Start 2: blueoil-test-image
2/4 Test #2: blueoil-test-image ...............   Passed    0.02 sec
    Start 3: blueoil-test-opencv
3/4 Test #3: blueoil-test-opencv ..............   Passed    0.09 sec
    Start 4: blueoil-test-resize
4/4 Test #4: blueoil-test-resize ..............   Passed    0.07 sec

100% tests passed, 0 tests failed out of 4

Total Test time (real) =   0.19 sec
```

iizukak:
@ruimashita Can you try `Update branch` button?
When I try, this button dose not work unknown reason... 🤔

ruimashita:
@iizukak 
Maybe, when the branch is latest the button is disable.

PR #13, opened 2018-11-02
add quantize_first_convolution option in config file

lm-jira:
* add question about quantizing first layer in init
* edit classification model to be able to set `NETWORK.QUANTINZE_FIRST_CONVOLUTION`
* set `NETWORK.QUANTINZE_FIRST_CONVOLUTION` value based on config file
* import `PerImageStandardization` class in the network
* set `pre_processor` to 
      1) `DivideBy255`, if `NETWORK.QUANTINZE_FIRST_CONVOLUTION` is true
      2) `PerImageStandardization`, otherwise


lmnet/configs/core/classification/lmnet_v1_quantize_cifar10.py
```diff
@@ -23,6 +23,7 @@
 from lmnet.pre_processor import (
     Resize,
     DivideBy255,
+    PerImageStandardization,
```
ruimashita: No need this line

lmnet/configs/core/object_detection/lm_fyolo_quantize_pascalvoc_2007_2012.py
```diff
@@ -23,6 +23,7 @@
 from lmnet.pre_processor import (
     ResizeWithGtBoxes,
     DivideBy255,
+    PerImageStandardization,
```
ruimashita: No need this line

lmnet/lmnet/networks/classification/lmnet_v1.py
```diff
@@ -127,6 +127,7 @@ class LmnetV1Quantize(LmnetV1):
 
     def __init__(
             self,
+            quantize_first_convolution=True,
```
ruimashita: This change not effect to classification model quantization or not. 

Please see this line.
https://github.com/blue-oil/blueoil/blob/master/lmnet/lmnet/networks/object_detection/lm_fyolo.py#L217

In future, I hope that to create new Mixin or base class which name is like `Quantizeable` which handle to quantization of all networks. but It takes some times to implement.

Our current stage, I think that step by step development is more productive, In this issue and PR, we only focus to pre process, ignore to handle to weight quantization. How do you think?

lm-jira: As you said, create the `Quantizable` base class would make the maintenance and further implementation easier.

However, what do you mean by ignore weight quantization?
Do you mean we just change the preprocessor based on config file 
and apply weight quantization for all cases?

ruimashita: As last conversation,
This change not effect to classification model quantization and it may takes several days to fix it,
In this issue and PR, don't need change classification model quantization.
To keep PR small changes makes productive, easy to test and review. 👍👍

lm-jira: I already found the reason why the result is similar between two pre-processors.
It's because the value of `quantize_first_convolution`  in config file is `True/False` not `yes/no`.
So the if condition should be `if quantize_first_convolution` instead of `if quantize_first_convolution == "yes"`.
I already corrected the code and waiting for the training result.

hadusam:
add to whitelist

lm-jira:
I merge `first_layer_quantization`, `first_layer` and `quantize_first_convolution` into `quantize_first_convolution`.
I also undo the code about first layer quantization so the model apply quantization to the first layer for every case.
Finally, I fix the bug about selecting the preprocessor and compare the performance between `DivideBy255` and `PerImageStandardization` as shown below.

These are the loss and accuracy for`PerImageStandardization` preprocessor.
[orange line:training, blue line:validation]
loss_perimagestandardization

acc_perimagestandardization

While these are the result of `DivideBy255` preprocessor.
[orange line:training, blue line:validation]
loss_divideby255

acc_divideby255

The result shows that although the loss and accuracy during training is similar for both preprocessors,
those during validation in `PerImageStandardization` is much better than that in `DivideBy255`.
This support the fact suggested in [https://github.com/LeapMind/juiz_dot/issues/701](url) that `DivideBy255` seems to face overfitting problem, while `PerImageStandardization` does not.

ruimashita:
@lm-jira 
Please write final metrics value, I didn't understand the value from the graph images, And Plz write about`dataset`, which `dataset`, how much train and validation(test) size, input image size, etc.

lm-jira:
I added the exact number of the result (loss and accuracy) ,
as well as the detail of the implementation in the comment above.

* Since I did not provide the test data during training, I use the value of the last iteration as the final result.

ruimashita:
@lm-jira Thanks 👍 Great!!
Could you implement unit test ?

lm-jira:
@ruimashita Is the unit test running `./blueoil_test.sh` command?

ruimashita:
@lm-jira 
> Is the unit test running ./blueoil_test.sh command?
Yes.
Also we don't have the unit test of `generate_lmnet_config.py`, But we need it.

lmnet/configs/core/classification/lmnet_v1_quantize_cifar10.py
```diff
@@ -91,6 +91,7 @@
 }
 NETWORK.WEIGHT_QUANTIZER = binary_mean_scaling_quantizer
 NETWORK.WEIGHT_QUANTIZER_KWARGS = {}
+NETWORK.QUANTIZE_FIRST_CONVOLUTION = True
```
ruimashita: Is `QUANTIZE_FIRST_CONVOLUTION` needs ?

lmnet/lmnet/networks/classification/lmnet_v1.py
ruimashita: Is the `quantize_first_convolution` need ?

lm-jira:
@ruimashita Okay. I will implement the unit test and add the result in the report.
Beside, I will check if those two lines of code you mentioned is necessary or not.

For the unit test of `generate_lmnet_config.py`, I will try designing it and discuss with you later.

lmnet/configs/core/classification/lmnet_v1_quantize_cifar10.py
lm-jira: This one is not needed so I'm gonna delete it in the next commit.

lm-jira: The code below is also not necessary.
Thank you so much for pointing it out.

ruimashita:
@lm-jira 
When I run `./blueoil_test.sh`, the error comes.
Also Jenkins CI output the error. 
https://jenkins.blue-oil.org/job/blueoil_main/31/console

Would you fix the error 😉 


```
$  ./blueoil_test.sh
Creating directory for test: ./tmp/tests/20181108160402

# Basic tests
## Test of caltech101_classification
### ./blueoil.sh init
OK!
mv: cannot stat 'config/created_by_test_script_caltech101_classification.yml': No such file or directory
### diff ./tmp/tests/20181108160402/created_by_test_script_caltech101_classification.yml ./tests/config/caltech101_classification.yml
##############################
ERROR result is 2 (expect 0)
##############################
ERROR: No such file : ./tmp/tests/20181108160402/created_by_test_script_caltech101_classification.yml

Usage
./blueoil_test.sh <YML_CONFIG_FILE(optional)>

````

**Comment by lm-jira** [comment]
I already updated the code for unit testing.
Please have a look at it.

**Comment by ruimashita** [comment]
@lm-jira 
When I choose
* `apply quantization at the first layer?` is `no`
* `Task` is `object detection`
Error comes.

**Comment by lm-jira** [comment]
@ruimashita Sorry for overlook checking object detection model.
I already corrected the code and pushed it in the repo.

**Comment by hadusam** [comment]
run dlk test

**Comment by hadusam** [comment]
run dlk test

**Comment by hadusam** [comment]
run dlk test

**Comment by hadusam** [comment]
run dlk test

**Comment by hadusam** [comment]
run dlk test

File: blueoil/templates/lmnet/classification.tpl.py [hunk 5]
```diff
@@ -72,7 +73,11 @@
 
 PRE_PROCESSOR = Sequence([
     Resize(size=IMAGE_SIZE),
+{% if quantize_first_convolution %}
```
ruimashita: It is slight things, you can ignore.

The generated config python file by `python ./blueoil/generate_lmnet_config.py xx.yaml` has empty line like this:
screen shot 2018-11-13 at 19 56 08

`{%-`,  `-%}` will eliminate the empty line.
Plz See the doc of jinja: http://jinja.pocoo.org/docs/2.10/templates/#whitespace-control

blueoil/templates/lmnet/object_detection.tpl.py
```diff
@@ -76,7 +77,11 @@
 
 PRE_PROCESSOR = Sequence([
     ResizeWithGtBoxes(size=IMAGE_SIZE),
+{% if quantize_first_convolution %}
```
ruimashita: Also here, empty line

ruimashita:
Solved 👍 

> @lm-jira
> When I choose
> 
>     * `apply quantization at the first layer?` is `no`
> 
>     * `Task` is `object detection`
> 
> 
> Error comes in train.

ruimashita:
@lm-jira 
Would you add the short description about first layer quantization in this page https://docs.blue-oil.org/usage/init.html ?

hadusam:
run dlk test

blueoil/templates/lmnet/classification.tpl.py
lm-jira: I already solved this issue.
However, using {%- .... -%} also removes the tab space in front of DivideBy255() 
so I just moved all the if clause into one line.

blueoil/templates/lmnet/object_detection.tpl.py
lm-jira: This line is also removed.

lm-jira:
@ruimashita I removed the blank lines from template files.
As well as editing document about the initialization step.

blueoil/templates/lmnet/classification.tpl.py
ruimashita: @lm-jira 
Thanks 😄 

Sorry, `{%-` is not needs.
The code be solved, also. 
```
PRE_PROCESSOR = Sequence([
    ResizeWithGtBoxes(size=IMAGE_SIZE),
    {% if quantize_first_convolution -%}
    DivideBy255()
    {% else -%}
    PerImageStandardization()
    {% endif -%}
])
```

ruimashita:
@lm-jira 
Thanks, LGTM.
But the `[dlk] jenkins test` is failure. Plz ask yamada or antonio.

blueoil/templates/lmnet/classification.tpl.py
lm-jira: Removing the `{%-` still remove the tab space at the beginning of the line as shown below.
screen shot 2018-11-14 at 16 19 41

hadusam:
run dlk test

hadusam:
run dlk test

PR #16, opened 2018-11-09
Quantize first convolution

lm-jira:
Edit classification model to quantize first layer of CNN or not based on config file.


lmnet/lmnet/networks/base_quantize.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(
+            self,
+            activation_quantizer=None,
+            activation_quantizer_kwargs=None,
+            weight_quantizer=None,
+            weight_quantizer_kwargs=None,
+            quantize_first_convolution=None,
+            quantize_last_convolution=None,
+    ):
+
```
ruimashita: Please write docs comments.

ruimashita:
@lm-jira 
Thanks :+1: 

All quantized network class have `_quantized_variable_getter()` , can we delete the redundancy?

lmnet/lmnet/networks/base_quantize.py
lm-jira: Okay, I just added it.

lm-jira:
I finished creating the common `_quantized_variable_getter()` in the BaseQuantize class.

ruimashita:
@lm-jira 
> I finished creating the common _quantized_variable_getter() in the BaseQuantize class.

Thanks 👍 
Some tests failed, Please fix it.

lm-jira:
@ruimashita I already fixed it. Pleas have a look at it.

ruimashita:
@hadusam 
`blueoil` and `lmnet` jenkins not work ?? 
When I click `Details` of `blueoil` and `lmnet`, 404 error comes.

lmnet/lmnet/networks/classification/darknet.py
```diff
@@ -286,76 +287,39 @@ def __init__(
...
             *args,
             **kwargs,
         )
 
-        self.quantize_first_convolution = quantize_first_convolution
-        self.quantize_last_convolution = quantize_last_convolution
-
-        activation_quantizer_kwargs = activation_quantizer_kwargs if not None else {}
-        weight_quantizer_kwargs = weight_quantizer_kwargs if not None else {}
+        BaseQuantize.__init__(
```
ruimashita: We need 2 initialization `Darknet.__init__()`  and `BaseQuantize.__init__()` ? .
Why don't use below
```
super().__init__(
           activation_quantizer,
           activation_quantizer_kwargs,
            weight_quantizer,
            weight_quantizer_kwargs,
            quantize_first_convolution,
            quantize_last_convolution,
            *args,
            **kwargs,
)
```
?

lmnet/lmnet/networks/object_detection/lm_fyolo.py
```diff
@@ -167,74 +168,38 @@ def __init__(
...
-
-                # Apply weight quantize to variable whose last word of name is "kernel".
-                quantized_kernel = weight_quantization(var)
-                tf.summary.histogram("quantized_kernel", quantized_kernel)
-                return quantized_kernel
-
-        return var
-
-    def base(self, images, is_training):
+#    def base(self, images, is_training):
```
ruimashita: Why comment out ?
is it work ?

lm-jira: I think that what the base function in `lm_fyolo` do is 
1) define the `custom_getter` function and 
2) call the superclass' base function

Because the arguments of base function (images and is_training) are not related with the `custom_getter` function at all, 
it would be better to define `custom_getter` only once in the initialization instead of define it every time base function is called
and call the superclass' base function directly.

lmnet/lmnet/networks/object_detection/lm_fyolo.py
```diff
@@ -167,74 +168,38 @@ def __init__(
...
+            weight_quantization=self.weight_quantization,
+            first_layer_name="block_1/",
+            last_layer_name="block_last/",
+            quantize_first_convolution=self.quantize_first_convolution,
+            quantize_last_convolution=self.quantize_last_convolution,
+            use_histogram=True
         )
-        with tf.variable_scope("", custom_getter=custom_getter):
-            return super().base(images, is_training)
+#        with tf.variable_scope("", custom_getter=custom_getter):
```
ruimashita: Why do you comment out?
The network not quantized, I thiknk

lmnet/lmnet/networks/object_detection/lm_fyolo.py
```diff
@@ -167,74 +168,38 @@ def __init__(
...
+        BaseQuantize.__init__(
+            self,
+            activation_quantizer,
+            activation_quantizer_kwargs,
+            weight_quantizer,
+            weight_quantizer_kwargs,
+            quantize_first_convolution,
+            quantize_last_convolution,
+        )
         assert callable(weight_quantizer)
```
ruimashita: Why  `assert callable(weight_quantizer)` and `assert callable(activation_quantizer)` remains in here?

lmnet/lmnet/networks/classification/darknet.py
ruimashita: Please see
https://www.ianlewis.org/en/mixins-and-python

lmnet/lmnet/networks/object_detection/lm_fyolo.py
ruimashita: `LMFYoloQuantize.base()` do

1. define the `custom_getter` function 
2. set `tf.variable_scope` with `custom_getter`  
```
with tf.variable_scope("", custom_getter=custom_getter):
```
3. call the superclass' base function 

Need `2` to quantize network.


> Because the arguments of base function (images and is_training) are not related with the custom_getter function at all,
yes. 

> it would be better to define custom_getter only once in the initialization instead of define it every time base function is called
not: `base` function is also only once called. 
Why do you not define `custom_getter` in `BaseQuantize` ?

And, Why you leave comment out (not deletel) ?

lm-jira: You're right.
Thank you for clarifying the second task. (set `tf.variable_scope` with `custom_getter`)
Unlike classification task, there is not a `base` class for the object detection task.
Accordingly, the base function in this class is necessary.
I'm gonna delete the comment out.

> Why do you not define `custom_getter` in `BaseQuantize` ?
> 

That's because the arguments used to create custom_getter depends on the classes.
e.g. some classes require `quantize_first_convolution`, some require `quantize_last_convolution` 
, while some return `tf.summary.histogram` of the quantized value.

> And, Why you leave comment out (not delete) ?

I'm also sorry for forgetting to delete it.
I planned to delete it after the Jenkins test pass.

lmnet/lmnet/networks/object_detection/lm_fyolo.py
lm-jira: I misunderstood about that there is base class define the `custom_getter` for the `tf.variable_scope` and think there is no need to define it here.
I'm gonna fix it soon.

lmnet/lmnet/networks/object_detection/lm_fyolo.py
lm-jira: Because the original version, some network classes call `assert callable(weight_quantizer)` and `assert callable(activation_quantizer)`, while some classes do not call,
I put these `assert` only in the classes of which original one calls the `assert`.

lm-jira: Should I include 
`assert callable(weight_quantizer)` and `assert callable(activation_quantizer)` 
into the base classes and ignore 
`assert weight_quantizer` and `assert activation_quantizer`?

lmnet/lmnet/networks/classification/darknet.py
lm-jira: The reference tells that if you call a function appears in Mixins and base class (test function), only the function in the leftmost Mixins will be called, doesn't it?

ruimashita: When someone create new network class and also want quantized one,
I hope to provide easy way to quantize, something like this.
```
class Newnet(Base):
    ...
    ...
    ...


class QuantizedNewnet(QuantizeMixin, Newnet):
   pass
```

lm-jira: I'm not sure we can do this.I tried testing with a simple code below.
```
class base1:
    def __init__(self,i):
        print("base1")
        self.i = i

class base2:
    def __init__(self, j):
        print("base2")
        self.j = j

class child(base2, base1):
    pass
```
and found that when I initialize an object of child class, only `base2` is printed out.
which means that the init function of `base1` is not called by default and we have to call it manually.

lm-jira: It seems like there is a way to do that.
I will let you know after testing it.

lmnet/lmnet/networks/object_detection/lm_fyolo.py
lm-jira: I changed the `assert ..` to `assert callable( .. )` in `BaseQuantize` class and remove all `assert callable ( .. )` from all subclasses.

lmnet/lmnet/networks/classification/darknet.py
iizukak: When you use mix-in, I think second parent class should not have any instance variable.

For example, `class Child(P1, P2)`, `P2` class should only contain static methods.

In this PR,  `BaseQuantize` has only static method.
So, `__init__` method can be removed. This understand is wrong?

https://github.com/blue-oil/blueoil/blob/4b6f159bc619f3dae21041b1193c1b8bfdb0469b/lmnet/lmnet/networks/base_quantize.py#L20-L41

After remove `__init__` from `BaseQuantize`, next you can remove `BaseQuantize.__init__`.
Finally, you can use `super().__init__` in `DarknetQuantize`.

lm-jira: > In this PR, `BaseQuantize` has only one static method.
> So, `__init__` method can be removed. This understand is wrong?
> 

The `BaseQuantize` contain one static method as you mentioned.
However, in my opinion, the __init__ method is still necessary because it is used to initialize variables for quantization, e.g. quantization function.

iizukak: Oh. Ok, I missed some code.

Do you talking about [base](https://github.com/blue-oil/blueoil/blob/4b6f159bc619f3dae21041b1193c1b8bfdb0469b/lmnet/lmnet/networks/classification/darknet.py#L314-L325) function using `self.weight_quantization`, `self.quantize_first_convolution`, `self.quantize_last_convolution`?

My new suggestion (psudecode) is here

1: Remove Multiple inheritance `DarknetQuantize(Darknet, BaseQuantize)` -> `DarknetQuantize(Darknet)`
2: Rename class name `BaseQuantize` -> `QuantizeParameters`
3: Add instance variables to `QuantizeParameters`
4: Add getter function to `QuantizeParameters`
5: 

```
self.quantize_params = QuantizeParameters(
    weight_quantization=self.weight_quantization,
    use_histogram=True,
    first_layer_name="block_1/",
    last_layer_name="conv_23/",
    quantize_first_convolution=self.quantize_first_convolution,
    quantize_last_convolution=self.quantize_last_convolution,
)
```

6: Simplify `base` function.

```
def base(self, images, is_training):
    with tf.variable_scope("", custom_getter=self.quantize_params.getter()): 
    return super().base(images, is_training)
```

This idea is little complicated.
Somebody may have more better idea...

lm-jira: Thank you for your suggestion.
If I understand correctly, your code initializes quantization parameter in the `__init__` function of `DarknetQuantize` class, doesn't it?

lmnet/lmnet/networks/quantize_param_init.py
```diff
@@ -0,0 +1,101 @@
...
+        assert callable(weight_quantizer)
+        assert callable(activation_quantizer)
+
+        activation_quantizer_kwargs = activation_quantizer_kwargs if activation_quantizer_kwargs is not None else {}
+        weight_quantizer_kwargs = weight_quantizer_kwargs if weight_quantizer_kwargs is not None else {}
+
+        weight_quantization = weight_quantizer(**weight_quantizer_kwargs)
+        self.activation = activation_quantizer(**activation_quantizer_kwargs)
+
+        self.custom_getter = functools.partial(self._quantized_variable_getter,
```
ruimashita: This class depends on that `self.first_layer_name` and `self.last_layer_name` must be defined at aother place of network class, It looks much tight coupling.
Please implement unit test of the class for the suitable desing of this class.

lm-jira: I implemented the unit test to confirm the definition of `self.first_layer_name` and `self.last_layer_name` as you suggested.

I also add the detail of class in the class document.

ruimashita: Where is the unit test ?

lmnet/lmnet/networks/quantize_param_init.py
```diff
@@ -0,0 +1,101 @@
+import tensorflow as tf
+import functools
+
+
+class QuantizeParamInit:
+    """QuantizeParamInit.
```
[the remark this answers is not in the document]
ruimashita: The current documentation still need more document. If I was new comer, I can't understand how to use `QuantizeParamInit`.

ruimashita:
@lm-jira 
I write some comments, Plz see 👍

lmnet/lmnet/networks/quantize_param_init.py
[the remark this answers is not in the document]
lm-jira: I add more detail about how to use the mixing and the requirement of the base class in doc.
Please check if it is clear enough.

lm-jira:
@ruimashita The unit test for quantized network is done.

lmnet/lmnet/networks/quantize_param_init.py
```diff
@@ -0,0 +1,123 @@
...
+    1) Extend this mixin
+    2) Set the custom_getter variable to `self.custom_getter` if its base class does not set it yet
+       The setting is done by adding the base function defining custom_getter for tensorflow inside the class as follows
+             def base(self, images, is_training):
+                 with tf.variable_scope("", custom_getter=self.custom_getter):
+                     return super().base(images, is_training)
+
+    In the object construction step, users must pass
+       * arguments for the quantization (e.g. activation_quantizer) followed by
+       * arguments of the subtask class (e.g. weight_decay_rate)
```
ruimashita: What is `subtask class` ? need to change.

lmnet/lmnet/networks/quantize_param_init.py
```diff
@@ -0,0 +1,123 @@
+import tensorflow as tf
+import functools
+
+
+class QuantizeParamInit:
+    """QuantizeParamInit is a mixin creating a quantized network class more convenient by providing two functions.
+    1) __init___ : for initializing variables necessary for quantization, such as quantization function
```
ruimashita: These functions `__init___` and `_quantized_variable_getter` is not public variable or function. 🤔🤔🤔 
Please write about public var and func.

lmnet/tests/lmnet_tests/networks_tests/test_quantized_network_arguments.py
```diff
@@ -0,0 +1,45 @@
...
+from lmnet.networks.object_detection.yolo_v2_quantize import YoloV2Quantize
+from lmnet.networks.segmentation.lm_segnet_v0 import LmSegnetV0Quantize
+from lmnet.networks.segmentation.lm_segnet_v1 import LmSegnetV1Quantize
+from lmnet.quantizations import (
+    binary_mean_scaling_quantizer,
+    linear_mid_tread_half_quantizer,
+)
+
+
+def test_required_arguments():
```
ruimashita: If you want to test `required arguments`, Should test that when the required augments are not provided raise error.

lmnet/lmnet/networks/quantize_param_init.py
```diff
@@ -0,0 +1,123 @@
...
+        **kwargs: keyword arguments for the initialization of subtask's network class
+    """
+
+    def __init__(
+            self,
+            activation_quantizer=None,
+            activation_quantizer_kwargs=None,
+            weight_quantizer=None,
+            weight_quantizer_kwargs=None,
+            quantize_first_convolution=None,
```
ruimashita: Why the `quantize_first_convolution` and `quantize_last_convolution` default value is `None` ? Why not boolean ?

lmnet/tests/lmnet_tests/networks_tests/test_quantized_network_arguments.py
```diff
@@ -0,0 +1,45 @@
...
+from lmnet.networks.classification.darknet import DarknetQuantize
+from lmnet.networks.object_detection.lm_fyolo import LMFYoloQuantize
+from lmnet.networks.object_detection.yolo_v2_quantize import YoloV2Quantize
+from lmnet.networks.segmentation.lm_segnet_v0 import LmSegnetV0Quantize
+from lmnet.networks.segmentation.lm_segnet_v1 import LmSegnetV1Quantize
+from lmnet.quantizations import (
+    binary_mean_scaling_quantizer,
+    linear_mid_tread_half_quantizer,
+)
+
```
ruimashita: I hope to implement unit tests using `QuantizeParamInit` mixin class with new network class 
* When the param `quantize_first_convolution=True` and `quantize_last_convolution=True`, the all of network's activations before convolution and weights are quantized.

* When the param `quantize_first_convolution=True` and `quantize_last_convolution=False`, the all of network's activations before convolution and weights are quantized excluding last convolution.

* When the param `quantize_first_convolution=False` and `quantize_last_convolution=True`, the all of network's activations before convolution and weights are quantized excluding first convolution.

ruimashita: Please change test python file name correspond to `QuantizeParamInit`'s file.

lmnet/lmnet/networks/quantize_param_init.py
```diff
@@ -0,0 +1,123 @@
...
+    2) Set the custom_getter variable to `self.custom_getter` if its base class does not set it yet
+       The setting is done by adding the base function defining custom_getter for tensorflow inside the class as follows
+             def base(self, images, is_training):
+                 with tf.variable_scope("", custom_getter=self.custom_getter):
+                     return super().base(images, is_training)
+
+    In the object construction step, users must pass
+       * arguments for the quantization (e.g. activation_quantizer) followed by
+       * arguments of the subtask class (e.g. weight_decay_rate)
+       to the init function, this mixin will call init function of subtask's base class automatically.
```
ruimashita: What is `subtask's base class` ? need to change.

lmnet/lmnet/networks/quantize_param_init.py
lm-jira: I mean the `network class` that doing the subtask(classification, object detection).
I changed `subtask class` to `network class`.

lmnet/lmnet/networks/quantize_param_init.py
lm-jira: `_quantized_variable_getter` is a private function so I will delete the explanation about it.

but I think `__init__` is a public function called via the network class. (e.g. when a network object is generated)
Should I delete explanation about `__init__`?

lmnet/lmnet/networks/quantize_param_init.py
lm-jira: I don't know what is the appropriate default value. (implement or not implement quantize at first layer).
However, before implementing this PR, the first layer is quantized by default 
so if the `quantize_first_convolution` is `None`, the first layer is quantized. 

After you mentioned, setting the default value to `True` is easier to understand and maintenance so I will set the default value to `True`.

lmnet/tests/lmnet_tests/networks_tests/test_quantized_network_arguments.py
lm-jira: I changed file's name to `test_quantize_param_init_arguments.py`

lmnet/lmnet/networks/quantize_param_init.py
lm-jira: I mean parent class of the network class. (e.g. `classification.base` for `lmnet_v0` ).
I changed it to `the parent class of the network class`.

lm-jira:
@ruimashita 
I added the unit test for `quantize_first_convolution` in the Blueoil.
Thanks to the guidance from @iizukak and @nlpng about how to run the unit test locally using `tox`, the code finally passes the Jenkins test.
Please have a look at it.

ruimashita:
@lm-jira Nice 👍 
I will review it !

lmnet/tests/lmnet_tests/networks_tests/test_quantize_param_init.py
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    for model in model_classes:
+        network_kwargs["quantize_first_convolution"] = True
+        network_kwargs["quantize_last_convolution"] = True
+
+        with tf.variable_scope("", reuse=tf.AUTO_REUSE):
+            network1 = model(**network_kwargs)
+
+            base1, graph1 = network1.base(tf.ones([10, 32, 32, 3]), True)
```
ruimashita: I can't approve network class API changes base() returning graph only for unit test :thinking:

lmnet/lmnet/networks/quantize_param_init.py
ruimashita: All of python class have `__init__` , don't need to write describe about what is `__init__`,  I think :+1:

lmnet/tests/lmnet_tests/networks_tests/test_quantize_param_init.py
```diff
@@ -0,0 +1,115 @@
+import tensorflow as tf
```
ruimashita: I wrote this:
> I hope to implement unit tests using `QuantizeParamInit` mixin class with new network class
> 
>     * When the param `quantize_first_convolution=True` and `quantize_last_convolution=True`, the all of network's activations before convolution and weights are quantized.
> 
>     * When the param `quantize_first_convolution=True` and `quantize_last_convolution=False`, the all of network's activations before convolution and weights are quantized excluding last convolution.
> 
>     * When the param `quantize_first_convolution=False` and `quantize_last_convolution=True`, the all of network's activations before convolution and weights are quantized excluding first convolution.

Meaning of `new network class` is to implement simple network class only for test on this file to avoid dependency, to simplify :smile:

ruimashita: If you use `reset_default_graph`,  you can implement test more simple :+1: 
https://github.com/blue-oil/blueoil/blob/acb5a270f201f34fe5a5b27a4b395d9c3a838b27/lmnet/tests/conftest.py#L23

lm-jira: Does this mean implementing the unit test on the `new network class` is enough?
No need to implement it with the existing network, e.g. lmnet.

ruimashita: `test_required_arguments()` is not need. Because first and last layer are already asserted in `QuantizeParamInit`.

My intention was only the new test class, but I think it is good to test LMNetQuantize etc, also.
However, When I implement new quantize network, I might worry about  whether should add the new network to the test. How do you think?

lm-jira: I got your point now.
I think that the test class function should check the new class too, and automatically if possible.
So I will find a way to get all quantized classes automatically from the module and check them.

However, if I cannot find the way, developers of the new class must import that new class in the test manually.
I will update later.

lm-jira: As far as I know, fixture is used to setup the test automatically.
The smallest scope that fixture can be called is function [REF](https://docs.pytest.org/en/latest/fixture.html), which means it can be called before function is called.

However, I think `tf.reset_default_graph()` must be called every time before we create the network so I think using `scope=function` for fixture may not be enough in this test?

ruimashita: you can divide test function.
Ideally, test function should do one test.

lmnet/tests/lmnet_tests/networks_tests/test_quantize_param_init.py
lm-jira: Fixed

lmnet/tests/lmnet_tests/networks_tests/test_quantize_param_init.py
lm-jira: I divided each test case into different functions and use the `reset_default_graph` fixture.

PR #35, opened 2018-11-21
improve performance with simple change(not to reallocate memory)

lm-konda:
To improve performance, modify template and not to reallocate memory.

### DIFF
--- a/dlk/python/dlk/templates/src/func/conv2d.cpp
+++ b/dlk/python/dlk/templates/src/func/conv2d.cpp
@@ -54,7 +54,7 @@ void conv3x3_kn2row(T input[],
   assert(p.input_height > 0);
   assert(p.input_width > 0);
 
-  U *buf = new U[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER]();
+  static T_FLOAT buf[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER];
 
   Measurement::Stop();
 
@@ -65,7 +65,6 @@ void conv3x3_kn2row(T input[],
 
   dlk::matrix_multiplication(kernels_, input_, buf_);
   dlk::matrix_shift_add(buf_, output_, p);
-  delete[] buf;
 
   Measurement::Stop();
 }
--- a/dlk/python/dlk/templates/src/func/conv2d.cpp
+++ b/dlk/python/dlk/templates/src/func/conv2d.cpp
@@ -54,7 +54,7 @@ void conv3x3_kn2row(T input[],
   assert(p.input_height > 0);
   assert(p.input_width > 0);
 
-  static T_FLOAT buf[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER];
+  static U buf[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER];
 
   Measurement::Stop();
 


tkng:
Thanks!

Could you share the benchmark result of before/after applying this patch?

lm-konda:
Here is the result(only important parts).
task type : classification using lmnet

after
```
TotalRunTime,303862,  sum:303.862ms
kn2row,166522,  sum:166.522ms
kn2row-1x1,2089,  sum:2.089ms
kn2row-buf,4,  sum:0.004ms
```
before
```
TotalRunTime,375713,  sum:375.713ms
kn2row,239785,  sum:239.785ms
kn2row-1x1,2079,  sum:2.079ms
kn2row-buf,125558,  sum:125.558ms
```

ruimashita:
awesome :+1:

ruimashita:
Plz review this changes

tkng:
I worried a bit about the change of behavior.

Before:
- each element of `buf` is initialized with `0`

After:
- each element of `buf` is not initialized, since it's statically allocated

According to the code of matrix multiplication, it looks good to me. https://github.com/blue-oil/blueoil/blob/acb5a270f201f34fe5a5b27a4b395d9c3a838b27/dlk/python/dlk/templates/include/matrix/multiplication.h

n-nez:
@tkng was wandering about the same, but it seems we don't use the initial content of the buffer anyway, so it should be safe.

also, wonder what would happen if we just rewrite 
```cpp
U *buf = new U[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER]();
```
as
```cpp
U *buf = new U[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER];
```
intsead of making it static

may it be the case that it is initialization took most of the time? :thinking:

lm-konda:
@n-nez san, I will try 'U *buf = new U[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER];' case, and
report performance today.

lm-konda:
@n-nez  san, here is the result.
Filling with zero costs so much.
And this code reallocate memory at every time, so Performance goes up and down.
```
TotalRunTime,300962,  sum:300.962ms
kn2row,165160,  sum:165.16ms
kn2row-1x1,2086,  sum:2.086ms
kn2row-buf,63,  sum:0.063ms
matrix_multiplication,109015,2078,  sum:111.093ms
```

n-nez:
@lm-konda Thank you. Was wondering how much initialization costs us.

Do not allocate memory dynamically is better anyway, so keeping it static is good to me.

dlk/python/dlk/templates/src/func/conv2d.cpp
```diff
@@ -54,7 +54,7 @@ void conv3x3_kn2row(T input[],
   assert(p.input_height > 0);
   assert(p.input_width > 0);
 
-  U *buf = new U[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER]();
+  static T_FLOAT buf[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER];
```
n-nez: could we make it be
```cpp
static U buf[MAX_SIZE_KN2ROW_BUFFER_PER_LAYER]
```
?

as it was supposed to be a template function...

or making everything to be `T_FLOAT` instead of `U` for consistency is also a way to go, as we only use the functions in this file, thus they have not to be templated anyway.

lm-konda: I see. I will try it too. Wait a moment please.

PR #36, opened 2018-11-22
apply loop fusion

lm-konda:
apply loop fusion and improve performance slightly.


ruimashita:
The changes only delete the redundant loop `for(unsigned i = 0; i < in_channel; i++)` and unnecessary initialization `mean[i] = 0;`. 
LGTM.

Is it OK to merge ?

dlk/python/dlk/templates/include/quantizer.h
```diff
@@ -53,25 +53,17 @@ void func_QTZ_binary_channel_wise_mean_scaling(
...
-  {
-    sum[i] += std::abs(input[i * num_elems_in_channel + j]);
-  }
-
-  for(unsigned i = 0; i < in_channel; i++)
-  {
+  for(unsigned i = 0; i < in_channel; i++) {
+    for(unsigned j = 0; j < num_elems_in_channel; j++) {
+      sum[i] += std::abs(input[i * num_elems_in_channel + j]);
+    }
```
n-nez: does `sum` still has to be array now?

it seems we could just rewrite it like
```cpp
for(unsigned i = 0; i < in_channel; i++) {
  T_FLOAT sum = 0;
  for(unsigned j = 0; j < num_elems_in_channel; j++) {
    sum += std::abs(input[i * num_elems_in_channel + j]);
  }

  mean[i] = sum / num_elems_in_channel;
...
```
without any semantic loss, as `sum` is not used anywhere else... :thinking:

lm-konda:
It was an old version. So, I modified it and removed arrays.
Local validation test has passed.
Please review this.

dlk/python/dlk/templates/include/quantizer.h
lm-konda: Thanks to you, we can remove two array. Thank you n-nez san.

dlk/python/dlk/templates/include/quantizer.h
```diff
@@ -48,21 +48,16 @@ void func_QTZ_binary_channel_wise_mean_scaling(
...
-    }
-    mean[i] = sum[i] / num_elems_in_channel;
-    for(unsigned j = 0; j < num_elems_in_channel; j++) {
-      unsigned in_index = i * num_elems_in_channel + j;
-      output[in_index] = (input[in_index] >= 0) ? mean[i] : -1 * mean[i];
+      sum += std::abs(input[i * num_elems_in_channel + j]);
+      for(unsigned j = 0; j < num_elems_in_channel; j++) {
+	unsigned in_index = i * num_elems_in_channel + j;
+	output[in_index] = (input[in_index] >= 0) ? sum / num_elems_in_channel : -1 * sum / num_elems_in_channel;
+      }
```
n-nez: are you sure you want to have two nested

~~~cpp

for(unsigned j = 0; j < num_elems_in_channel; j++) {
...
    for(unsigned j = 0; j < num_elems_in_channel; j++) {
        unsigned in_index = i * num_elems_in_channel + j;
        output[in_index] = ...
~~~
which will just overwrite each `output[in_index]` `num_elems_in_channel` times keeping only last value? :thinking: 

how about something like below instead?
```cpp
T_FLOAT sum = 0;
for(unsigned j = 0; j < num_elems_in_channel; j++) {
    sum += std::abs(input[i * num_elems_in_channel + j]);
}
T_FLOAT mean = sum / num_elements_in_channel;
for(unsigned j = 0; j < num_elems_in_channel; j++) {
    unsigned in_index = i * num_elems_in_channel + j;
    output[in_index] = (input[in_index] >= 0) ? mean : -1 * mean;
}
```
which is more straight forward transformation of initial code...

lm-konda: You are right. This code is more simple. Thanks!

n-nez:
@antonionevado could you look through this PR too?

it seems just my approval is still not enough to merge... :sweat_drops:

ruimashita:
LGTM.

@n-nez @antonionevado @lm-konda 
May I think we don't have the unit test of `func_QTZ_binary_channel_wise_mean_scaling()`, is it right?
And, How to do manually test of this ?

n-nez:
@ruimashita, I was under impression what we are using the function when translating some of the test graphs from dlk.

But after discussion with @antonionevado it seems what usually we do not use this type of quantization for activations, but only for weights. And quantized weights are computed by the compiler in advance, so this one is probably never called...

So we question do we really need this function at all.
If we could imagine a network which applies `func_QTZ_binary_channel_wise_mean_scaling` to activation results and want to support it, we probably should probably prepare cpp unit test for it.

Otherwise, we could maybe remove the function and forget about its existence.

Merging this PR, if we not sure, will probably not hurt as well, as there is nothing which could be broken by it at the moment, and at least the implementation became a bit more readable and (hopefully) performant than it was before. :slightly_smiling_face:

ruimashita:
@n-nez 
Thanks. 
Someday, we might have a chance to use 1bit activation, you know. But if we use 1bit activation, probably we don't use the  `func_QTZ_binary_channel_wise_mean_scaling` as activation.
Because all of these famous SOTA binary network don't use `func_QTZ_binary_channel_wise_mean_scaling` as activation.
* [BNN](https://arxiv.org/pdf/1602.02830.pdf)
* [XNOR](https://arxiv.org/pdf/1603.05279.pdf)
* [BNN+](https://openreview.net/pdf?id=SJfHg2A5tQ)
So, I can't be 100% convinced, we will not need it, I think.

PR #37, opened 2018-11-22
optimize calculation include lambda

lm-konda:
promote optimization slightly not to capture many arguments.

### DIFF
--- a/dlk/python/dlk/templates/src/matrix/shift_add.cpp
+++ b/dlk/python/dlk/templates/src/matrix/shift_add.cpp
@@ -64,36 +64,37 @@ void matrix_shift_add(MatrixView<float, MatrixOrder::ColMajor>& buf,
 
   std::vector<std::thread> threads;
   for (unsigned int begin = 0; begin < (h * w); begin += chunk_size) {
-    threads.emplace_back(std::thread([buf, begin, chunk_size, h, w, p, &result] {
+    threads.emplace_back(std::thread([buf, begin, chunk_size, p, &result] {
           const int kh = p.kernel_height;
-          const int kw = p.kernel_width;
-          const int oc = p.output_channels;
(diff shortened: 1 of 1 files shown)


dlk/python/dlk/templates/src/matrix/shift_add.cpp
```diff
@@ -170,11 +171,10 @@ void matrix_shift_add(MatrixView<int32_t, MatrixOrder::ColMajor>& buf,
...
-  }
-
-  for (auto& th: threads) {
-    if (th.joinable())
-      th.join();
+    for (auto& th: threads) {
+      if (th.joinable())
+	th.join();
+    }
   }
```
ruimashita: The change looks wrong, because exec `th.join()` in each `for (unsigned int begin = 0; begin < h * w; begin += chunk_size)` loop.

```
for (unsigned int begin = 0; begin < h * w; begin += chunk_size)

...
    for (auto& th: threads) {
      if (th.joinable())
	th.join();
    }
}
```

Is it OK ?

ruimashita:
Why you remain the `h` and `w` in int32_t's  matrix_shift_add ?

https://github.com/blue-oil/blueoil/blob/2a58b346f4b0803ed1b26515e13cda4c1dc64503/dlk/python/dlk/templates/src/matrix/shift_add.cpp#L144

ruimashita:
Also there are many white space changes. I'm not sure it is based on our code style.

Plz review

n-nez:
@ruimashita 

I think we need reformat everything to new style with `clang-format` first.
As now almost everything is inconsistent anyway, so maybe we could not worry much about it for now.

Preserving original white space in this PR would make it easier to review, though.

lm-konda:
close this because It has some bugs

PR #43, opened 2018-11-27
remove unnecessary old test

hadusam:
## Purpose
Current tests of dlk part are including old tests of arm part by using old x-tool's compiler instead of system's cross-compiler.
This tests are designed to use qemu for testing arm binaries, but we already have local FPGA and tests of arm binaries is running on FPGA.
And on docker, the test will fail because it's setting (ex. absolute path) is specified for local jenkins.
So, we should remove this unnecesary old tests.


n-nez:
Wonder should not we also remove `dlk/python/dlk/templates/mains/test_arm_main.cpp` as well?
As it was used only by this test. :thinking:

hadusam:
@n-nez 
Thank you. Yes, we should also remove `dlk/python/dlk/templates/mains/test_arm_main.cpp` because it is only used by `dlk/tests/test_arm_specific_code.py`.
I've removed it, please review 😆

PR #44, opened 2018-11-27
Optimize qtz linear

lm-konda:
It is a place where the performance shakes,
but it took more than 5 milliseconds to get out of the 4 millisecond range.

example.

before
```
QTZ_linear_mid_tread_half,5366,  sum:5.366ms
```
after
```
QTZ_linear_mid_tread_half,4876,  sum:4.876ms
```


lm-konda:
may cause overrun too. I will fix it.

lm-konda:
fixed.

lm-konda:
still may overrun, but I confirmed to be able to fix it.

lm-konda:
I withdraw this temporarily.

PR #45, opened 2018-11-27
improve batch normalization

lm-konda:
software pipelining without epilogue.


lm-konda:
still may overrun, but I confirmed to be able to fix it.

lm-konda:
I withdraw this temporarily.

PR #46, opened 2018-11-27
remove redundant initilization

lm-konda:
need not to initialize with 0, because shift_add fill all elements.

and kernels_hwoi is filled in ohwi_to_hwoi.


lm-konda:
performance.

before
```
TotalRunTime,284275,  sum:284.275ms
```
after
```
TotalRunTime,282946,  sum:282.946ms
```

lm-konda:
It's a bug.

PR #47, opened 2018-11-27
improve conv2d

lm-konda:
improve loop optimization.

PR #48, opened 2018-11-28
Optimize qtz linear

lm-konda:


dlk/python/dlk/templates/include/quantizer.h
```diff
@@ -115,33 +115,118 @@ void func_QTZ_linear_mid_tread_half_body(
...
+      tmp = vminq_f32(tmp, max_value_x4);
+      tmp = vmlaq_f32(round_offset, tmp, max_value_rn);
+      int32x4_t r = vcvtq_s32_f32(tmp);
+      output[i]   = r[0];
+      output[i+1] = r[1];
+      output[i+2] = r[2];
+      output[i+3] = r[3];
+    }
+  } else {
+#if 0 // this code is slow(?)
```
n-nez: This fragment is "uncommented".
Let's remove it.

tkng:
(This is a duplicated comment to #47) For this kind of performance improvement, please share the result of benchmark.

Also, plesae include the procedure to reproduce the benchmark result, so that reviewer can check the result.

dlk/python/dlk/templates/include/quantizer.h
```diff
@@ -115,33 +115,118 @@ void func_QTZ_linear_mid_tread_half_body(
...
+	output[i+3+4] = r[3];
+      }
+      tmp = vmaxq_f32(tmp, min_value_x4);
+      tmp = vminq_f32(tmp, max_value_x4);
+      tmp = vmlaq_f32(round_offset, tmp, max_value_rn);
+      r = vcvtq_s32_f32(tmp);
+      output[length]   = r[0];
+      output[length+1] = r[1];
+      output[length+2] = r[2];
+      output[length+3] = r[3];
```
n-nez: Looks like we are writing behind of the buffer end here... Is it correct? :thinking:

lm-konda: It may be a bug.Sorry.

dlk/python/dlk/templates/include/quantizer.h
```diff
@@ -115,33 +115,118 @@ void func_QTZ_linear_mid_tread_half_body(
...
-    output[i+1] = r_tmp[1];
-    output[i+2] = r_tmp[2];
-    output[i+3] = r_tmp[3];
+  const float32x4_t max_value_x4 = vdupq_n_f32(max_value);
+  const float32x4_t min_value_x4 = {0.0f, 0.0f, 0.0f, 0.0f};
+  const float32x4_t round_offset = {0.5f, 0.5f, 0.5f, 0.5f};
+  const float32x4_t max_value_rn = vdupq_n_f32(max_value_r * n);
+
+  const int length = end - begin;
+  if (length <= 8) {
```
n-nez: the current control structure of this function is like this (assuming we removed uncommented code)
```cpp
if (length <= 8) {
  for (; i + 3 < end; i += 4) { // this
  }
} else {
  if (length % 8 == 4) {
    for (; i + 7 < length; i += 8) {
    }
  }
  else {
    for (; i + 3 < end; i += 4) { // that
    }
  }
}
```
which looks a bit too nested, without obvious benefits from it.
futhermore *this* and *that* loops contains absolutely same code...

will not it be easier to read if we rewrite it like:
```cpp
if (length > 8 && length % 8 == 4) {
  for (; i + 7 < length; i += 8) { 
  }
} else {
  for (; i + 3 < end; i += 4) { // this and that
  }
}
```
?

lm-konda: Thank you!
I do not know the reason why the case of mod 8 = 0 is slow, so I feel like leaving it for a while.

lm-konda:
Sorry.
As a result of the measurement again, performance is hardly changed or it is slightly bad, so I will stop PR.

lm-konda:
Probably less loop iteration

PR #49, opened 2018-11-28
Improve shift add

lm-konda:


lm-konda:
performance
before
```
TotalRunTime,302901,  sum:302.901ms
matrix_shift_add_f1,3565,  sum:3.565ms
matrix_shift_add_f2,54829,  sum:54.829ms
```
after
```
TotalRunTime,297029,  sum:297.029ms
matrix_shift_add_f1,3499,  sum:3.499ms
matrix_shift_add_f2,48325,  sum:48.325ms
```

```
network
　LmnetV1Quantize
dataset you used
　Caltech101
training configurattion
　epoch is 2. other is default.
hardware
　DE10-nano
how many times you run the benchmark
　As much as average average value is known (around 10 or more)
```

ruimashita:
@antonionevado @n-nez @nlpng please review 👍

dlk/python/dlk/templates/src/matrix/shift_add.cpp
```diff
@@ -81,16 +80,70 @@ void matrix_shift_add(MatrixView<float, MatrixOrder::ColMajor>& buf,
...
-                float32x4_t b_ = vld1q_f32(b+j);
-                float32x4_t r_ = vld1q_f32(r+j);
-                float32x4_t r__ = vaddq_f32(b_, r_);
-                vst1q_f32(r+j, r__);
-              }
+              if (oc <= 8) {
+                for (; j + 3 < oc; j += 4) {
+                  float32x4_t b_ = vld1q_f32(b+j);
+                  float32x4_t r_ = vld1q_f32(r+j);
+                  float32x4_t r__ = vaddq_f32(b_, r_);
```
n-nez: wonder if we could come up with better variable naming than `r_` and `r__` ... :thinking:

lm-konda: I think so too. Do you have some nice idea?

lm-konda:
OK. I will. Please wait for a while.

dlk/python/dlk/templates/src/matrix/shift_add.cpp
n-nez: In this context it is complicated, but for example
`r`, `b` -> `result_ptr`, `buf_ptr`
`r_`, `b_` -> `result_element`,  `buf_element`
`r__` -> ... it seems that we always store it immediately after assignment, so it may happen that do not create temporary varyable is actually better in this case...

so it becomes something like
```cpp
float32x4_t buffer_element = vld1q_f32(buffer_ptr+j);
float32x4_t result_element = vld1q_f32(result_ptr+j);
vstd1q_f32(result_ptr+j, vaddq_f32(buffer_element, result_element);
```

lm-konda: Is the code correct? I am trying to implement software pipelining.

n-nez: discussed offline. :innocent:

n-nez:
@lm-konda are you going to rename variables as well? :thinking:

dlk/python/dlk/templates/src/matrix/shift_add.cpp
lm-konda: I re-thought about this problem.
`vst1q_f32(result_ptr+j, vaddq_f32(buffer_element, result_element));`
and
`r__ = vaddq_f32(buffer_element, result_element);
vst1q_f32(result_ptr+j, r__);`
are not equivalent in term of instruction scheduling.

But this is a trivial problem and performance deterioration is slight.

n-nez: The reason why they are not equivalent is not obvious...

lm-konda:
OK. I will.

dlk/python/dlk/templates/src/matrix/shift_add.cpp
lm-konda: function(intrinsic) arguments have strong dependency. 
So, I think we have no room to re-schedule instructions.

So we might not be able to hide instruction latency. But a little.

iizukak:

PR #51, opened 2018-11-28
Optimize batch norm

lm-konda:


lm-konda:
I will fix indent.

lm-konda:
Performance slightly changed. So I withdraw this PR.
(Probably less loop iteration.)

Issue #23, opened 2018-11-13
kernel_transform_NHWC_to_HWNoCNi is called every time


tkng:
[kernel_transform_NHWC_to_HWNoCNi](https://github.com/blue-oil/blueoil/blob/acb5a270f201f34fe5a5b27a4b395d9c3a838b27/dlk/python/dlk/templates/src/func/impl/fpga/quantized_conv2d_kn2row.cpp#L30) is called every time before execution of quantized convolution, however, since weights are fixed values, we can skip this computation if we can store the result of transformation to somewhere.

We have several choices to solve the problem.

1. prepare cache for the result of transformation
2. rewrite weight value directly

I think 1. is better than 2., since the best memory layout for convolution kernel depends on the implementation of convolution, we cannot choose one memory layout as of now.

lm-konda:
I tested.
But Kernel transpose NHWC to HWNoCNi accounts for a low percentage of the whole.
Is my understanding wrong?

tkng:
In my understanding, it consumes roughly 6 to 7% of whole execution time.

This benchmark result (ResNet-18 like architecture on FPGA) is quoted from LeapMind internal wiki:

```
Kernel transpose NHWC to HWNoCNi,38,34,37,32,316,99,119,99,1145,445,557,451,4598,1795,1885,1784,  sum:13.434ms
(snip)
TotalRunTime,123006,  sum:123.006ms
```

In this case, `Kernel transpose NHWC to HWNoCNi` consumes 13.4ms of 123ms = 10.8% of whole execution time. I think 10%  is a very good number as a low hanging fruite.

lm-konda:
@lm-jira and I solved this issue. We will send PR after a bit more verification.

PR #52, opened 2018-11-28
remove redundant memory region

lm-konda:
remove unused memory region.

### DIFF
--- a/dlk/python/dlk/templates/src/func/impl/fpga/quantized_conv2d_kn2row.cpp
+++ b/dlk/python/dlk/templates/src/func/impl/fpga/quantized_conv2d_kn2row.cpp
@@ -126,14 +126,9 @@ void QuantizedConv2DKn2Row(QUANTIZED_NOT_PACKED input[],
     const T_UINT k_size = k_n_aligend_with_num_pe * k_h * k_w * k_c_by_word;
 
     T_UINT kernel_hwnocni[k_size];
-    T_UINT kernel_filled_extra[k_size];
-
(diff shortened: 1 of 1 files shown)


dlk/python/dlk/templates/src/func/impl/fpga/quantized_conv2d_kn2row.cpp
```diff
@@ -126,14 +126,9 @@ void QuantizedConv2DKn2Row(QUANTIZED_NOT_PACKED input[],
...
 
     T_UINT kernel_hwnocni[k_size];
-    T_UINT kernel_filled_extra[k_size];
-
-    for (size_t k = 0; k < k_n * k_h * k_w * k_c_by_word; k++) {
-      kernel_filled_extra[k] = kernel[k];
-    }
 
     Measurement::Start("Kernel transpose NHWC to HWNoCNi");
-    kernel_transform_NHWC_to_HWNoCNi(kernel_filled_extra, kernel_hwnocni,
```
n-nez: I guess, the purpose of introducing additional buffer was that `k_n_aligend_with_num_pe` could be theoretically bigger than `k_n`, so it should prevent accessing of memory outside of `kernel[]` array.

At the moment everything is aligned at network architecture level anyway, so we probably will not see any difference. :thinking: 

BTW, Implementing #23 seems to be the best thing to do here, as it will
 - (a) make performance better
 - (b) allow us to keep implementation of `kernel_transform_NHWC_to_HWNoCNi()` as safe and readable as possible 
    - as we call it only once on the start, we do not need care about optimizing it right? :wink:

lm-konda:
withdraw this.

PR #54, opened 2018-11-30
[WIP] ternary weight quantization

lm-jira:
Reproducing ternary weight quantized VGG7 network using https://arxiv.org/pdf/1605.04711.pdf as a reference.
By setting the learning rate as 0.1 as suggested in the paper, the test results are all zero which is not the same as paper's result.

The network class for VGG7 is at `lmnet/lmnet/networks/classification/vgg7.py`.
Some part of codes are changed with the following result.
1) `blueoil/generate_lmnet_config.py` to accept VGG7 network and CIFAR-10 dataset
2) `blueoil/templates/lmnet/classification.tpl.py` to set the training parameters, and
3) `lmnet/executor/train.py` to apply the learning rate decay and monitor the learning rate.

Other modification are parts of another PR.


lm-jira:
The Jenkins test raise a Failure because there is no twn_weight_quantizer  class in core.operators (dlk/python/dlk/core/operators.py).

lmnet/executor/train.py
```diff
@@ -226,7 +226,12 @@ def start_training(config):
     print("max_steps: {}".format(max_steps))
 
     for step in range(last_step, max_steps):
-        print("step", step)
+        num_epoch = step/(train_dataset.num_per_epoch/config.BATCH_SIZE)
+
+        if num_epoch == 80 or num_epoch == 120:
+            model.optimizer_kwargs["learning_rate"] *= 0.1
```
ruimashita: Maybe it is wrong.

lmnet/lmnet/networks/classification/vgg7.py
```diff
@@ -0,0 +1,268 @@
...
+                                #activation=None,
+                                #bias_initializer=biases_initializer,
+                                use_bias=False,
+                                padding=padding,
+                                strides=strides,
+                                data_format=data_format,
+                                kernel_regularizer=tf.contrib.layers.l2_regularizer(scale=0.0001),
+        )
+
+        batch_normed = tf.contrib.layers.batch_norm(conv,
```
ruimashita: Author use caffe, the batch norm eps and decay is different.
http://caffe.berkeleyvision.org/tutorial/layers/batchnorm.html

lmnet/lmnet/networks/classification/vgg7.py
```diff
@@ -0,0 +1,268 @@
...
+                                filters=filters,
+                                kernel_size=kernel_size,
+                                kernel_initializer=kernel_initializer,
+                                #activation=None,
+                                #bias_initializer=biases_initializer,
+                                use_bias=False,
+                                padding=padding,
+                                strides=strides,
+                                data_format=data_format,
+                                kernel_regularizer=tf.contrib.layers.l2_regularizer(scale=0.0001),
```
ruimashita: weight decay is defined at here also, might be deleted.  
https://github.com/blue-oil/blueoil/blob/57d39650988d83684b72897ed2d9ac4cca6e1990/lmnet/lmnet/networks/classification/base.py#L89

lmnet/lmnet/networks/classification/vgg7.py
```diff
@@ -0,0 +1,268 @@
...
+            inputs,
+            filters,
+            activation,
+            *args,
+            **kwargs
+    ):
+        kernel_initializer = tf.contrib.layers.xavier_initializer()
+        biases_initializer = tf.zeros_initializer()
+
+        output = tf.contrib.layers.fully_connected(
```
ruimashita: It looks not be applied weight decay.

lmnet/lmnet/networks/classification/vgg7.py
```diff
@@ -0,0 +1,268 @@
...
+
+        self.conv5 = self.conv_layer("conv5", self.pool2, filters=512, kernel_size=3, is_training=is_training)
+        self.conv6 = self.conv_layer("conv6", self.conv5, filters=512, kernel_size=3, is_training=is_training)
+
+        self.pool3 = self.max_pool("pool3", self.conv6, kernel_size=2, strides=2)
+
+        self.flatten = tf.contrib.layers.flatten(self.pool3)
+
+        self.fc1 = self.fc_layer("fc1", self.flatten, filters=1024, activation=tf.nn.relu)
+        self.fc1_drop = tf.nn.dropout(self.fc1, keep_prob)
```
ruimashita: We can try to no use dropout.

lmnet/lmnet/networks/classification/vgg7.py
```diff
@@ -0,0 +1,268 @@
...
+
+        assert callable(weight_quantization)
+        var = getter(name, *args, **kwargs)
+        with tf.variable_scope(name):
+            # Apply weight quantize to variable whose last word of name is "kernel".
+            if not quantize_first_convolution:
+                if var.op.name.startswith("conv1/"):
+                    return var
+
+            if "kernel" == var.op.name.split("/")[-1]:
```
ruimashita: I'm not sure, fully connect weight are quantized.

lm-jira: I look at .pb file and found that the fully connected layer is not quantized.
screen shot 2018-12-06 at 18 07 03

lmnet/lmnet/networks/classification/vgg7.py
lm-jira: From this code, I think that the weight decay is automatically applied if self.weight_decay_rate is set.
Did I misunderstand?

https://github.com/blue-oil/blueoil/blob/2a04fccb507f6ec7f88337162eee3aa3669e14fc/lmnet/lmnet/networks/classification/base.py#L113

lmnet/lmnet/networks/classification/vgg7.py
lm-jira: I got it and deleted it now.

lm-jira:
@ruimashita Thank you very much for your review. It helps me understand much more about how to check the structure of the network.


I found the best network for learning rate 0.1 and pushed it to the repository.
This network use only one fc-layer (removing the 1024 output fc-layer) and applying L2 regularizer to all convolutional layer.
Apart from the best accuracy, this network is also the most consistent (require less number of training to get a good result).

I also tried removing L2 regularizer from convolutional layer because it's already included in the loss as you mentioned. However, the accuracy drops and I still cannot find the network that gives good result for binarized network.

The accuracy is in the table below.

<table><tr><td rowspan="2">Model setting</td><td colspan="3">Accuracy</td></tr> <tr><td>Full precision</td> <td>Binarize</td> <td>Ternarize</td></tr> <tr><td>lr = 0.01 </td> <td>91.46%</td><td>90.23%</td><td>90.67%</td></tr> <tr><td>lr = 0.1, regularizer </td><td>92.19%</td><td>90.65%</td><td>91.72%</td></tr> <tr><td>lr = 0.1, no regularizer </td><td>91.85%</td><td>-</td><td>90.96%</td></tr>  <tr><td>original paper </td><td>92.88%</tdMo><td>90.18%</td><td>92.56%</td></tr>  </table>

lmnet/lmnet/networks/classification/vgg7.py
ruimashita: Yes you misunderstand.

`weight_decay_loss` come from here
https://github.com/blue-oil/blueoil/blob/57d39650988d83684b72897ed2d9ac4cca6e1990/lmnet/lmnet/networks/classification/base.py#L85

Only the operation name contain `kernel` is applied `weight_decay_rate`.
fc layer's weight name is no `kernel`

ruimashita:
Why do you remove the 1024 output fc-layer ?  To add fc-layer seems better accuracy.

ruimashita:
@lm-jira 
Looks getting close to original paper 👍 
I wrote some comment, And I hope to push the latest code when you ask me.

lmnet/lmnet/networks/classification/vgg7.py
lm-jira: I got it. 
I will add weights_regularizer into the fully_connected layers.

lm-jira:
> @lm-jira
> Looks getting close to original paper 👍
> I wrote some comment, And I hope to push the latest code when you ask me, to avoid misunderstand and mistake .

@ruimashita  Thank you very much for your comments.
I added the epsilon parameter in the batch_norm layer and the weights_regularizer in the fully_connected layer.

> Why do you remove the 1024 output fc-layer ? To add fc-layer seems better accuracy.

At first, I think the same way as you, however, the result is still unnatural (the model return the same output for every input) if the 1024 output fc-layer is added.
This may be adding 1024 output fc-layer makes the model too complex for the dataset which is only 32x32 size.

lm-jira:
@ruimashita 
After trying 2 fc layers (1024 output and num_class output) for several times with different settings, I cannot get a good result.
So I decide to use the model with only one fc layer (removing 1024 fc layer) and this is the finalized code for VGG7 network.
From now on, I'm gonna work with ResNet18.

I also updated the result in the table below.
<table>
<tr><td rowspan="2">Model setting</td><td colspan="3">Accuracy</td></tr> <tr><td>Full precision</td> <td>Binarize</td> <td>Ternarize</td></tr>
 <tr><td>lr = 0.01 </td> <td>91.46%</td><td>90.23%</td><td>90.67%</td></tr>
 <tr><td>lr = 0.1, regularizer </td><td>92.19%</td><td>90.65%</td><td>91.72%</td></tr> 
<tr><td>lr = 0.1, no regularizer </td><td>91.85%</td><td>90.13</td><td>90.96%</td></tr>  
<tr><td>original paper </td><td>92.88%</tdMo><td>90.18%</td><td>92.56%</td></tr>  
</table>

ruimashita:
@lm-jira 
What is updated in the table? Looks only `Binarize` results is updated, but it is not important.

* When trying 2 fc layers, the layers are quantized ? the layers at least 1024 layer should be quantized.
* How effect the normalization changes ?

ruimashita:
@lm-jira 
Which weight initializer is used in original paper?

lm-jira:
> Which weight initializer is used in original paper?

The ternary weight paper doesn't mentioned about the weight initializer.
However, the VGG paper said that 

> It is worth noting that after the paper submission we found that it is possible to initialise the weights without pre-training by using the random initialisation procedure of `Glorot & Bengio (2010)`.

, where the reference paper proposes a Xavier initializer.

lm-jira:
@ruimashita 
After reading code you suggested, I found that the function for ternarize weight is called `ternarize_data`, and there are two types of layers calling it
1) `convolutional layer`

https://github.com/fengfu-chris/caffe-twns/blob/51916a4b661ab4153287ca15f677f96dd11b3374/src/caffe/layers/conv_layer.cpp#L38

2) `inner_product layer`, which I think it is a fully connected layer because there is an inner-product between weight and input in the fully connected layers.

https://github.com/fengfu-chris/caffe-twns/blob/51916a4b661ab4153287ca15f677f96dd11b3374/src/caffe/layers/inner_product_layer.cpp#L98

Because they use a global variable TERNARY to set if the weight must be ternarized or not, 
I think **all fully-connected layer must be ternarized**.

ruimashita:
@lm-jira 
Thanks, I think so 👍 They apply quantization to all layers.

ruimashita:
@lm-jira 
I found 2 point should be change
1. Remove bias from 1st fc layer
1st fc layer have bias operation before batch norm, but it is not need.

2. fc layer's regularizer don't work
If you set `weights_regularizer` parameter, you also need to manually add `ops.GraphKeys.REGULARIZATION_LOSSES` loss.
See also.
https://github.com/tensorflow/tensorflow/blob/r1.4/tensorflow/python/layers/base.py#L333
https://stackoverflow.com/questions/37107223/how-to-add-regularizations-in-tensorflow

lm-jira:
@ruimashita 
After editing the code based on your previous comment, the accuracy is getting much closer to the original paper, and by far the best accuracy we can achieve.
(The 0.00001 epsilon model is running.)

<table>
<tr>
  <td colspan="2">Model setting</td> <td colspan="2">Accuracy</td> <td rowspan="2">Commit number</td>
</tr>
<tr>
  <td> FC-layers </td> <td>batch normalization epsilon</td> <td> Full precision </td> <td> Ternarize </td>
</tr>
<tr>
  <td  colspan="2">original paper </td>  <td> 92.88% </td> <td> 92.56% </td> <td></td>
</tr>
<tr>
  <td rowspan="2"> Quantized both layers </td> <td> 0.00001 (caffe default) </td> <td> - </td> <td>  - </td> <td>e8c167c3b363b059d0b9016b3e15078f8a8f1bb5</td>
</tr>
<tr>
  <td> 0.001 (tf default) </td> <td> 92.85% </td> <td> 91.76% </td> <td>06961900811df3dc2a4897b49fb9c4f98e253903</td>
</tr>
</table>

PR #55, opened 2018-12-03
modify asm for portability

lm-konda:
To enable LLVM compilation, modified .S file with no output's change.


lm-konda:
duplicated #56

PR #56, opened 2018-12-03
Llvm7 support (anyway)

lm-konda:
add llvm compilation support. (may not be used.)
This patch improves performance a lot.

before
```
TotalRunTime,304253,  sum:304.253ms
```
after
```
TotalRunTime,265058,  sum:265.058ms
```


hadusam:
@lm-konda 
Could you describe how did you create or get `omplib.tar.gz` ?
For maintainability, I think we should keep it clear that all software what we use and how to setup them.
So I would like to know how to create that file.
After that, we can consider about creating that file by each builds or setting the static file on our repository.

lm-konda:
OK. This binary was created as below.
https://github.com/blue-oil/blueoil/pull/59/commits/2d8de3b251b3e705f58e95ed46c68a22f3cb70b4

lm-konda:
Duplicated #59

PR #58, opened 2018-12-04
modify tie behaviour of top-k accuracy calculation

joelN123:
The change is in the calculation of classification accuracy. In almost all cases, there are no ties, and the behaviour is no different in those cases. But, in some special cases (where the network gives ties for prediction probabilities) the behaviour is not great. This pull request is to modify that behaviour for those special cases, to be more easy to understand / more easy to spot poor performance in network if something goes wrong.

tf.nn.in_top_k is typically used in tensorflow to calculate top-1,top-3, and top-5 accuracies. It accepts the probabilities of each class (which is predicted by the network), and the 1-hot vector corresponding to the ground truth class.

The tf.nn.in_top_k is 'including ties'. For example, predictions [0.1639,0.1967,0.2131,0.2295,0.1967] and ground truth [false,false,false,false,true] has index 1 and 4 tied with same probability prediction. With tf.nn.in_top_k, top-3 is evaluated as correct because index 4 is tied in third place with value 0.1967.

An especially strange behaviour is when the network outputs all equal probabilities [0.2,0.2,0.2,0.2,0.2] in this case, tf.nn.in_top_k evaluates the network as being correct 100% of the time. The case of [0.2,0.2,0.2,0.2,0.2] could happen in some cases where the network loss etc. has been miss-written, or learning rate is too high, and causes all output probabilites to be equal in some cases (for example). This is not great behaviour, because at a first glance of tensorboard the accuracy could be going up, making it seem like the network is doing well, when it is not.

This is an ongoing issue for Tensorflow, what to do about this https://github.com/tensorflow/tensorflow/issues/10767

A 'fix' for this problem (although maybe should not be called a fix, maybe a change of function behaviour) is to use a combination of tf.nn.top_k and tf.reduce_any, which results in different behaviour.

In the new behaviour, the accuracy is calculated as top k not including ties. In the case of tied predictions, the index order is used to decide the order of the network's class prediction preference.

For example, a ground truth of [true,false,false,false,false] evaluates as correct (top-1) for the predictions [0.2,0.2,0.2,0.2,0.2]. But, it evaluates as (top-1) incorrect for the the ground truth [false,true,false,false,false], and the same predictions. For these cases, both evaluate as correct for top-3.

This could be said as not ideal behaviour, but I think it improves over the previous. If there are many ties, the accuracy will depend on the prevalence of the various classes. For example, if most of the images correspond to the 'first' class, we will again get the behaviour that accuracy will be evaluated as higher than we would expect. If the classes are fairly balanced, I think there will be more resonable behaviour under the new method.

This change was tested by [ PYTHONPATH=. python executor/train.py -c configs/core/classification/lmnet_quantize_cifar10.py -i testing1234 --recreate ] The tensorboard output of the first few thousand steps shows it is running normally.

modified_accuracy_metric_during_training

Also here is a code for testing the difference between previous and modified accuracy calculations. With random label and probability inputs, there is no difference in the calculation output (there are no ties between random float numbers, with high likelihood). But giving manual input that has tying probabilities, it is possible to see the difference in behaviour.

```python
import numpy as np
import tensorflow as tf
# 4 December test for modified behaviour nn.in_top_k
# should match previous method in non-degenerate cases
# so under random labels and probabilities, both should
# give the same output. You can check by using manual input
# that the behaviour is different in special cases where
# there are ties in the predicted probabilities.

# to test batch size 2 and 10 classes:
# labels = tf.Variable([[0,1,0,0,0,0,0,0,0,0],[0,0,0,0,0,0,0,1,0,0]])
# labels = tf.cast(labels, tf.bool)
# # In the actual code, the 'labels_placeholder' gets fed here is tf.bool

# # test_probs = np.asarray([[1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0],[1,2.1,2,4.00001,1.34,1.21,2.02,1.301,2.011,2.53]])
# test_probs = np.asarray([[1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0],[1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0,1.0]])
# test_probs = test_probs / np.expand_dims(test_probs.sum(axis=1), axis=1)
# softmax = tf.Variable(test_probs)
# softmax = tf.cast(softmax, tf.float32)
# In the actual code, the 'output' gets fed here is tf.float32

# variable batch size, random labels, random probabilities
batch_size = 128
my_random_indices = np.random.randint(10, size=batch_size)
labels = tf.one_hot(indices=my_random_indices, depth=10, on_value=True, off_value=False)

my_random_probs = np.random.rand(batch_size, 10)
my_random_probs = my_random_probs/np.expand_dims(my_random_probs.sum(axis=1), axis=1)
softmax = tf.Variable(my_random_probs)
softmax = tf.cast(softmax, tf.float32)

# Here, not the full method is tested. Just the behaviour of the part that changed.
# Self is removed (self is only used for debug option).
# The tf.metrics.mean is removed - it is usually called in a separate update op.


def previous_metrics(softmax, labels):
    """metrics.
    Params:
        softmax: probabilities applied softmax. shape is (batch_num, num_classes)
        labels: onehot labels tensor. shape is (batch_num, num_classes)
    """
    with tf.name_scope("previous_metrics_calc"):
        labels = tf.to_float(labels)

        argmax_labels = tf.argmax(labels, 1)
        current_accuracy = tf.cast(tf.nn.in_top_k(softmax, argmax_labels, k=1), tf.float32)

        current_accuracy_top3 = tf.cast(tf.nn.in_top_k(softmax, argmax_labels, k=3), tf.float32)

        current_accuracy_top5 = tf.cast(tf.nn.in_top_k(softmax, argmax_labels, k=5), tf.float32)

        return tf.stack([current_accuracy, current_accuracy_top3, current_accuracy_top5], axis=-1)


def metrics(softmax, labels):
    """metrics.

    Params:
        softmax: probabilities applied softmax. shape is (batch_num, num_classes)
        labels: onehot labels tensor. shape is (batch_num, num_classes)
    """
    with tf.name_scope("metrics_calc"):
        labels = tf.to_float(labels)

        argmax_labels = tf.cast(tf.argmax(labels, 1), tf.int32)

        _, top_predicted_indices = tf.nn.top_k(softmax, k=1)
        current_accuracy = tf.cast(tf.reduce_any(tf.equal(tf.transpose(top_predicted_indices), argmax_labels), 0), tf.float32)

        _, top_predicted_indices = tf.nn.top_k(softmax, k=3)
        current_accuracy_top3 = tf.cast(tf.reduce_any(tf.equal(tf.transpose(top_predicted_indices), argmax_labels), 0), tf.float32)

        _, top_predicted_indices = tf.nn.top_k(softmax, k=5)
        current_accuracy_top5 = tf.cast(tf.reduce_any(tf.equal(tf.transpose(top_predicted_indices), argmax_labels), 0), tf.float32)

        return tf.stack([current_accuracy, current_accuracy_top3, current_accuracy_top5], axis=-1)

previous_accuracies = previous_metrics(softmax, labels)
accuracies = metrics(softmax, labels)

sess = tf.InteractiveSession()
sess.run(tf.local_variables_initializer())
sess.run(tf.global_variables_initializer())

print("softmax: ", sess.run(softmax))
print("labels: ", sess.run(labels))

previous_output = sess.run(previous_accuracies)
new_output = sess.run(accuracies)

print("previous accuracies: ", previous_output)
print("new accuracies: ", new_output)
print("difference: ", new_output-previous_output)

print("sum of differences: ", abs(new_output-previous_output).sum())
```


tkng:
What a nice pull request, I'm really happy to be free from this problem!

ruimashita:
Jenkins, run to test

ruimashita:
@joelN123 
Thanks a lot. 
This is so so so nice explanation, and manually test :100: :100: :100:

lmnet/lmnet/networks/classification/base.py
```diff
@@ -180,17 +180,21 @@ def metrics(self, softmax, labels):
                 softmax = tf.Print(softmax,
                                    [tf.shape(softmax), tf.argmax(softmax, 1)], message="softmax:", summarize=200)
 
-            argmax_labels = tf.argmax(labels, 1)
+            argmax_labels = tf.cast(tf.argmax(labels, 1), tf.int32)
+
+            _, top_predicted_indices = tf.nn.top_k(softmax, k=1)
             accuracy, accuracy_update = tf.metrics.mean(
-                tf.cast(tf.nn.in_top_k(softmax, argmax_labels, k=1), tf.float32)
+                tf.cast(tf.reduce_any(tf.equal(tf.transpose(top_predicted_indices), argmax_labels), 0), tf.float32)
```
ruimashita: The `tf.transpose` looks strange for me.
* `argmax_labels` shape is [batch_size,] and 
* k=3's `top_predicted_indices` shape is [batch_size, 3]

to apply `tf.equal` with broadcasting, we need to change whether shape.

But In my view point, It is readable that the`batch_size` order of shape is always first.
So, then, `argmax_labels` shape change to [batch_size, 1]
 ```
argmax_labels = tf.cast(tf.argmax(labels, 1), tf.int32)
argmax_labels = tf.expand_dims(argmax_labels, 1)

....
....
....

tf.cast(tf.reduce_any(tf.equal(top_predicted_indices, argmax_labels), axis=1), tf.float32)
```
This is better for me.

**Comment by hadusam** [comment]
ok to test

File: lmnet/lmnet/networks/classification/base.py (hunk 1 above)
Comment (joelN123) [reply lmnet/lmnet/networks/classification/base.py]: yes ! thank you for the tip, that's much nicer to read.

**Comment by lm-jira** [comment]
Just read the detail of Joel's code behavior.

> In the new behaviour, the accuracy is calculated as top k not including ties. In the case of tied predictions, the index order is used to decide the order of the network's class prediction preference.

I just wonder whether it'd be better or not if the accuracy of top-1 should be divided by the number of classes with the same value.

For example, if the result of Softmax is equal for all 5 classes ([0.2, 0.2, 0.2, 0.2, 0.2]), the accuracy is 1/5 = 0.2.

**Comment by joelN123** [comment]
yes, that's true! I guess I thought there might be a problem if we wanted to extract true/false values for top-k of individual input images. But it actually looks like we don't have anything that requires this, so it is no problem. I agree, the behaviour will be better, because it won't depend on the index position of the classes (which is arbitrary).
For the calculation, I'll try putting it in three steps. First is to use tf.nn.in_top_k to get the lowest softmax value within the top_k. Second is to count how many times this value appears. Third is decide : if softmax value at the ground truth index is equal to top_k value, output 1/count. If softmax at ground truth index > top_k value, output 1. If softmax at ground truth index < top_k value, output 0.

**Comment by lm-jira** [comment]
I tried writing a code that calculate the 1/count for the top-1 accuracy as below.
I tested it with using your code and think it works fine.
However, I'm not sure if there is a better way to calculate or not.

```
        argmax_labels = tf.argmax(labels, 1)
        current_accuracy = tf.cast(tf.nn.in_top_k(softmax, argmax_labels, k=1), tf.float32)
        
        max = tf.reduce_max(softmax, axis=-1)
        batch_num, num_classes = labels.get_shape().as_list()
        #count the number of elements with the same value as max
        sum_max = tf.reduce_sum(tf.cast( [tf.equal(softmax[i], max[i]) for i in range(batch_num)] , tf.float32), axis=-1)  
        
        current_accuracy = tf.div(current_accuracy, sum_max)
```

**Comment by joelN123** [comment]
In the test code block, can now just put the modified version.
```python

def metrics(softmax, labels):
    """metrics.

    Params:
        softmax: probabilities applied softmax. shape is (batch_num, num_classes)
        labels: onehot labels tensor. shape is (batch_num, num_classes)
    """
    with tf.name_scope("metrics_calc"):
        labels = tf.to_float(labels)

        argmax_labels = tf.cast(tf.argmax(labels, 1), tf.int32)
        top_5, _ = tf.nn.top_k(softmax, k=5)

        current_accuracy = tf.cast(tf.nn.in_top_k(softmax, argmax_labels, k=1), tf.float32)
        count_border_values = tf.reduce_sum(tf.cast([tf.equal(softmax, tf.expand_dims(top_5[:, 0], 1))], tf.float32), axis=-1)
        current_accuracy = tf.div(current_accuracy, count_border_values)

        current_accuracy_top3 = tf.cast(tf.nn.in_top_k(softmax, argmax_labels, k=3), tf.float32)
        count_border_values = tf.reduce_sum(tf.cast([tf.equal(softmax, tf.expand_dims(top_5[:, 2], 1))], tf.float32), axis=-1)
        current_accuracy_top3 = tf.div(current_accuracy_top3, count_border_values)

        current_accuracy_top5 = tf.cast(tf.nn.in_top_k(softmax, argmax_labels, k=5), tf.float32)
        count_border_values = tf.reduce_sum(tf.cast([tf.equal(softmax, tf.expand_dims(top_5[:, 4], 1))], tf.float32), axis=-1)
        current_accuracy_top5 = tf.div(current_accuracy_top5, count_border_values)

        return tf.stack([current_accuracy, current_accuracy_top3, current_accuracy_top5], axis=-1)
```

joelN123:
Of course, still need to make one more commit for the case when classes less than 5. (or less than 3).

ruimashita:
The current Delta-Lite display `top3` and `top5` accuracy,  I talked with Delta-Lite team the Impact of deleting `top3` and `top5` accuracy. then It not need to care.

ruimashita:
@joelN123 
Thanks, But to divide by number of time make wrong top3 or top5 in some case, please see ⬆️ .
 
To handle the case when classes less than 5. (or less than 3), You can use `self.num_classes` as in condition.

@lm-jira 
Thanks your contribution.

joelN123:
After the most recent change, now if there are 5 classes and all predictions same value, the top-5 is 1, top-3 is 0.6 and top-1 is 0.2.
For me, this makes sense because if we imagine randomly nudging all predictions by a very small value (so they are no longer exactly equal), then top-1 would be evaluated as correct with 1/5 probability. And top-3 would be evaluated as correct with 3/5 probability. And top-5 would be evaluated as correct with 100% probability.

joelN123:
I guess the most important special case is when all predictions are equal. In this case, we want to easily see by top-k evaluations that the network has no idea what the image is.
Let’s say there are ‘c’ classes (and c>5). In the new code, top-1 evaluates as 1/c , top-3 is 3/c , and top-5 is 5/c.
We can compare this to a network that outputs non-degenerate predictions, but is fed white noise instead of the actual image. In that case, (on average), the top-1 is 1/c , top-3 is 3/c , and top-5 is 5/c (just by making totally random predictions).
So in both cases, the top-k evaluations come out as the same. I think this is good behavior for the top-k because uninformed predictions will result in the same top-k evaluation in both cases (on average).

joelN123:
ah, when I say ‘fed white noise’ I mean new white noise for each input. Not repeated over epochs, or the network could learn by memorization.

ruimashita:
@joelN123 @lm-jira 
I am very thankful for the discussion and contribution, trying to make the product better 👍👍👍

joelN123:
Good tips, thanks ! Yes, I think it is only the case of all equal softmax that happened. I think no-one saw the case of just some softmax outputs being equal. I guess the important point is just that the accuracy should appear as small when the 'all equal softmax' happens.

joelN123:
I also put the calculation of topk as a separate function to reduce the code repeat. But actually I'm not sure, maybe it is better to repeat, it is only small amount of code.

ruimashita:
@joelN123 Thanks for your contribution, to effort to be better product and clearly description. 😄 
@lm-jira also thanks 😄

Issue #61, opened 2018-12-05
Update tensorflow version


ruimashita:
Update tensorflow version.
Currently we use so old version of tensorflow `1.4`, we should update version to latest tensorflow.

ruimashita:
@lm-konda says that It is better to create new branch `update-tf-version` in `blueoil/blueoil` repo instead of developer self forked repo, because updateing version needs many member collaborate.

iizukak:
👍

hadusam:
@ruimashita 
It's nice to create a new branch to upgrade tensorflow version! 👍 
To update tensorflow, we need to develop with collaborating on each part of functions on the same branch.

lm-konda:
These may help.

https://hub.docker.com/r/nvidia/cuda/
https://github.com/tensorflow/tensorflow/tags

https://github.com/tensorflow/tensorflow/issues/15604
https://camo.githubusercontent.com/976c7051e504a15d0d50317e7b8e8f6daaa67417/68747470733a2f2f692e737461636b2e696d6775722e636f6d2f784461704a2e706e67

hadusam:
Will we treat `update-tf-version` as develop branch? 🤔 
I think the development flow is following,
1. We fork `update-tf-version` branch and create the PR to merge into `update-tf-version` by each task for updating tensorflow. (may not need to pass tests)
2. After completing the development to upgrade, we will merge that branch into master.

Another selection, we can select directly pushing into `update-tf-version`.
What do you think? @ruimashita

iizukak:
@hadusam 's The first plan seems more natural
I think we should not push to `update-tf-version` branch directory.

n-nez:
agree with @iizukak the first plan at least isolates changes from each other :+1: 

it may break testing of the branch in meantime when it is still half updated, but it seems to be inevitable anyway

tkng:
I'm sorry to turn the table up side down, but I'm using lmnet on TF 1.10 on my local environment, and it works fine for classification problems. Object detection requires a little modification, but it's not too difficult. (My modification doesn't work with TF 1.4, that's why I cannot make a pull request yet.)

Therefore, I think we don't need a "collaborative" branch in this case.

lm-konda:
@tkng The problems occurs when we use blueoil on Docker. Native environment may be OK.

lm-konda:
@n-nez supposing value itself is correct in this case (from tf-1.4).
I will attach two pb files. One is OK and another is NG.

[...]

PR #62, opened 2018-12-06
TensorFlow 1.12.0 support (anyway)

lm-konda:
TensorFlow v1.12.0 support.

requirements
　CUDA 9
　cuDNN 7

change to install tensorflow v1.12.0
change to use metric_variable instead of _create_local

(Dec, 13th add) modify for tensorboard

related to issue #61


lm-konda:
There is no compute_shapes. So only we have to do is rewrite Graph.create_op.

lm-konda:
I wrote like below.
```
TODO:
・not to use compute_shapes
・calling Graph.create_op (from tensorflow.python.framework.ops) with compute_shapes is deprecated, so rewrite them.
```
But, I confirmed that it's an environmetal problem.
`/usr/local/pyenv/versions/python3.6/lib/python3.6/site-packages/tensorflow/python/framework/function.py`
We don't have to care about these cases.

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:8.0-cudnn6-devel
+FROM nvidia/cuda:9.0-cudnn7-devel
```
iizukak: Why do you use `9.0` ? 
`9.2` looks available.
https://hub.docker.com/r/nvidia/cuda/

lmnet/lmnet/metrics/mean_average_precision.py
```diff
@@ -104,10 +104,10 @@ def _streaming_tp_fp_array(
             scores = tf.boolean_mask(scores, mask)
 
         # Local variables accumlating information over batches.
-        tp_value = metrics_impl._create_local('tp_value', shape=[0, ], dtype=tf.bool, validate_shape=False)
-        fp_value = metrics_impl._create_local('fp_value', shape=[0, ], dtype=tf.bool, validate_shape=False)
-        scores_value = metrics_impl._create_local('scores_value', shape=[0, ], validate_shape=False)
-        num_gt_boxes_value = metrics_impl._create_local('num_gt_boxes_value', shape=[], dtype=tf.int64)
+        tp_value = metrics_impl.metric_variable(name='tp_value', shape=[0, ], dtype=tf.bool, validate_shape=False)
```
iizukak: You  changed `_create_local` to `metric_variable` because of TensorFlow 1.12 dose not have `metrics_impl._create_local`.

Are there any other reasons?

docker/Dockerfile
lm-konda: For compatibility. If you want to know in details, please refer to below.
https://github.com/tensorflow/tensorflow/issues/15604

lmnet/lmnet/metrics/mean_average_precision.py
lm-konda: There is no other reason.

iizukak:
After rebuild Docker image, TensorFlow version looks nice 👍 
```
:/home/blueoil# pip list | grep tensorflow
tensorflow               1.12.0   
```

lm-konda:
I will check tensorboard works or not.

lm-konda:
> I will check tensorboard works or not.

At first classification result.
screenshot from 2018-12-06 13-43-11

docker/Dockerfile
iizukak: ok.

lmnet/lmnet/metrics/mean_average_precision.py
iizukak: ok thanks.

lm-konda:
@iizukak Many thanks!

iizukak:
@lm-konda 
Jenkins failed for python coding style.
https://jenkins.blue-oil.org/job/blueoil_lmnet/192/console

lm-konda:
jenkins error. Does it depends on environment?
ImportError: libcublas.so.9.0: cannot open shared object file: No such file or directory

lm-konda:
object detection's result with small epoch number.
It looks fine.
screenshot from 2018-12-06 17-30-49

hadusam:
@lm-konda 
`ImportError: libcublas.so.9.0: cannot open shared object file: No such file or directory`
We need to change base docker image of dockerfiles for lmnet as same as blueoil's (`nvidia/cuda:9.0-cudnn7-devel`).
https://github.com/blue-oil/blueoil/blob/master/lmnet/dockerfiles/gpu.Dockerfile#L1
https://github.com/blue-oil/blueoil/blob/master/lmnet/dockerfiles/multi-gpu.Dockerfile#L1

Currently, lmnet tests are run on docker created by above dockerfile, so please change them.

lm-konda:
@hadusam OK. I will modify base docker image.

ruimashita:
@tkng 
This changes looks good to me, But I think we have to system level test before merge this PR.
I want to merge this PR in this Q, I suppose the test flow is below.

1. Create system level test plan document (something like design doc). The doc describe schedule and member and test scenarios and test environment etc.
scenarios e.g.  
    * Can train classification quantized LMNetV1 with Cifar10, archive 90%.  Can convert the models, and prediction on DE10nano with Cifar10 test dataset archive the same accuracy as tensorflow.
   * Can train object detection ...
   * Can convert old checkpoints files to .so and prediction archive  the same accuracy as tensorflow on x86, arm, DE10-nano.
   * Can resume training with old checkpoints files.

2. Review the test plan. Reviewer is you and blueoil and Delta-lite team members.
3. Following the test plan, Write command level detail specification.
4. Do test following the detail specification.
5. When all of tests pass, we can merge this PR.

iizukak:
Most recent TensorFlow release is 1.13.
We want to use 1.13 not 1.12.

Issue #64, opened 2018-12-07
Enable code editing without docker rebuild


iizukak:
When I edit some python code in `lmnet` or `dlk`, docker image should be rebuild.
It's not comfortable.

I want to mount `lmnet` and `dlk` using `-v` when running docker.
https://github.com/blue-oil/blueoil/blob/master/blueoil.sh#L130-L137

iizukak:
@tkng @ruimashita How do you think about this issue?

ruimashita:
For development, It is so nice.
There are possibility someday we will provide docker image, so then,
I think, It is better to add option to `blueoil.sh` somethink like `-mount-volume` to enable mount `lmnet` and `dlk`.

tkng:
I also think that this feature would be very nice.

Considering the possibility of "official docker image", I wonder if we can mount `lmnet` and `dlk` directory automatically. Does it sounds like a nice idea?

iizukak:
I implemented tkng-san's way for the moment.

PR #67, opened 2018-12-10
[WIP] Mount lmnet and blueoil directory at runtime of Docker container

iizukak:
Issue : #64 

## Problem

As described in #64 , when I change source code in `lmnet`, I should rebuild docker image.
This cause waste of time.

## Solution

If `lmnet` and `blueoil` directories are exist, mount these directories at docker runtime.
Some Blueoil user may not have these directories.
So, I added checking directories exist or not.


iizukak:
run test

hadusam:
The error: `ModuleNotFoundError: No module named 'lmnet.utils'` is occured.
```
#################################################
Erorr log : ./tmp/tests/20181210055628/test_1.log
#################################################
### ./blueoil.sh train ./tmp/tests/20181210055628/created_by_test_script_caltech101_classification.yml
Directory does not exist, creating ./saved
#### Run training (created_by_test_script_caltech101_classification_20181210055636) ####
/usr/local/pyenv/versions/python3.6/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: compiletime version 3.5 of module 'tensorflow.python.framework.fast_tensor_util' does not match runtime version 3.6
  return f(*args, **kwds)
  File "blueoil/blueoil_train.py", line 19, in <module>
    from executor.train import run as run_train
  File "/home/blueoil/lmnet/executor/train.py", line 23, in <module>
    from lmnet.utils import executor, module_loader, config as config_util
ModuleNotFoundError: No module named 'lmnet.utils'
ERROR: Training exited with a non-zero status
```
In this situation, lmnet module is refered from `/home/blueoil/dlk/python/dlk2/lmnet` because there is `__init__.py`. https://github.com/blue-oil/blueoil/blob/master/dlk/python/dlk/lmnet/__init__.py
We are now defining which lmnet module we use by adding `RUN touch lmnet/lmnet/__init__.py` on [dockerfile](https://github.com/blue-oil/blueoil/blob/master/docker/Dockerfile#L133), but your adding option is overriding docker image's directory by host directory without `__init__.py`
We are now using python3, so we don't need `__init__.py` each directory, but there are some `__init__.py` in dlk directory.
This import directory structure is too complicated and this was depended from submodule structure. So sorry about this😢But lmnet and dlk are already integrated, so we can fix this now!
Probably, removing the file `dlk/python/dlk/lmnet/__init__.py` will fix this error, could you try to do? @iizukak

iizukak:
Oh, dlk need his own `libpacker.so`...
```
OSError: /home/blueoil/dlk/build/libpacker.so: cannot open shared object file: No such file or directory
```

hadusam:
> Oh, this time, dlk needs his own libpacker.so...

@iizukak 
Oh, this is because there is no installation process for dlk.
To fix it, we need to add run `python setup.py install` on dlk directory before running convert.
So we should add this process to use local mounting directory of dlk instead of docker's directory.

tkng:
Hmm, I think we can remove `libpacker.so` from blueoil, though it needs some refactoring...

iizukak:
I close this PR temporary. Because I found removing `libpacker.so` dependency will take few days. I'll work on this task in near feature and then reopen this PR.

PR #70, opened 2018-12-10
Typo correction in unit_test.cpp

lm-lily:
Fix typo of add_1 (all_1).


dlk/backends/src/tb/unit_test.cpp
```diff
@@ -36,7 +36,7 @@ bool parse_input_type(int argc, char const *argv[], input_type &in_type) {
   } else if (std::strcmp(argv[1], "all_1") == 0) {
```
n-nez: are you going to update this one as well?

lm-lily: This one is correct, we don't have to change it. 
The correct one is "all_1".

n-nez: Oh... :sweat: Misread it as `add_1` somehow  :upside_down_face:

Issue #40, opened 2018-11-26
dlk jenkins output fatal: Not a a git repository


ruimashita:
dlk jinkins output 
```
.fatal: Not a git repository (or any of the parent directories): .git
```
<img width="583" alt="screen shot 2018-11-26 at 20 08 06" src="https://user-images.githubusercontent.com/138891/49010662-4690bd00-f1b7-11e8-9a2a-85c76b88b094.png">

full text:
http://jenkins.leapmind.local:8080/job/blueoil_dlk_test/107/consoleText

```
GitHub pull request #39 of commit dbc61b6dc8a1acb7ae1b16e7610a9600989664d2, no merge conflicts.
Setting status of dbc61b6dc8a1acb7ae1b16e7610a9600989664d2 to PENDING with url http://jenkins.leapmind.local:8080/job/blueoil_dlk_test/107/console and message: 'Build started for merge commit.'
Using context: [dlk] jenkins test
Building in workspace /var/lib/jenkins/workspace/blueoil_dlk_test
[WS-CLEANUP] Deleting project workspace...
[WS-CLEANUP] Deferred wipeout is used...
[WS-CLEANUP] Done
Cloning the remote Git repository
Cloning repository git@github.com:blue-oil/blueoil.git
 > git init /var/lib/jenkins/workspace/blueoil_dlk_test # timeout=10
Fetching upstream changes from git@github.com:blue-oil/blueoil.git
 > git --version # timeout=10
 > git fetch --tags --progress git@github.com:blue-oil/blueoil.git +refs/heads/*:refs/remotes/origin/*
 > git config remote.origin.url git@github.com:blue-oil/blueoil.git # timeout=10
 > git config --add remote.origin.fetch +refs/heads/*:refs/remotes/origin/* # timeout=10
 > git config remote.origin.url git@github.com:blue-oil/blueoil.git # timeout=10
Fetching upstream changes from git@github.com:blue-oil/blueoil.git
 > git fetch --tags --progress git@github.com:blue-oil/blueoil.git +refs/pull/*:refs/remotes/origin/pr/*
 > git rev-parse refs/remotes/origin/pr/39/merge^{commit} # timeout=10
 > git rev-parse refs/remotes/origin/origin/pr/39/merge^{commit} # timeout=10
Checking out Revision bbe223ffab206b910bf5f2593ecc2f49dd88940f (refs/remotes/origin/pr/39/merge)
 > git config core.sparsecheckout # timeout=10
[...]

PR #73, opened 2018-12-11
Improve kernel transform

lm-konda:
note:
This patch was created by programming with me and @lm-jira

## Motivation and Context
In order to improve performance,
prepare cache for the result of transformation.
This part costs about 10% of whole execution time.
related issues: #23

## Description
prepare results of transposed array result in advance(AOT).
And, use in QConv-kn2row.

## How has this been tested?
Validation with 2 patterns.
　Classification (LmnetV1Quantize / Caltech101 / epoch=2, other settings=default / DE10-nano)
　Object Detection (LMFYoloQuantize / OpenImagesV4 / epoch=2, other settings=default / DE10-nano)
run blueoil_test.sh
pair-programming (@lm-jira and @lm-konda)

no screenshot

## Types of changes


lm-konda:
Array passing is probably not so good. So please give me some advice.

lm-konda:
performance

before
```
Kernel transpose NHWC to HWNoCNi,12,11,21,58,310,530,239,7,7,5,2150,289,232,219,231,  sum:4.321ms
TotalRunTime,59321,  sum:59.321ms
```
after
```
Kernel transpose NHWC to HWNoCNi,7,5,4,3,5,3,3,4,5,4,3,4,3,3,3,  sum:0.059ms
TotalRunTime,56770,  sum:56.77ms
```
configuration
```
network
　LMFYoloQuantize
dataset
　OpenImagesV4
training configurattion
　epoch is 2. other is default.
hardware
　DE10-nano
how many times I run the benchmark
　As much as average average value is known (around 10 or more)
```

lm-konda:
In `Kernel transpose NHWC to HWNoCNi` section, there is no meaningful code.
So I removed this timer and work well.

lm-konda:
jenkins error is same as issue #40

tkng:
Thanks, removing Kernel transpose is one of our earnest desire 😄 

On the other hand, I have a question about the proposed design. This design assumes that the best memory layout is for convolution kernel is HWNoCNi. (BTW, what the `N` and `C` means? I'm not sure...) However, there might be other possibility. For example, Kanai-san's convolution assumes other memory layout.

To support such a possibility, how about keep the original memory layout and cache the transposed data also? That will consume more memory, but it would not be a problem in most cases.

I don't have strong opinion so far, I'm just asking a question. Please tell me your idea.

lm-konda:
Thank you for comments. I understand.

I think that we may be able to add kernel shape in `binary_convolution_parameters` and prepare according to this information.
How is this way of thinking?

lm-konda:
Supplementally, the `N` and `C` is from original C++ source.
We may have to change the variable to a name that is easy to understand.

lm-konda:
jenkins test fails in case
　data_format = "NCHW"
because NHWC is assumed.
I will fix it.

tkng:
> I think that we may be able to add kernel shape in binary_convolution_parameters and prepare according to this information.

Sometimes it's difficult to know the best kernel shape beforehand, because it depends on hardware. For example, the best kernel shape might be different at ARM cpu and FPGA. At the code generation phase, it would be difficult to know which hardware will be used.

There's another bad news, sometimes different hardware is used to execute different layer, even in 1 network. For example, we execute 1st (and last) non-quantized conv on CPU, but quantized convs are executed on FPGA. Alse, in the past, we considered to execute 1x1 conv on CPU, since it might difficult to support both of 1x1 and 3x3 convolution on the same FPGA.

The best way might be generating customized code for all possible hardware, but, it looks difficult so far for me.

dlk/python/dlk/core/optimizer.py
```diff
@@ -65,6 +65,69 @@ def node_is_activation_quantizer(node: Operator) -> bool:
...
+                    for w in range(kw):
+                        for c in range(k_c_by_word):
+                            idx_dst = h * (kw * kn_out * k_c_by_word * NUM_PE)
+                            idx_dst += w * (kn_out * k_c_by_word * NUM_PE)
+                            idx_dst += no * (k_c_by_word * NUM_PE)
+                            idx_dst += c * (NUM_PE)
+                            idx_dst += ni
+                            transpose_values[idx_dst] = copy_value[idx_src]
+                            idx_src += 1
+    elif dimension_format == "NCHW":
```
ruimashita: As discussion with @lm-konda @antonionevado , Maybe it is not need.

dlk/python/dlk/core/operators.py
```diff
@@ -612,20 +612,27 @@ def __init__(self,
                  data: np.ndarray,
                  dimension_format: str = 'NHWC',
                  packed: bool = False,
-                 actual_shape: List[int] = []) -> None:
+                 actual_shape: List[int] = [],
+                 transposed_data=None) -> None:
```
nlpng: Please include the `type` for consistency.

dlk/python/dlk/core/operators.py
```diff
@@ -612,20 +612,27 @@ def __init__(self,
...
         self._packed = packed
+        self._transposed_data = transposed_data
         super().__init__(name, shape, dtype, {}, data, dimension_format=dimension_format)
 
     @property
     def is_packed(self) -> bool:
         return self._packed
 
+    @property
+    def transposed_data(self):
```
nlpng: Please write the return `type` for consistency.

dlk/python/dlk/core/optimizer.py
lm-konda: Thank you. I remove this part.

dlk/python/dlk/core/operators.py
lm-konda: Thank you for review. Modified this.

dlk/python/dlk/core/operators.py
lm-konda: Fixed it.

dlk/python/dlk/core/optimizer.py
```diff
@@ -65,6 +65,60 @@ def node_is_activation_quantizer(node: Operator) -> bool:
...
 
+def transpose_kernels(kernel_data: np.ndarray,
+                      dimension_format: str,
+                      oh: int,
+                      ow: int,
+                      od: int,
+                      kh: int,
+                      kw: int,
+                      kd: int) -> List[int]:
+    NUM_PE = 16
```
iizukak: These parameters are fixed value?

dlk/python/dlk/core/optimizer.py
```diff
@@ -65,6 +65,60 @@ def node_is_activation_quantizer(node: Operator) -> bool:
...
+    for elem in kernel_data:
+        flatten_value.extend(elem)
+    while len(flatten_value) != k_size:
+        flatten_value.extend("0")
+    
+    copy_value = [0] * k_size
+    for i in range(od * kh * kw * k_c_by_word):
+        copy_value[i] = flatten_value[i]
+
+    transpose_values = [0] * k_size
```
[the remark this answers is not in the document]
lm-konda: I think so too. Thanks!

dlk/python/dlk/core/optimizer.py
lm-konda: Currently I say yes. But depends on configuration, these value may change.

ruimashita:
@lm-konda Need to resolve conflict, Plz 👍

lm-konda:
@ruimashita I solved the conflicts ! Please review this PR again.

dlk/python/dlk/core/optimizer.py
```diff
@@ -364,12 +365,72 @@ def pass_pack_weights(graph: Graph) -> None:
         data = packer.run(op_data.astype(np.float32), weight_quantizer.dimension)
 
         # Create the new constant with the quantized weights
+        def transpose_kernels(kernel_data: np.ndarray,
```
ruimashita: I prefer to avoid nested function except closure. Could you change the function to none nested ?

See also 
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#264-decision

dlk/python/dlk/core/operators.py
```diff
@@ -560,14 +560,16 @@ def __init__(self,
                  data: np.ndarray,
                  dimension_format: str = 'NHWC',
                  packed: bool = False,
-                 actual_shape: List[int] = []) -> None:
+                 actual_shape: List[int] = [],
```
ruimashita: Looks  `actual_shape` and `transposed_data` is duplicated with `dimension_format` and `data` 🤔   
In generally, less code is better than more code for maintenance and readable, I think,
Could you use `data` and `dimension_format` instead of `actual_shape` and `transposed_data` ? 😄

dlk/python/dlk/core/optimizer.py
lm-konda: I do not have confidence, but I think that only the function of pass_ should be present at the top level of optimizer.

dlk/python/dlk/core/operators.py
lm-konda: data don't mean transposed data because both data and transposed data may be used.

dimension format is such as `NCHW' (string),
and actual shape is such as [3,3,3,32] (List[int]).

These two are different.

dlk/python/dlk/core/optimizer.py
lm-konda: But, OK. I will fix it.

ruimashita: In python manner, adding `_` prefix is local function.
https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces
Thus, recommend name is `def _transpose_kernels()`, and don't need care it is top level.

dlk/python/dlk/core/optimizer.py
```diff
@@ -19,13 +19,66 @@
...
-from core.operators import Constant, Operator
+from core.operators import Constant, Operator, Conv
 from core.data_types import Uint32, QUANTIZED_NOT_PACKED
-from typing import cast
+from typing import cast, List, Any
 from collections import defaultdict
 from modules.packer import Packer
 
 
+def transpose_kernels(kernel_data: np.ndarray,
```
ruimashita: Please add `_` prefix because it is not public (global) function, and please write comment of what is this function.

dlk/python/dlk/templates/manual/consts/input.tpl.cpp
```diff
@@ -28,4 +28,12 @@ limitations under the License.
   {%- endfor %}
 };
 
+{% if node.transposed_data %}
```
ruimashita: I think `{{ node.name }}_transposed[]` is not need in C++ level, in order to reduce of memory size of `.elf` and `.so`, recommend to use only `{{ node.name }}[]` with `RUN_ON_FPGA` flag condition.

i.e.
```
{% if node.transposed_data %}

#if defined RUN_ON_FPGA
{{ node.dtype.cpptype() }} {{ node.name }}[] = {
  {% for d in node.transposed_data -%}
  {{- d -}},
  {%- endfor %}
};
#else
{{ node.dtype.cpptype() }} {{ node.name }}[] = {
  {% for d in node.data.flatten() -%}
  {{- d -}},
  {%- endfor %}
};

#endif

{% else -%}

{{ node.dtype.cpptype() }} {{ node.name }}[] = {
  {% for d in node.data.flatten() -%}
  {{- d -}},
  {%- endfor %}
};

{% endif %}
```

How is this ?

dlk/python/dlk/core/operators.py
ruimashita: Oh, I mistaken 😭 
`dimension_format` is different from `actual_shape`. Both `data` and `transposed_data` is need to compile for FPGA and others. Thanks.

Maybe there are more nice design to handle kernel format, and nice name of property of `transposed_data` , but I don't have idea, now 😭😭😭

dlk/python/dlk/core/optimizer.py
lm-konda: I fixed it.

dlk/python/dlk/templates/manual/consts/input.tpl.cpp
ruimashita: @lm-konda 
What I thinking as a problem is that there are 2 input data (`{{ node.name }}_transposed[]` and `{{ node.name }}[] ` ) in `input.cpp`, but because we only use one of these at compile time, useless data to increase memory.

in node name input.cpp
```
{{ node.name }}_transposed[] = {
0.2, 0.3. ...............................................
};

{{ node.name }}[] = {
0.4, 0.2. ...............................................
};
```

We should avoid to increase memory with useless data.
How is to add `use_transposed_data_on_fpga ` at `Constant` operation and below snippets ?

```
{% if node.use_transposed_data_on_fpga %}

#if defined RUN_ON_FPGA
{{ node.dtype.cpptype() }} {{ node.name }}[] = {
  {% for d in node.transposed_data -%}
  {{- d -}},
  {%- endfor %}
};
#else
{{ node.dtype.cpptype() }} {{ node.name }}[] = {
  {% for d in node.data.flatten() -%}
  {{- d -}},
  {%- endfor %}
};

#endif

{% else -%}

{{ node.dtype.cpptype() }} {{ node.name }}[] = {
  {% for d in node.data.flatten() -%}
  {{- d -}},
  {%- endfor %}
};

{% endif %}
```

lm-konda: I think `node.data.flatten()` include not only kernel data to be transposed but other data.
It's a little difficult problem.

PR #78, opened 2018-12-14
[WIP] Check augmentation can be used in target task

ruimashita:
to be written.

## Types of changes

## Checklist:


ruimashita:
I am thinking that task type (e.g. object detection) is not 1 on 1 related to dataset type (e.g. bbox, mask) , so we will need to change structure. We should write design documentation beforehand.

PR #80, opened 2018-12-17
utilize fused multipy and add

lm-konda:
## Motivation and Context
refactoring and convert mul+add to multiplyadd

## Description
add const specifiers
(refactoring is only for such kinds for a while, because the team coding style is not fixed yet).

change mul + add to multiplyadd
(The original coder seems to have not known the intrinsics specification).

(The both may improve performance slightly)

## How has this been tested?
Validation with 2 patterns.
　Classification (LmnetV1Quantize / Caltech101 / epoch=2, other settings=default / DE10-nano)
　Object Detection (LMFYoloQuantize / OpenImagesV4 / epoch=2, other settings=default / DE10-nano)
run blueoil_test.sh

no screenshot

## Types of changes


lm-konda:
We should change this code
```
    int r_tmp[4];
    vst1q_s32(r_tmp, r);
    output[i] = r_tmp[0];
    output[i+1] = r_tmp[1];
    output[i+2] = r_tmp[2];
    output[i+3] = r_tmp[3];
```
into below?
```
    output[i] =   r[0];
    output[i+1] = r[1];
    output[i+2] = r[2];
    output[i+3] = r[3];
```
note:
`r` is int32x4_t.

lm-konda:
I cut this part from the source code and confirmed assembly.
Compiler doesn't fuse multiply and add.

I think it is because ARMv7's FMA is FMAC.
I confirm simd store also.

lm-konda:
If we don't use simd store, instruction counts does not change, but memory accesses decrease.

PR #84, opened 2018-12-21
ignore .DS_Store class name

ruimashita:
ignore `.DS_Store` class name on Caltech format dataset

## Types of changes

## Checklist:


lmnet/lmnet/datasets/image_folder.py
```diff
@@ -60,6 +60,7 @@ def classes(self):
         """Returns the classes list in the data set."""
 
         classes = os.listdir(self.data_dir)
+        classes = [class_name for class_name in classes if class_name != ".DS_Store"]
```
iizukak: How about `classes .remove(".DS_Store"")` ?

ruimashita: @iizukak 
`remove()` raise ValueError when there is no item.
https://docs.python.org/3/tutorial/datastructures.html

iizukak: @ruimashita 
Should we care about no item case?

Anyway, I understood. It's ok to use list comprehension.

ruimashita: @iizukak
my poor english...
No item means the `classes` don't include `.DS_Store`.
`remove()` raise ValueError when `classes` don't include `.DS_Store`.

```
>>> fruits = ["apple", "lemon"]
>>> fruits.remove("orange")
  File "<stdin>", line 1, in <module>
ValueError: list.remove(x): x not in list
```

iizukak: Sorry, I miss read.

PR #85, opened 2018-12-25
[WIP] Support distributed training

hadusam:
Support distributed training with multi GPUs on the single node.

## Motivation and Context

`lmnet` has `is_distribution` option to run as distributed training with multi GPUs.
But this setting is not so easy, because we need to create the config file and modify dataset class for distribution.
My motivation is to make it easy that running distributed training.

## Description

To be easier to run distributed training by the user, I changed training script to select distribute or not dynamically. And if run with distribute option, it will be run with a fixed learning rate strategy for distributed training.
The details:
* scaling learning rate in proportion to the number of workers
* to change the boundary for the schedule of learning rate, scaling global step in proportion to the number of workers
* add `dist-train` operation into blueoil.sh
* add `--distribute` option into `blueoil/blueoil_train.py` and `lmnet/executor/train.py`
* add Dockerfile for distributed training and add build commads into `build_docker.sh`.
* add `DistributionInterface` into supported dataset classes (Caltech101, OpenImagesV4, DeLTA-Mark formats)

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Check out the [Installation](https://docs.blue-oil.org/install/install.html) and
 - GNU/Linux x86_64 with kernel version > 3.10
 - NVIDIA GPU with Architecture > Fermi (2.1)
 - NVIDIA drivers >= 367.48
-- Docker >=1.12 (or >=17.03.0)
+- Docker >=17.05.0
 - nvidia-docker >= 2.0
 
 The blueoil is run on docker container with original docker image based on NVIDIA's [CUDA images](https://github.com/NVIDIA/nvidia-docker/wiki/CUDA#requirements) (cuda:8.0-cudnn6-devel).
--- a/blueoil.sh
+++ b/blueoil.sh
@@ -142,11 +142,11 @@ function set_lmnet_docker_options(){
 
 function blueoil_train(){
 	set_variables_from_config $1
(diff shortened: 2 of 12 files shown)


tkng:
Hi @hadusam, thanks for this PR! I have some comments, please take a look.

# user interface

I don't want to increase the number of subcommands if possible. Is it possible to integrate `dist-train` subcommand into `train` subcommand? e.g., is it possible to integrate like `./blueoil.sh train config/test.yml --process-num 4`? (I don't think `--process-num` is a good name for the option, but I don't have good idea so far. sorry.)

# learning rate

> it will be run with a fixed learning rate strategy for distributed training.

I would like to freely select the scheduling strategy while enjoying distributed training. In my understanding, distributed training increase the batch size, but there's no other difference.

I think there're two choices.

1. do nothing with learning rate strategy
2. increase maximum learning rate automatically

If the number of distributed processes is 4, it's better to increase the learning rate by the factor of 4, but in most case, "just keep the current setting" will work enough. if there're a lot of distributed process like 64 processes, of course it's better to increase the value of learning rate, but in such a case, user will want to tune learning rate strategy carefully, since 1 experiment will cost a lot.

We know about the all learning rate strategy adopted in blueoil, hence we can increase the maximum learning rate automatically, but learning rate strategy is quite important hyper parameter, I want to have full control in this area.

PR #87, opened 2018-12-28
use xml.etree.ElementTree instead of bs4

tkng:
This PR rewrites Pascal VOC parser with `xml.etree.ElementTree`.

## Motivation and Context

I found that we can remove dependency to BeautifulSoup4 if we rewrite pascal voc dataset parser. I did. During the rewriting process, I found some problems in the current code.

- current code is trying to ignore "difficult" bounding boxes, but actually it looks failing to ignore, since the result of`difficult == 1` is always false, as the type of `difficult` is not an integer.
- curret code is not handling some special bounding boxes correctly. More concretely, if `object` tag includes `part` tags, current code use 1st bounding box, but it's inside the `part` tag in most cases. For example, `VOC2007/Annotations/008690.xml` includes annotation like below:

```
        <object>
                <name>person</name>
                <pose>Unspecified</pose>
                <truncated>1</truncated>
                <difficult>0</difficult>
                <part>
                        <name>hand</name>
                        <bndbox>
                                <xmin>166</xmin>
                                <ymin>242</ymin>
                                <xmax>221</xmax>
                                <ymax>272</ymax>
                        </bndbox>
                </part>
                <part>
                        <name>head</name>
                        <bndbox>
                                <xmin>212</xmin>
                                <ymin>15</ymin>
                                <xmax>282</xmax>
                                <ymax>89</ymax>
                        </bndbox>
                </part>
                <bndbox>
                        <xmin>147</xmin>
                        <ymin>14</ymin>
                        <xmax>337</xmax>
                        <ymax>267</ymax>
                </bndbox>
        </object>
```
We want to use last `bndbox` values (`147, 14, 337, 267`), but current code returns `166, 242, 221, 272`.

## Description

I just rewrite the code with `xml.etree.ElementTree`, since it is available by default.

- This rewriting also improved performance, though I didn't expected at first.
- I fixed `difficult` tag problem, now `difficult` bounding boxes will be ignored if `skip_difficult` option is set to `True`.
- I iterate elements in `object` tag by using `for e in list(obj):` idiom, to see just under the `object` tag, since I wanted to ignore bounding boxes in the `part` tags. New code returns expected values in the case of example above.
- After the modification, now XML parsing time is ignorable, hence I removed caching code, to reduce code size.

## How has this been tested?

we can test by executing `pascalvoc_2007_2012.py` directly.

```
before: elapsed: 35.47619867324829
after: elapsed: 1.383957862854004
```

Since this PR changes the behavior a bit, I run some experiments.

pascalvoc-test

lightblue is current code, red line is this PR. I expected that the accuracy will increase, but in fact it decreased a little. It's weird, I still cannot understand why :thinking:. I believe that this is within the error range.

## Types of changes

## Checklist:


ruimashita:
@tkng 
Thanks for finding and fixing the 2 problems (difficult and  1st part bbox).
I will verify the 2 problems fixed. Also, I checked `Bug fix (non-breaking change which fixes an issue)` in Types of changes.
Also, the result of the above experiments looks wired. Is the option of above experiment  `skip_difficult = True` ?

tkng:
> Also, the result of the above experiments looks wired. Is the option of above experiment skip_difficult = True ?

Yes, if I rembered correctly. Also I had an experiment with `skip_difficult = False`, but result looks similar to other settings.

ruimashita:
# default `skip_difficult` value.
The official PASCAL VOC 2007 say that difficult are  ignored in the evaluation,  and some public libraries `darknet` and `faster-rcnn` ignore difficult bbox also.
That's why I recommend `skip_difficult = True` as default.

official PASCAL VOC 2007: http://host.robots.ox.ac.uk/pascal/VOC/voc2007/htmldoc/index.html#SECTION00034000000000000000
> `difficult': an object marked as `difficult' indicates that the object is considered difficult to recognize, for example an object which is clearly visible but unidentifiable without substantial use of context. Objects marked as difficult are currently ignored in the evaluation of the challenge. 

public library
darknet: https://github.com/pjreddie/darknet/blob/master/scripts/voc_label.py#L37
faster-rcnn: https://github.com/rbgirshick/py-faster-rcnn/blob/781a917b378dbfdedb45b6a56189a31982da1b43/lib/datasets/pascal_voc.py#L189-L191

# yolov2 
I just evaluate yolov2 with darknet's public yolov2 weights (not training), with or without skip difficult. 
(download weight from https://pjreddie.com/darknet/yolov2/, using `executor/convert_weight_from_darknet.py yolov2` and `executor/evaluate.py -i convert_weight_from_darknet/yolo_v2` )

image size: 416x416
original paper's MAP: `76.8`

| skip difficult | MAP |
|----------|:-------------:|
| True | 77.41 |
| False | 69.83 |

screen shot 2019-01-11 at 21 27 17

The skip difficult MAP is just a little difference with original paper's.
What make this difference? Preprocess resize algorithm is different, but I'm not sure it is the reason.


# lm_fyolo_quantize 
Now, I am training lm_fyolo_quantize these 3 patterns.
* new dataset skip difficult
* new dataset not skip difficult
* old dataset

ruimashita:
# lm_fyolo_quantize training

I tried training LMFYolo model with below 3 patterns dataset using this [config](https://github.com/blue-oil/blueoil/blob/master/lmnet/configs/core/object_detection/lm_fyolo_quantize_pascalvoc_2007_2012.py) in order to verify this PR cahnges.
* skip difficult (this PR dataset class)
* no skip difficult (this PR dataset class)
* old (current master dataset class)

## results

|  dataset  | MAP at 80k steps | color of tensorboard |
| --- | --- | --- |
|  skip difficult | 41.31 | sky blue |
|  no skip difficult | 35.51 | red |
|  old | 32.87 | orange |

screen shot 2019-01-15 at 20 57 59

## discussion
In my experiments, This PR's `no skip difficult` MAP little bit increase compare to `old`. More over, `skip difficult` MAP  significantly increase compare to `old`  as we expected. These results and above yolov2 evaluation indicate this PR changes are correct. The yolov2 evaluation takes `0.6` points MAP difference from original paper, It is still problem, but the difference maybe come from pre-processor or evaluation method, I think.
 
# verify special bbox case `VOC2007/Annotations/008690.xml`
So I verify special bbox case `VOC2007/Annotations/008690.xml` by `python3 tests/check_dataset_storage.py` of this changes: https://github.com/ruimashita/blueoil/commit/abed28a47aefe776dd6929a8599891adfc0f87be
old (current mater):
screen shot 2019-01-15 at 22 17 07


this PR:
screen shot 2019-01-15 at 22 12 59


# Can I take over ?
The verification of this PR is finished, It means this nice PR should be merged ASAP, But I guess you have no time to fix some tiny things for merged.
@tkng Can I take over this PR from you ?
I will do these things. 
* Change default `skip_difficult` value to `True`
* Fix `flake8` errors
* Delete `BeautifulSoup4` dependency from python codes and `requirements.txt`

PR #97, opened 2019-01-09
Utilize GCC builtin popcnt

lm-konda:
## Motivation and Context
Using GCC builtin popcount, improve performance (mainly X86).

## Description
Original code calculates popcount not using hardware native instruction.
So change to utilize it.

compiler flag -march=native is very important to use native popcnt.

## How has this been tested?
run on X86 and ARM(for check).

## Types of changes


lm-konda:
I tried GCC builtin_popcount on ARM. And the performance is same as original.
Because GCC builtin_popcount is expanded to native instruction if it exists, 
conversely if native instruction does not exist, output optimal code for the target.

＞Also, could you make each pull request independently, so that the reviewer can concentrate to the each pull request? 
Sorry. I was concerned about it. Some PR depends on previous PR, but review process is not completed...

lm-konda:
Supplementarily, LLVM outputs popcount code from only prototype without implementation.
So, this PR includes code gen test of LLVM.

iizukak:

Issue #99, opened 2019-01-10
redundant or bug branch instruction (not for FPGA version)


lm-konda:
same conditional branch. It may not be intended.

https://github.com/blue-oil/blueoil/blob/4e6abd97f5f740af89484c08a7d69934e58c2d56/dlk/python/dlk/templates/src/func/impl/apply_thresholds.cpp#L49
https://github.com/blue-oil/blueoil/blob/4e6abd97f5f740af89484c08a7d69934e58c2d56/dlk/python/dlk/templates/src/func/impl/apply_thresholds.cpp#L53

PR #100, opened 2019-01-11
fix branch bug

lm-konda:
## Motivation and Context
Fix apply threshold logic bug
related to #99 

## Description
change branch condition same as
```
https://github.com/blue-oil/blueoil/blob/master/dlk/python/dlk/templates/src/func/impl/quantized_conv2d_dim2col.cpp
```
This code looks more correct.

## How has this been tested?
run on X86 / ARM / FPGA. All networks are correct.

## Types of changes


tkng:
Sorry, I still don't fully understand the code, but the original code looks correct for me. Could you explain a little bit more detail?

lm-konda:
The original repository is probably updated after this PR.
Because I committed once and difference is these part only.

lm-konda:
This is an evidence.
https://github.com/blue-oil/blueoil/blob/v0.2.1/dlk/python/dlk/templates/src/func/impl/apply_thresholds.cpp

tkng:
Oops, sorry, I was checking the code after you fixed 🙇 

LGTM.

Issue #101, opened 2019-01-11
Use Advanced SIMD on AArch64


lm-konda:
We should use Advanced SIMD instead of NEON for performance.
Performance gain is probably not so bad.

And we have to change compile flag also (I had overlooked).
It may be necessary to consider whether compatibility between NEON and Advanced SIMD is possible.

Issue #94, opened 2019-01-09
generic code type issue (ARM binary)


lm-konda:
As for arm binary, std::memset type causes an error.
```
src/func/impl/generic/quantized_conv2d_kn2row.cpp:138:17: error: cannot initialize a parameter of type
      'void *' with an lvalue of type 'BIN_CONV_OUTPUT *' (aka 'volatile short *')
    std::memset(kn2row_buf, 0, bufsize);
                ^~~~~~~~~~
/usr/arm-linux-gnueabihf/include/string.h:62:28: note: passing argument to parameter '__s' here
extern void *memset (void *__s, int __c, size_t __n) __THROW __nonnull ((1));
```

It may occur when you use llvm7 or below.

lm-konda:
already fixed. but I will update with #95 patch.

iizukak:
This bug looks only occur when using LLVM.
And currently, we have no plan to support LLVM, So I close this issue.

PR #102, opened 2019-01-11
Enable aarch64 build

lm-konda:
## Motivation and Context
AArch64 correspondence was inadequate,
so respond formally.
related to #101 (not work on this yet), #94 

## Description
・add new repo and aarch64 cross toolchain
・modify blueoil driver to build aarch64 binary
・modify Makefile template to build aarch64 binary properly
　(performance tuning is from now on)

modified
　blueoil/blueoil_convert.py
　dlk/python/dlk/templates/CMakeLists.txt
　dlk/python/dlk/templates/cmake/config.cmake
　dlk/python/dlk/templates/include/matrix/multiplication.h
　dlk/python/dlk/templates/include/quantizer.h
　dlk/python/dlk/templates/src/func/impl/generic/quantized_conv2d_kn2row.cpp
　dlk/python/dlk/templates/src/func/quantized_conv2d.cpp
　dlk/python/dlk/templates/src/matrix/shift_add.cpp
　dlk/scripts/dlk_script_wrapper.sh
　docker/Dockerfile

## How has this been tested?
not tested yet because there is no evaluation environment.

## Types of changes


lm-konda:
partial `readelf -a` of lm_aarch64.elf result.
binary format is .
```
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           AArch64
  Version:                           0x1
  Entry point address:               0x70000
  Start of program headers:          64 (bytes into file)
  Start of section headers:          5175360 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         9
  Size of section headers:           64 (bytes)
  Number of section headers:         42
  Section header string table index: 40
```

lm-konda:
When I used aarch64 cross toolchain(GCC), there is no problem. (able to get asm by objdump from binary)

It is may be an LLVM cross compiler problem (and not able to get asm using objdump from binary)

lm-konda:
With the compiler flag -Wl,-s, errors in the symbol table can be avoided. After that, I only need to confirm the execution.
(maybe we should use -Wl,-s instead of invoking strip command)

lm-konda:
Installed OS for AArch64(raspberry pi) and run binary

```
localhost:~ # uname -a
Linux localhost.localdomain 4.12.14-lp150.11-default #1 SMP Fri May 11 08:28:30 UTC 2018 (a9fee09) aarch64 aarch64 aarch64 GNU/Linux
localhost:~ # ./lm_aarch64.elf input.npy output.npy 
localhost:~ # 
```
The result is OK!

lm-konda:
close once.

Issue #27, opened 2018-11-14
WIDER_FACE dataset includes some 0 size bounding boxes


tkng:
Though we don't make it public, we support WIDER_FACE dataset format class. However, WIDER_FACE  includes some weird annotations, the height or width of bonding box is 0.

We have some choices to fix the problem.

- just skip 0 size bounding box
- set minimum width/height for bounding box
- just skip the image if a bounding box has 0 width or 0 height.

I don't know which one is better, but the 2nd one looks better.

lm-konda:
Can we fix it in this way? (and set values ​​appropriately)
```
                    x, y, w, h, blur, expression, illumination, invalid, occlusion, pose, _ = line.split(" ")
                    w = max(w, 1); h = max(h, 1); # add this line 178
                    temp = [int(x), int(y), int(w), int(h), 0]
                    bbox.append(temp)
```

iizukak:
@lm-konda @tkng 
Size 0 annotations have any reason?
If annotations are meaningless,

I think it's ok to skip
```
x, y, w, h, blur, expression, illumination, invalid, occlusion, pose, _ = line.split(" ")
if w <= 0 or h <= 0:
    continue
```

or set some threshold

```
x, y, w, h, blur, expression, illumination, invalid, occlusion, pose, _ = line.split(" ")
if w <= threshold or h <= threshold:
    continue
```
https://github.com/blue-oil/blueoil/blob/master/lmnet/lmnet/datasets/widerface.py#L177

tkng:
@iizukak
> Size 0 annotations have any meaning?

I checked some images, found that size 0 annotation don't have any meaning.

tkng:
This is an example, this image doesn't have any annotation except `0 0 0 0`. I think it seems better to skip the image itself, in the case of WIDER_FACE dataset.

![0_parade_parade_0_452](https://user-images.githubusercontent.com/96674/48828543-5d986f00-edb3-11e8-8a09-090fccaf12a0.jpg)

a-hanamoto:
@tkng https://github.com/blue-oil/blueoil/blob/master/blueoil/datasets/widerface.py#L108
Now it seems that the script skips the image if a bounding box has 0 width or 0 height.
Can we close this issue?

tkng:
@a-hanamoto Yes, we can.

PR #119, opened 2019-01-22
[WIP] Modify WIDER_FACE

lm-konda:
## Motivation and Context
0 size bounding boxes are harmful.
So skip them and improve accuracy.

related to #27 
＞Though we don't make it public, we support WIDER_FACE dataset format class. However, ＞WIDER_FACE includes some weird annotations, the height or width of bonding box is 0.

＞I think it seems better to skip the image itself, in the case of WIDER_FACE dataset.

## Description
apply @iizukak san's method (line 178).
```
if int(w) <= 0 or int(h) <= 0:
    continue
```

## How has this been tested?
not tested. We need to test.

## Types of changes


tkng:
Thanks for sending this pull request! However, this pull request contains multiple changes those should be reviewed as other pull requests.

Could you remake this pull request in more concise manner?

lm-konda:
Sorry. I will fix it.

iizukak:

PR #122, opened 2019-01-22
implement DatasetIterator class

tkng:
This pull request implements a new class `DatasetIterator`. The class acts like a current dataset class, e.g. "Cifar10" class. Since the `DatasetIterator` itself supports data shuffling, we can eliminate data shuffling code from dataset classes.

Moreover, the new `DatasetIterator` class supports distributed training on Horovod. Currently, if someone wants to use Horovod, then he/she need to write "distributed" version of a dataset class. Since the new `DatasetIterator` class supports the feature, dataset class need not to care about Horovod anymore.

In addition, the new "DatasetIterator" class supports asynchronous muilti process prefetching. Hence, all dataset class will be much faster. For example, the training of ImageNet will be over 3 time faster than the current code.

## Motivation and Context

Currently, lots of duplications are exists in `lmnet/lmnet/datasets` directory. We should avoid code duplication for faster feature development.

Also, dataset class needs a wrapper that supports distributed training and multi process asynchronous prefetching. DatasetIterator class is made for such purpose.

## Description

Currently, the most important method of a dataset class is `feed`, which returns a minibatch of a dataset. Instead of the `feed` method, after this PR merged, a dataset class should provide `__getitem__` method, which returns a datum of the dataset.

## How has this been tested?

I confirmed that the all unit tests are passed by the following command:

```
docker-compose run --rm tensorflow  pytest
```

However, I'm really worring about the code quality, since the PR is very large. Maybe I'm missing something.

## Thanks to

Thanks to @mymultiverse and @yd8534976, they cooperated to make this feature while they are in the last summer internship of LeapMind.

## Types of changes

## Checklist:


tkng:
force-pushed to rebased to master

ruimashita:
@tkng Thanks

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -0,0 +1,331 @@
...
+        # TODO(tokunaga): the number of processes should be configurable
+        self.seed = seed + 1  # seed must not be 0
+        self.support_getitem = hasattr(dataset, "__getitem__")
+        self.pool = Pool(processes=8, initializer=_prefetch_setup,
+                         initargs=(dataset, self.seed, not self.support_getitem))
+        self.result_queue = result_queue
+        self.batch_size = dataset.batch_size
+        self.dataset = dataset
+        self.data_ids = []
+        self.terminate = False
```
yd8534976: I'm sorry to comment on this old PR, yet when reading this file, I have some questions.
- What's the purpose of this `self.terminate`, I think it keeps `False` always.

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -0,0 +1,331 @@
...
+
+    cifar10 = Cifar10()
+
+    augmentor = Sequence([
+        FlipLeftRight(0.5),
+        Hue((-10, 10)),
+        Blur(),
+    ])
+
+    dataset_iterator = DatasetIterator(dataset=cifar10, enable_prefetch=True, augmentor=augmentor)
```
yd8534976: Unexpected arg `augmentor`

yd8534976:
When I test it with code:

    from lmnet.datasets.cifar10 import Cifar10
    from lmnet.data_processor import Sequence
    from lmnet.data_augmentor import FlipLeftRight, Hue, Blur

    cifar10 = Cifar10()

    augmentor = Sequence([
        FlipLeftRight(0.5),
        Hue((-10, 10)),
        Blur(),
    ])

    dataset_iterator = DatasetIterator(dataset=cifar10, enable_prefetch=True)
    time.sleep(2)
    import time
    t0 = time.time()
    for i in range(10000):
        data_batch = next(dataset_iterator)
    t1 = time.time()
    print("time of prefetch: {}".format(t1 - t0))

    dataset_iterator2 = DatasetIterator(dataset=cifar10, enable_prefetch=False)
    t0 = time.time()
    for i in range(10000):
        data_batch = next(dataset_iterator2)
    t1 = time.time()
    print("time with/o prefetch: {}".format(t1 - t0))

I got a result:
Screen Shot 2019-05-07 at 17 03 26

lmnet/lmnet/datasets/dataset_iterator.py
ruimashita: @yd8534976 
Yes. `self.terminate` keep `False` unfortunately , It need to change to handle terminate.

lmnet/lmnet/datasets/dataset_iterator.py
ruimashita: @yd8534976 
Sure~. Need to fix

ruimashita:
@yd8534976 
In the current code, pre-fetch number of batch is `200` , and num of parallel process is `8`. Maybe that is not good for your environment.
Just you can change these parameter in source code to fit your environment and this PR is enable to change these params from config file. 
https://github.com/blue-oil/blueoil/pull/186

yd8534976:
@ruimashita Thanks very much for your patience!

PR #118, opened 2019-01-22
Separate generic kn2row

lm-konda:
## Motivation and Context
We can remove branch from generic kn2row implementation now,
because I have separated qconv implementation into 3x3 case and 1x1 case.
Performance improvement is expected.

This PR is based on #114 

## Description
There are two branch patterns.
(kernel height(kh) / kernel width(kw))
case 1) if (kh == 3 && kw == 3)
case 2) if (kh == 1 && kw == 1)

in case 1, kn2row_3x3 function need not this branch.
in case 2, kn2row_1x1 function need not this branch.

So, remove them.

## How has this been tested?
run on ARM / FPGA / X86

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature / Optimization (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #123, opened 2019-01-24
Refactoring two apply threshold implementations to one

lm-konda:
## Motivation and Context
There were two apply_threshold implementations.
one is in generic quantized_conv2d_kn2row.cpp
another is in apply_threshold.cpp

Since the definition is the same and confused,
I decided to adopt one `apply_threshold.cpp`

After being merged,
I plan to update #118 and move the position of apply_threshold.cpp.
The reason is to think that TF function only should exist in the `func` directory.

## Description
remove apply_threshold definition in generic quantized_conv2d_kn2row.cpp
and include header for apply_threshold.cpp

Now the definition of apply_threshold is one.

## How has this been tested?
run on ARM / FPGA / X86

## Types of changes


iizukak:

PR #124, opened 2019-01-25
Update llvm to 7.0.1

lm-konda:
## Motivation and Context
Today pre-built binary of ARM / AArch 64 version LLVM 7.0.1 was released,
(also for x86 version) apply them.
Because this time it is a minor version upgrade,
motivation is mainly improvement of quality.

This PR depends on #118 
The reason is that master's aarch64 correspondence is still, to confirm whether there is no problem with the previous modification of each arch etc. 

## Description
There are two modifications.

modify Dockerfile to download and install
llvm7.0.1 pre-built binary.

modify Makefile to use llvm7.0.1 instead of llvm7.0.0.

## How has this been tested?
run on ARM / AArch64 / X86 / FPGA.

## Types of changes


iizukak:

PR #126, opened 2019-01-25
Enable to build ARM binary with llvm

lm-konda:
## Motivation and Context
For performance, I think it's good to compile with llvm.
So, I responded to compile.

## Description
At first, make to download llvm pre-built binary and install.
and I modified Makefile to use installed llvm.

Change .S file's syntax to enable compilation.

OpenMP is not default in llvm, so uncomment macro.

modified
　docker/Dockerfile 
　dlk/python/dlk/templates/src/pack2b_neonv7.S

## How has this been tested?
run on ARM

## Types of changes


iizukak:

PR #128, opened 2019-01-25
Tune performance of llvm aarch64 binary

lm-konda:
## Motivation and Context
#127 AArch64 correspondence has no tuning, so bring a possible place from ARM,
Improve performance.

AArch64 instructions and AArch32 instructions are similar, but differ.

## Description
This PS is based on #127 

I introduced the USE_ASIMD macro
so that we can use what we can do in the part branched by USE_NEON.

And I changed the file itself to be compiled to tuned one for ARM.

modified
　dlk/python/dlk/templates/CMakeLists.txt 
　dlk/python/dlk/templates/cmake/config.cmake 
　dlk/python/dlk/templates/include/quantizer.h 
　dlk/python/dlk/templates/src/matrix/shift_add.cpp 
　dlk/python/dlk/templates/include/matrix/multiplication.h

## How has this been tested?
run on AArch64 / X86 / ARM / FPGA

## Types of changes


iizukak:

PR #153, opened 2019-02-14
Fix branch condition in generic quantized matmul

lm-konda:
## Motivation and Context
Branch condition in generic quantized matmul is not correct.
There are same conditional branches, so in small block_size case,
It may fail.

## Description
src file:
　src/matrix/generic/quantized_multiplication.cpp
```
            if (A.rows() - i > 0) {★
              T_UINT a = *A.data(i+0, j+j2);
              if (end - k > 0) CONV(0, 0)
              if (end - k > 1) CONV(0, 1)
              if (end - k > 2) CONV(0, 2)
              if (end - k > 3) CONV(0, 3)
            }
            if (A.rows() - i > 0) {★
              T_UINT a = *A.data(i+1, j+j2);
              if (end - k > 0) CONV(1, 0)
              if (end - k > 1) CONV(1, 1)
              if (end - k > 2) CONV(1, 2)
              if (end - k > 3) CONV(1, 3)
            }
            if (A.rows() - i > 0) {★
              T_UINT a = *A.data(i+2, j+j2);
              if (end - k > 0) CONV(2, 0)
              if (end - k > 1) CONV(2, 1)
              if (end - k > 2) CONV(2, 2)
              if (end - k > 3) CONV(2, 3)
            }
            if (A.rows() - i > 0) {★
              T_UINT a = *A.data(i+3, j+j2);
              if (end - k > 0) CONV(3, 0)
              if (end - k > 1) CONV(3, 1)
              if (end - k > 2) CONV(3, 2)
              if (end - k > 3) CONV(3, 3)
            }
```
four star parts are same conditional branches.

So fixed the same as another architecture implementation.

refs:
　src/matrix/arm_neon/quantized_multiplication.cpp

## How has this been tested?
run on X86.
```
konda@leapmind-server-08:~/work/blueoil/lmnet/saved/konda_20190123114447/export/save.ckpt-957/128x128/output/models/lib$ ./lm_x86.elf input.npy output.npy 
```

## Types of changes


tkng:
Thanks for fixing the bug!

Could you add a unit test to prevent the same bug?

lm-konda:
OK. I will consider a method.

lm-konda:
I thought about the problem and asked @iizukak about the problem, but since I could not get a conclusion please let me question.
In this case, problems arise when A.rows() is less than 4.
Is the unit test at that time assert by preparing a special input which is less than 4,
or detecting an equivalent branch by static analysis?

iizukak:

PR #156, opened 2019-02-15
Remove redundant -g option in Makefile

iizukak:
## Motivation and Context

Remove some redundant option in Makefile.

## Description

I don't know why, but there is some multiple `-g` in some target.
I simply remove that.

## How has this been tested?

To be written.

## Types of changes

## Checklist:


iizukak:
@lm-konda I just created PR. Thanks for suggestion.

Issue #150, opened 2019-02-12
There are cases to fail packing


lm-konda:
in src file：
　dlk/python/dlk/templates/src/pack_input_to_qwords.cpp

`nbit_qinput_word`  equals 32.
And when fractional processing does not go well,
there is a possibility of erroneous results.

```
  for (; idx_in < len; idx_in++) {
    QUANTIZED_NOT_PACKED tmp_input = input[idx_in];
    unsigned int b0 = tmp_input & 0x1;
    unsigned int b1 = (tmp_input & 0x2) >> 1;

    qinput_words_buf[0] |= (b0 << bit_count);
    qinput_words_buf[1] |= (b1 << bit_count);
    ++bit_count;

    if (bit_count == nbit_qinput_word) ★not reach nbit_qinput_word
      {
        for (unsigned i_bit = 0; i_bit < input_bitwidth; i_bit++) {
          output[idx_out++] = qinput_words_buf[i_bit];
          qinput_words_buf[i_bit] = 0;
        }

        bit_count = 0;
      }
  }
```

PR #157, opened 2019-02-15
[WIP] Fix pack_input_to_qwords

lm-konda:
## Motivation and Context
function pack_input_to_qwords was buggy.
It might not write back full data to output.

So I fixed it.

related issue #150 

## Description
change not to check full length(check if equals to 32) and write back.

Modify to update remainder output after loop-ends.

## How has this been tested?
run on X86 / ARM(without -DUSE_NEON) / FPGA(without -DUSE_NEON)

## Types of changes


iizukak:
@n-nez @n-nez This PR also need your review 🙇

dlk/python/dlk/templates/src/pack_input_to_qwords.cpp
```diff
@@ -68,24 +68,21 @@ void pack_input_to_qwords(QUANTIZED_NOT_PACKED input[],
...
+  {
+    for (; idx_in < len; idx_in++) {
+      QUANTIZED_NOT_PACKED tmp_input = input[idx_in];
+      unsigned int b0 = tmp_input & 0x1;
+      unsigned int b1 = (tmp_input & 0x2) >> 1;
+
+      qinput_words_buf[0] |= (b0 << bit_count);
+      qinput_words_buf[1] |= (b1 << bit_count);
+      ++bit_count;
+    }
```
n-nez: There is another corner case problem with the code:
As `qinput_words_buf` is a static buffer, it should be zeroed each time we have to zero it each time we completed use it, to keep expression `qinput_words_buf[0] |= (b0 << bit_count);` work as expected.
And as now we do it only under condition `if (bit_count == nbit_qinput_word)` it will keep old value for cases when `len` is not multiple of `4`. And the old value will be used to accumulate packed values on the next call of the function. :sweat_smile: 

We could fix it by zeroing it in the loop below like:
```cpp
for (unsigned i_bit = 0; i_bit < input_bitwidth; i_bit++) {
  output[idx_out++] = qinput_words_buf[i_bit];
  qinput_words_buf[0] |= (b0 << bit_count);
}
```

At the other hand, as only two values `qinput_words_buf[0]` and `qinput_words_buf[1]` are ever used in this code, we could stop to pretend to be generic, and just do something like:
```cpp
int qinput_words_low = 0;
int qinput_words_high = 0;
```
and stop pretending to be generic.

did not look at the assembler, but doing so we also should simplify compiler life for putting the values into registers, instead of storing them into memory, to be able to traverse them in the for loop later...

lm-konda:
Thank you for review!
As we talked about, in order to correct the behavior, I will clear the static memory region at first.
And let's make an issue other improvements(using two registers).

n-nez:
LGTM

iizukak:
This WIP PR has no progress recently, So I close this PR temporary.

n-nez:
@iizukak What kind of progress do you want?
It was reviewed and approved, two weeks before, we should been only merge it... :thinking:

iizukak:
@n-nez I hesitate to merge this PR, because the title still include `WIP` .
If you think this PR was completed, Please remove `WIP` and merge 🙇 

Thank you for your attention.

n-nez:
@iizukak Seems github doesn't allow me change the PR name...
I can click `Edit` button and change title, but clicking `Save` doesn't change anything... :thinking: 

Could still merge it when tests pass, anyway.

iizukak:
@n-nez 
Oh..., I also tried and fail to save 🤕

PR #159, opened 2019-02-15
Refactoring for source codes in dlk/backends directory

tkclimb:
I refactored many files in `backends` directory.
This doesn't change the behavior, just making it more visible and understandable.
This is mainly for the Blueoil training course (device-porting) but still useful on the purpose to clean the projects.


dlk/backends/include/common/global.h
```diff
@@ -21,7 +21,6 @@ limitations under the License.
 #include "common/type_info.h"
 
 #include "params/conv.h"
-#include "params/gemm.h"
 
 #define IP_CSR_ADDR 0xFF200000
 #define A8W1_IP_CSR_ADDR 0xFF200100
```
lm-lily: Is line-27 still in use?

dlk/backends/include/cpp.h
```diff
@@ -52,27 +52,6 @@ void conv(T_in in_data[], T_out out_data[], T_k k_data[], T_out threshold_data[]
...
-  assert(stride == 1);
-
-  cpp::conv_kn2row_tiling_impl(in_data, out_data, k_data, threshold_data, in_w, in_h, in_c, out_w, out_h, out_c, KW, KW,
-                               pad, stride);
-}
-/// @}
-
 /// 2A 1W quantized convolution with 1x1 and 3x3 kernel
 template <int KH, int KW>
 void qconv(T_q in_data_packed[], T_out out_data[], T_q k_data_packed[], unsigned in_w, unsigned in_h,
```
n-nez: do we need qconv? :thinking: 

if it is something like implementation example we could probably separate it from the main one.

dlk/backends/include/cpp.h
```diff
@@ -90,28 +69,26 @@ void qconv(T_q in_data_packed[], T_out out_data[], T_q k_data_packed[], unsigned
...
   assert(((KH == 3) && (pad == 1)) || ((KH == 1) && (pad == 0)));
   assert(stride == 1);
 
-  qconv_with_kn2row_impl(in_data_packed, out_data, k_data_packed, out_data, threshold_data, in_w, in_h, in_c_by_word,
-                         out_w, out_h, out_c, KH, KW, pad);
+  cpp::conv_kn2row_tiling_impl(in_data, out_data, k_data, threshold_data, in_w, in_h, in_c, out_w, out_h, out_c, KW, KW,
+                               pad, stride);
 }
 
 void qconv_kn2row_tiling_impl(T_q in_data[], T_out out_data[], T_q k_data[], T_out threshold_data[],
```
n-nez: Should be asking in the previous PR, but what is the reason for splitting implementation into `qconv_with_kn2row` and `qconv_with_kn2row_impl`, considering that we have only *one* implementation? :thinking:

dlk/backends/include/params/conv.h
```diff
@@ -269,52 +269,6 @@ static const unsigned num_in_line = (max_in_w_with_pad * (k_h - 1)) + k_w;
 static const bool has_thresholds = true;
 } // namespace conv1x1_params
```
n-nez: do we still need `conv3x3_params` and `conv1x1_params` ? :thinking:

PR #160, opened 2019-02-18
[WIP] Improve qconv (generic version)

lm-konda:
## Motivation and Context
I found out that MAX_SIZE_KN2ROW_BUFFER_PER_LAYER can be used in generic qconv.
So, prepare the static region in advance and improve performance.

## Description
modify not to reallocate everytime and reuse static memory.

zero clear also probably not needed, because matmul is executed after that (later confirm in detail).

modified：
　src/func/impl/generic/quantized_conv2d_kn2row.cpp

## How has this been tested?
run on X86(use generic pattern)

(and later, confirm whether zero clear is needed or not.)

## Types of changes


iizukak:
This WIP PR has no progress recently, So I close this PR temporary.

PR #162, opened 2019-02-18
Prepare static region for conv2d

lm-konda:
## Motivation and Context
kernel width and kernel height are known, and max input channel and max output channel are defined.
So by preparing static region and reusing, we can improve performance.

In addition, @n-nez -san suggested the way to pre-compute
the value of transformed value at conv2d in advance. 

## Description
MAX_IN_C (for max value for channel) is pre-defined.
kernel height and kernel width are equal to 3.

So we can know the max number of kernel size.
and remove dynamic memory allocation.

Modified：
　src/func/conv2d.cpp

## How has this been tested?
run on X86 / ARM

## Types of changes


iizukak:

PR #166, opened 2019-02-21
Implement SimpleDatasetReader

ruimashita:
which is read batch size data from Dataset object simply (not using thread and multiprocess) when DatasetIterator is not enable_prefetch.

## Motivation and Context

As in #122, we implement DatasetIterator class, But when we use enable_prefetch is False, it is not good for debugging because dataset iterator using thread.

## Description

* Implement `_SimpleDatasetReader` 
* Implement unit tests.
  * `test_dataset_iterator_batch_size()` which is asset that dataset iterator give batch size data.
  * `test_dataset_iterator_batch_orde()` which is asset that data given by iterator is same whether enabele_prefetch ture or false.
* Delete `_SingleProcessDatasetPrefetchThread` class, it is not used.

## Types of changes

## Checklist:


ruimashita:
@iizukak @nlpng @lm-jira Plz review 👍

iizukak:
@ruimashita 
`_SimpleDatasetReader_` 's name is maybe too generic.
Is there more descriptive name for this class?  I have no idea...

ruimashita:
@iizukak
Yes 👍 the class has bad name, I thought about class name a lot, But I didn't come up with.
ok, I will re-think

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -1,6 +1,5 @@
 from multiprocessing import Pool
```
nlpng: Should this have the copyright label as well?

ruimashita: @nlpng 
Oh, thanks. I will add copyright.

ruimashita: changed https://github.com/blue-oil/blueoil/pull/166/commits/667d13857325ceccfa463d49024bcfb9d0556ffe

Issue #169, opened 2019-02-25
OverflowError: Python int too large to convert to C long


ruimashita:
When I try to convert model,  get the below error.
```
Create pb and yaml files in: /home/blueoil/saved/person_segmentation/13_v7/export/save.ckpt-130000/352x480
Create npy files in under each image name's folder
pb: minimal_graph_with_shape.pb
yaml: /home/blueoil/saved/person_segmentation/13_v7/export/save.ckpt-130000/352x480/config.yaml, /home/blueoil/saved/person_segmentation/13_v7/export/save.ckpt-130000/352x480/meta.yaml
npy: ['000_images_placeholder:0', '510_output:0', 'raw_image', 'preprocessed_image']
finish
import pb file
optimize graph step: start
Traceback (most recent call last):
  File "./blueoil/cli.py", line 166, in <module>
    main()
  File "/usr/local/pyenv/versions/python3.6/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/pyenv/versions/python3.6/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/pyenv/versions/python3.6/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/pyenv/versions/python3.6/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/pyenv/versions/python3.6/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "./blueoil/cli.py", line 119, in convert
    export_dir = run_convert(experiment_id, checkpoint, template)
  File "/home/blueoil/blueoil/blueoil_convert.py", line 159, in run
    cache_dma=cache_dma
  File "/home/blueoil/dlk/python/dlk/scripts/generate_project.py", line 149, in run
    optimize_graph_step(model, config)
  File "/home/blueoil/dlk/python/dlk/scripts/generate_project.py", line 60, in optimize_graph_step
    pass_compute_thresholds(graph)
  File "/home/blueoil/dlk/python/dlk/core/optimizer.py", line 362, in pass_compute_thresholds
    else int(math.ceil(th_per_ch))
OverflowError: Python int too large to convert to C long
```

convert command:
[...]

PR #170, opened 2019-02-26
Fix overflowError: Python int too large to convert to C long

nlpng:
## Motivation and Context

This PR is trying to hot fix the issue of #169

## Description

Main change is in the compute threshold pass (`pass_compute_threshold`) of optimizer.py

## How has this been tested?

The overflow error can be reproduced if the kernel weight is a extremely small value.
Test is plotted as below:
```python
    def test_pass_compute_thresholds_for_huge_threshold_values(self) -> None:
        """Test pass."""
        data1 = np.float32(np.random.rand(1, 2, 2, 3))
        data2 = np.float32(np.random.uniform(10 ** (-30), 10 ** (-40), size=(1, 2, 2, 3)))
        graph1 = self.create_sample_graph(data1, data2)

        pass_compute_thresholds(graph1)

        self.assertEqual(graph1.get_op('conv2').has_thresholds, True,
                         '[Failed] Found threshold of Conv not calculated')

        print("Test pass #8-1 compute_thresholds of enormously small values passed!")
```
By experimentally feeding super small weights of `data2` (around 10**(-30) - 10**(-40)) to the optimization pass of `pass_comput_threshold`, we will get following errors:
```
======================================================================
ERROR: test_pass_compute_thresholds_for_huge_threshold_values (test_optimizer.TestPassComputeThresholds)
Test pass.
----------------------------------------------------------------------
  File "/Users/huang/oss_ws/blueoil/dlk/tests/test_optimizer.py", line 463, in test_pass_compute_thresholds_for_huge_threshold_values
    pass_compute_thresholds(graph1)
  File "/Users/huang/oss_ws/blueoil/dlk/python/dlk/core/optimizer.py", line 370, in pass_compute_thresholds
    else int(math.ceil(th_per_ch))
OverflowError: Python int too large to convert to C long

----------------------------------------------------------------------
```
The fix is to take care the threshold value before performing further operations.

## Types of changes

## Checklist:


iizukak:
I think this PR must be merged in v0.4.0

dlk/python/dlk/core/optimizer.py
```diff
@@ -355,6 +355,10 @@ def pass_compute_thresholds(graph: Graph) -> None:
                     bn_nega_idx = [v for v in range(len(bn_scale)) if bn_scale[v] < 0]
             threshold = (trans_th['data'] * np.float64(n)) / (np.float64(max_v) * scaling_factor)
 
+            # take care of threshold values that are larger than 13-bit signed integer
```
ruimashita: This comment is wrong, 12 bit is correct.

https://github.com/blue-oil/blueoil/blob/7f704a253df7dadec6be05bda620f68009f3932f/dlk/python/dlk/core/optimizer.py#L337-L338

nlpng: Thanks for the comment, I thought a 13 bit signed integer has max value of 2^12 - 1 for the sign-bit being the highest bit.

ruimashita: Thanks, Aha, you are right!!
And I just remember, I wrote same comment in previous PR. sorry ~

Issue #173, opened 2019-02-27
Concat network compile error by invalid type


ruimashita:
When run `make lm_x86` in generated project with double concat network, the error comes.

```
...
g++ -I./include -O3 -std=c++0x -DUSE_PNG -pthread -g -c mains/main.cpp -o mains/main.o -DFUNC_TIME_MEASUREMENT
src/network.cpp: In member function ‘bool Network::run(float*, float*)’:
src/network.cpp:312:106: error: cannot convert ‘float*’ to ‘uint8_t* {aka unsigned char*}’ in initialization
   QUANTIZED_NOT_PACKED *second_concat_inputs[] = { first_QTZ_linear_mid_tread_half, second_conv2d_Conv2D };
                                                                                                          ^
src/network.cpp:351:86: error: cannot convert ‘float*’ to ‘uint8_t* {aka unsigned char*}’ in initialization
   QUANTIZED_NOT_PACKED *third_concat_inputs[] = { second_concat, third_conv2d_Conv2D };
                                                                                      ^
src/network.cpp:371:96: error: cannot convert ‘uint8_t* {aka unsigned char*}’ to ‘T_FLOAT* {aka float*}’ for argument ‘1’ to ‘void func_Conv2D(T_FLOAT*, T_FLOAT*, T_FLOAT*, convolution_parameters, T_UINT, T_UINT, T_UINT)’
 func_Conv2D(third_concat, fourth_conv2d_kernel, fourth_conv2d_Conv2D, Conv2D_struct, 32, 32, 10);

Makefile:231: recipe for target 'src/network.o' failed
make: *** [src/network.o] Error 1
make: *** Waiting for unfinished jobs....
...
```

git branch: https://github.com/ruimashita/blueoil/tree/double_concat

network file: https://github.com/ruimashita/blueoil/blob/double_concat/lmnet/lmnet/networks/classification/double_concat.py
config file: https://github.com/ruimashita/blueoil/blob/double_concat/lmnet/configs/double_concat.py

part of generated network.cpp:
```
  float *second_conv2d_Conv2D = first_conv2d_Conv2D;

                    Conv2D_struct.input_height = 32;
                    Conv2D_struct.input_width = 32;
                    Conv2D_struct.kernel_height = 3;
                    Conv2D_struct.kernel_width = 3;
                    Conv2D_struct.kernel_depth = 32;
                    Conv2D_struct.kernel_elements = 288;
                    Conv2D_struct.output_channels = 32;
[...]

PR #175, opened 2019-03-01
Fix check for activation quantizer counts of input in optimization pass

nlpng:
## Motivation and Context

This PR is trying to hot fix the issue of #173

## Description

Main change is in the `core/optimizer.py`.
This hot fix resolves following graph that with multiple concat ops:
screen shot 2019-03-01 at 9 54 00

## How has this been tested?

Manually tested with @n-nez and @antonionevado collaboratively

## Types of changes

## Checklist:


n-nez:
> Thanks. why we had the condition ? What we assumed to avoid by the condition ?

@antonionevado

antonionevado:
@ruimashita @n-nez That condition is wrong. It was there to check that every input is quantized. But this situation proved that this way of checking is not good.

antonionevado:
Oh ok. Actually we need to replace the old condition by something else.
@nlpng could you please add a condition that check if `qtzs` are all compatible?
In case they **are compatible** then this should happen:
```
quant_details[m.name] = qtzs
```
but if they are **not compatible** then it should happen:
```
quant_details[m.name] = []
```

n-nez:
@antonionevado, @nlpng 
Would prefer to have a warning when we reject to quantize something because of `qtzs` is not compatible, as it sounds somewhat suspicious... :thinking:

iizukak:
@ruimashita Please check this PR

ruimashita:
I will review tomorrow ~

PR #179, opened 2019-03-04
[Semantic Segmentation] Fix run.py execution failure

iizukak:
## Motivation and Context

In `master` branch, semantic segmentation is fail to execute `run.py` .
the `color_maps` list should be contain RGB colors. 

[Here](https://gist.github.com/iizukak/f3a9685a95941b325e6d2765d3fe372b) is the error message.

## How has this been tested?

Execute run.py in output_template manually.

## Types of changes

## Checklist:

### DIFF
--- a/output_template/python/lmnet/utils/output.py
+++ b/output_template/python/lmnet/utils/output.py
@@ -181,26 +181,18 @@ class ImageFromJson():
     """Create callable instance to return list of tuple (file_name, PIL image object) from prediction json."""
 
     color_maps = [
-        "#1f77b4",
-        "#aec7e8",
-        "#ff7f0e",
-        "#ffbb78",
-        "#2ca02c",
-        "#98df8a",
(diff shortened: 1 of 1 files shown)


iizukak:
Here is example of output of fixed run.py.
test4

output_template/python/lmnet/utils/output.py
```diff
@@ -181,26 +181,18 @@ class ImageFromJson():
...
+        [192, 192, 128],
+        [128, 64, 128],
+        [60, 40, 222],
+        [128, 128, 0],
+        [192, 128, 128],
+        [64, 64, 128],
+        [64, 0, 128],
+        [64, 64, 0],
+        [0, 128, 192],
+        [0, 0, 0],
```
ruimashita: The `color_map` should be the same as `label_colors` of dataset class, I think...

Issue #185, opened 2019-03-06
High CPU and Memory use by DatasetIterator class


ananno:
During dataset creation,
1. 8 processes is used for multi process dataset prefetch
> This will cause issue with CPU having less than 8 core. If the machine doesn't have enough cpu core available system can be frozen for long time.
2. A prefetch queue of 200 is used. For dataset with large images this number of queue can increase the RAM use significantly.

ruimashita:
@ananno 
Thanks.
I agree. process size and queue size should be able to custom. And default queue size should be smaller.

PR #186, opened 2019-03-06
Fix dataset iterator to limit CPU and RAM usage

ananno:
## Motivation and Context

1. Running in machine with 8 core (hard-coded), I found that the dataset iterator usase all 8 core and the system becomes non-responsive for some time. It should also add CPU over head if the system has less than 8 core.
2. The queue used in `DatasetIterator` class has fixed size of 200. Large dataset with big images (e.g. 800x480) will need huge RAM to load 200 data at a time. With a lower system RAM it will use swap memory which will further decrease the performance.

Issue : https://github.com/blue-oil/blueoil/issues/185

## Description

In the changes:
1. The number of process is determined by half of the available CPU or 1 in case of single core machine for dataset iterator multi processing. The number of process is kept by `self.num_proc` variable for reuse in initialization of `_MultiProcessDatasetPrefetchThread` class in  file `blueoil/lmnet/lmnet/dataset/dataset_iterator.py`.
2. A new settings `DATASET.BATCH_PRE_FETCH` has been added in config file and passed from the `train.py -> DatasetIterator` class to determine the queue size based on how many batches of data need to be prefetched. 
> Default value is set as 3

## How has this been tested?

The code fix was tested by running training script `blueoil/lmnet/executor/train.py` manually in `lm-server-05`
## Types of changes

## Checklist:

### DIFF
--- a/lmnet/configs/example/object_detection.py
+++ b/lmnet/configs/example/object_detection.py
@@ -136,3 +136,5 @@ DATASET.AUGMENTOR = Sequence([
     Hue((-10, 10)),
     SSDRandomCrop(min_crop_ratio=0.7),
 ])
+DATASET.ENABLE_PREFETCH = True
+DATASET.NUM_PARALLEL_PROCESS = 16
--- a/lmnet/executor/train.py
+++ b/lmnet/executor/train.py
@@ -40,7 +40,10 @@ def setup_dataset(config, subset, rank):
     dataset_kwargs = dict((key.lower(), val) for key, val in config.DATASET.items())
     dataset = DatasetClass(subset=subset, **dataset_kwargs)
     enable_prefetch = dataset_kwargs.pop("enable_prefetch", False)
-    return DatasetIterator(dataset, seed=rank, enable_prefetch=enable_prefetch)
+    num_parallel_process = dataset_kwargs['num_parallel_process']
+    return DatasetIterator(dataset, seed=rank,
+                           enable_prefetch=enable_prefetch,
+                           num_parallel_process=num_parallel_process)
 
 
 def start_training(config):
--- a/lmnet/lmnet/datasets/dataset_iterator.py
+++ b/lmnet/lmnet/datasets/dataset_iterator.py
@@ -13,13 +13,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # =============================================================================
-from multiprocessing import Pool
+import os
 import time
+import queue
 import threading
 import numpy as np
-import os
-import queue
 import tensorflow as tf
+
+from multiprocessing import Pool
+from multiprocessing import cpu_count
 from lmnet.datasets.tfds import TFDSBase
 from lmnet.datasets.base import SegmentationBase, ObjectDetectionBase
 
@@ -91,12 +93,13 @@ def _xorshift32(r):
 
 
 class _MultiProcessDatasetPrefetchThread(threading.Thread):
-    def __init__(self, dataset, result_queue, seed):
+    def __init__(self, dataset, result_queue, seed, num_parallel_process):
         super().__init__()
         # TODO(tokunaga): the number of processes should be configurable
         self.seed = seed + 1  # seed must not be 0 because using xorshift32.
         self.support_getitem = hasattr(dataset, "__getitem__")
-        self.pool = Pool(processes=8, initializer=_prefetch_setup,
+        self.num_parallel_process = num_parallel_process
+        self.pool = Pool(processes=self.num_parallel_process, initializer=_prefetch_setup,
                          initargs=(dataset, self.seed, not self.support_getitem))
         self.result_queue = result_queue
         self.batch_size = dataset.batch_size
@@ -132,7 +135,7 @@ class _MultiProcessDatasetPrefetchThread(threading.Thread):
     def refresh_pool(self):
         self.pool.close()
         self.seed += 1
-        self.pool = Pool(processes=8, initializer=_prefetch_setup,
+        self.pool = Pool(processes=self.num_parallel_process, initializer=_prefetch_setup,
                          initargs=(self.dataset, self.seed, not self.support_getitem))
 
     def loop_body(self):
@@ -229,18 +232,23 @@ class DatasetIterator:
     available_subsets = ["train", "train_validation_saving", "validation"]
 
     """docstring for DatasetIterator."""
-    def __init__(self, dataset, enable_prefetch=False, seed=0):
+    def __init__(self, dataset, enable_prefetch=False, seed=0, queue_size=None, num_parallel_process=None):
         self.dataset = dataset
         self.enable_prefetch = enable_prefetch
         self.seed = seed
+        num_parallel_process = num_parallel_process if num_parallel_process else max(int(cpu_count() // 2), 1)
+        queue_size = queue_size if queue_size else self.dataset.batch_size * num_parallel_process
 
         if issubclass(dataset.__class__, TFDSBase):
             self.enable_prefetch = False
             self.reader = _TFDSReader(self.dataset)
         else:
             if self.enable_prefetch:
-                self.prefetch_result_queue = queue.Queue(maxsize=200)
-                self.prefetcher = _MultiProcessDatasetPrefetchThread(self.dataset, self.prefetch_result_queue, seed)
+                self.prefetch_result_queue = queue.Queue(maxsize=queue_size)
+                self.prefetcher = _MultiProcessDatasetPrefetchThread(self.dataset,
+                                                                     self.prefetch_result_queue,
+                                                                     seed,
+                                                                     num_parallel_process)
                 self.prefetcher.start()
                 print("ENABLE prefetch")
             else:


lmnet/executor/train.py
```diff
@@ -40,7 +40,8 @@ def setup_dataset(config, subset, rank):
     dataset_kwargs = dict((key.lower(), val) for key, val in config.DATASET.items())
     dataset = DatasetClass(subset=subset, **dataset_kwargs)
     enable_prefetch = dataset_kwargs.pop("enable_prefetch", False)
-    return DatasetIterator(dataset, seed=rank, enable_prefetch=enable_prefetch)
+    cached_batch_count = config.DATASET.BATCH_PRE_FETCH
```
ruimashita: Please use the same variable name, not `BATCH_PRE_FETCH` or `cached_batch_count`.

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -79,12 +79,14 @@ def _xorshift32(r):
...
 
 class _MultiProcessDatasetPrefetchThread(threading.Thread):
-    def __init__(self, dataset, result_queue, seed):
+    def __init__(self, dataset, result_queue, seed, queue_size=200):
         super().__init__()
         # TODO(tokunaga): the number of processes should be configurable
         self.seed = seed + 1  # seed must not be 0
         self.support_getitem = hasattr(dataset, "__getitem__")
-        self.pool = Pool(processes=8, initializer=_prefetch_setup,
+        self.num_proc = max(cpu_count() // 2, 1)
```
ruimashita: Why do you use half of cpu count ? It is better to custom this number by user.

ruimashita:
@ananno 
Thanks PR 😄😄 Plz check my comment.

lmnet/executor/train.py
ananno: @ruimashita San: Fixed in the next commit

lmnet/lmnet/datasets/dataset_iterator.py
ananno: @ruimashita San: Fixed in the next commit.
I've created a config variable DATASET.NUM_PARALLEL_PROCESS and use this for number of processes. The default is half of CPU cores. I choose this value considering it is always good to have some cpu resources available for system processes to avoid system freezing.

ananno:
@ruimashita San:
Please review my updates. I've taken care of the issues you mentioned.

lmnet/executor/train.py
```diff
@@ -40,7 +40,12 @@ def setup_dataset(config, subset, rank):
     dataset_kwargs = dict((key.lower(), val) for key, val in config.DATASET.items())
     dataset = DatasetClass(subset=subset, **dataset_kwargs)
     enable_prefetch = dataset_kwargs.pop("enable_prefetch", False)
-    return DatasetIterator(dataset, seed=rank, enable_prefetch=enable_prefetch)
+    batch_pre_fetch = config.DATASET.BATCH_PRE_FETCH
```
ruimashita: Maybe, the line raise error, if config don't have `DATASET.BATCH_PRE_FETCH`. Is it tested other config file?

lmnet/executor/train.py
```diff
@@ -40,7 +40,12 @@ def setup_dataset(config, subset, rank):
...
+    num_parallel_process = config.DATASET.NUM_PARALLEL_PROCESS
```
ruimashita: also here

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -220,16 +221,22 @@ class DatasetIterator:
     available_subsets = ["train", "train_validation_saving", "validation"]
 
     """docstring for DatasetIterator."""
-    def __init__(self, dataset, enable_prefetch=False, seed=0):
+    def __init__(self, dataset, enable_prefetch=False, seed=0, batch_pre_fetch=3, num_parallel_process=1):
```
ruimashita: About `batch_pre_fetch`, In my feeling, the default `3` odd number looks strange 🤔 Why do you not use `4` ? Is it better setting for your dataset ? (This comment is not request to change)

ruimashita: About `num_parallel_process=1` , It is better to be number of CPU core as default. Or as thinking all most core has double logical core (using hyper-threading), to be half number of `cpu_count()` as default. 🤔

lmnet/executor/train.py
ananno: @ruimashita San:
I got your point. I thought all the config variables are mandatory. Anyway, I've handled it by providing default values if unavailable. Thank you for pointing out.

lmnet/executor/train.py
ananno: Same as above

lmnet/lmnet/datasets/dataset_iterator.py
ananno: This number will define the queue size resulting the RAM usage. The number 3 is arbitrarily chosen which is suitable for lm-server5 for running 2 training at a time.

ananno: @ruimashita San:
I prefer half of the `cpu_count()` as it will ensure having some free cpu resources for system processes or others even for the lower end machine having 2 or 4 cores.

ananno:
@ruimashita San:
I've commit a fix to your suggestion. Please review.

lmnet/executor/train.py
```diff
@@ -40,7 +41,12 @@ def setup_dataset(config, subset, rank):
     dataset_kwargs = dict((key.lower(), val) for key, val in config.DATASET.items())
     dataset = DatasetClass(subset=subset, **dataset_kwargs)
     enable_prefetch = dataset_kwargs.pop("enable_prefetch", False)
-    return DatasetIterator(dataset, seed=rank, enable_prefetch=enable_prefetch)
+    batch_pre_fetch = dataset_kwargs.get('batch_pre_fetch', 3)      # Default is arbitrarily set to 3, can be modified based on availabel RAM
```
ruimashita: The default value setting of `batch_pre_fetch` and `num_parallel_process` should be in `DatasetIterator`, because if the default value setting is outside of class, It takes cost to read and understand the code. Plz change 👍

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -220,16 +221,22 @@ class DatasetIterator:
...
         self.dataset = dataset
         self.enable_prefetch = enable_prefetch
         self.seed = seed
+        self.batch_pre_fetch = batch_pre_fetch
 
         if self.enable_prefetch:
-            self.prefetch_result_queue = queue.Queue(maxsize=200)
+            queue_size = self.batch_pre_fetch * self.dataset.batch_size
```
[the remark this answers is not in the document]
ananno: @ruimashita San: 
Hmmm, I misunderstood too than. In that case the queue size should be same as batch size and no need for config in that case.

I've pushed commit with the fix.

lmnet/executor/train.py
ananno: @ruimashita San:
> No need for `batch_pre_fetch` value as queue is the batch size data.

To set `num_parallel_process` default value in `DatasetIterator` I've added `num_parallel_process` attribute to the `Base` class of the dataset class. Through dataset class attribute `num_parallel_process` config variable can be read and set in class initializer `DatasetIterator.__init__`.

ananno:
@ruimashita San:
Please review the suggested changes.

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -89,12 +90,13 @@ def _xorshift32(r):
 
 
 class _MultiProcessDatasetPrefetchThread(threading.Thread):
-    def __init__(self, dataset, result_queue, seed):
+    def __init__(self, dataset, result_queue, seed, num_parallel_process=1):
```
ruimashita: no need default value

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -208,8 +211,17 @@ def __init__(self, dataset, enable_prefetch=False, seed=0):
         self.seed = seed
 
         if self.enable_prefetch:
-            self.prefetch_result_queue = queue.Queue(maxsize=200)
-            self.prefetcher = _MultiProcessDatasetPrefetchThread(self.dataset, self.prefetch_result_queue, seed)
+            queue_size = self.dataset.batch_size
+            if self.dataset.num_parallel_process and self.dataset.num_parallel_process > 0:
```
ruimashita: Please set num_process default value in DatasetItelator’s  __init__() .

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -208,8 +211,17 @@ def __init__(self, dataset, enable_prefetch=False, seed=0):
         self.seed = seed
 
         if self.enable_prefetch:
-            self.prefetch_result_queue = queue.Queue(maxsize=200)
-            self.prefetcher = _MultiProcessDatasetPrefetchThread(self.dataset, self.prefetch_result_queue, seed)
+            queue_size = self.dataset.batch_size
```
ruimashita: queue size is not related to barch_size. Please change.

ananno: @ruimashita 
I've updated queue size. Updated code for read from dataset config. if set it will use it or set to 50

lmnet/lmnet/datasets/dataset_iterator.py
ananno: @ruimashita 
Set default value of num_parallel_process in init(). If set in dataset config it will use it or else set to half of the available cpu count. The config is not mandatory.

ananno:
@ruimashita 
I've updated the code. Please review.

lmnet/lmnet/datasets/dataset_iterator.py
ananno: @ruimashita 
I've removed it.

lmnet/lmnet/datasets/dataset_iterator.py
ruimashita: @ananno 
Sorry I mistaken. 😭  The `num_process` default value is set in DatasetItelator’s init() .

lmnet/configs/example/object_detection.py
```diff
@@ -14,8 +14,8 @@
 # limitations under the License.
 # =============================================================================
 from easydict import EasyDict
+from multiprocessing import cpu_count
```
ruimashita: No need this changes

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -94,7 +95,9 @@ def __init__(self, dataset, result_queue, seed):
         # TODO(tokunaga): the number of processes should be configurable
         self.seed = seed + 1  # seed must not be 0 because using xorshift32.
         self.support_getitem = hasattr(dataset, "__getitem__")
-        self.pool = Pool(processes=8, initializer=_prefetch_setup,
+        self.num_parallel_process = dataset.num_parallel_process \
```
ruimashita: `num_parallel_process` is already decided in `DatasetIterator.__init()__`, no need to change in here.
Please change to `def __init__(self, dataset, result_queue, seed, num_parallel_process)` , 
and `self.num_parallel_process = num_parallel_process` in `_MultiProcessDatasetPrefetchThread.__init__()`

ananno:
@ruimashita 
Sorry for the overlook. I hope it is OK this time.

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -206,10 +210,18 @@ def __init__(self, dataset, enable_prefetch=False, seed=0):
         self.dataset = dataset
         self.enable_prefetch = enable_prefetch
         self.seed = seed
+        queue_size = dataset.queue_size if dataset.queue_size else 50
```
ruimashita: `queue_size` and `num_parallel_process` are only for `DatasetIterator` and `_MultiProcessDatasetPrefetchThread`, `Dataset` class shouldn't have these `queue_size` and `num_parallel_process` parameter 🤔🤔🤔 
Could you change these to according to encapsulation ways.

ruimashita:
@ananno 
> Sorry for the overlook. I hope it is OK this time.
Don't mind ❗️ I'm so happy for your contribution 👍👍

lmnet/lmnet/datasets/dataset_iterator.py
ananno: @ruimashita 
Sorry, I completely misunderstood your point. 
I fixed it.

lmnet/lmnet/datasets/dataset_iterator.py
ananno: @ruimashita 
Sorry, I was confused by the term `encapsulation` and where should I use it. Anyway, I've fixed as suggested by adding parameters in the DatasetIterator.__init__()

lmnet/configs/example/object_detection.py
```diff
@@ -15,7 +15,6 @@
 # =============================================================================
 from easydict import EasyDict
 import tensorflow as tf
-
```
ruimashita: Please don't delete line.
See also
https://www.python.org/dev/peps/pep-0008/#id23

lmnet/lmnet/datasets/dataset_iterator.py
ruimashita: @ananno 
Oh, No. You are not wrong. Sorry for my poor description 😭😭 
Thanks for contribution

lmnet/executor/train.py
```diff
@@ -18,8 +18,9 @@
 
 import click
 import tensorflow as tf
-from tensorflow.core.util.event_pb2 import SessionLog
 
+from multiprocessing import cpu_count
```
ruimashita: Please change import order following pep8 👍 
https://www.python.org/dev/peps/pep-0008/#id23

ruimashita:
ok to test

ruimashita:
@ananno 
Hi, There are no changes recently.

Could you fix this test error ?
```
    def test_train():
        """Test only didn't get error"""
    
        config_file = "tests/fixtures/configs/for_train.py"
        expriment_id = "for_test"
>       run(None, None, config_file, expriment_id, recreate=True)

tests/executor_tests/test_train.py:30: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
executor/train.py:382: in run
    start_training(config)
executor/train.py:83: in start_training
    train_dataset = setup_dataset(config, "train", rank)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

config = {'BATCH_SIZE': 2, 'CLASSES': ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'], 'DATASET': {'BATCH_SIZE': 2, 'DATA_FO...)], 'SAVE_STEPS': 1, 'SUMMARISE_STEPS': 100, 'TASK': <Tasks.CLASSIFICATION: 'IMAGE.CLASSIFICATION'>, 'TEST_STEPS': 100}
subset = 'train', rank = 0

    def setup_dataset(config, subset, rank):
        DatasetClass = config.DATASET_CLASS
        dataset_kwargs = dict((key.lower(), val) for key, val in config.DATASET.items())
        dataset = DatasetClass(subset=subset, **dataset_kwargs)
        enable_prefetch = dataset_kwargs.pop("enable_prefetch", False)
        queue_size = dataset_kwargs.pop("queue_size", None)
        num_parallel_process = dataset_kwargs.pop("num_parallel_process", None)
        return DatasetIterator(dataset, seed=rank,
                               enable_prefetch=enable_prefetch,
                               queue_size=queue_size,
>                              num_parallel_threads=num_parallel_process)
E       TypeError: __init__() got an unexpected keyword argument 'num_parallel_threads'

executor/train.py:48: TypeError
```

lmnet/configs/example/object_detection.py
ananno: I've fixed it

lmnet/executor/train.py
ananno: I've fixed it.

ananno:
I've successfully run the `test_train()`. Please review.

ananno:
This feature implemented separately.

PR #212, opened 2019-04-03
[LMBiSeNet] implement densenet block and conv_bn_act block

ruimashita:
## Motivation and Context

For new semantic segmentation model `LMBiSeNet` which is composed by `densenet` block and `conv bn act` block in context part. 
I implement `densenet` block and `conv_bn_act` block

## Types of changes

## Checklist:


lmnet/lmnet/blocks.py
```diff
@@ -126,3 +128,192 @@ def lmnet_block(
...
+        batch_norm_decay (float): Batch norm decay rate.
+        enable_detail_summary (bool): Flag for summarize feature maps for each operation on tensorboard.
+        data_format (string):  Fromat for inputs data. NHWC or NCHW.
+    Returns:
+        tf.Tensor: Output of current  block.
+    """
+
+    with tf.variable_scope(name):
+        x = inputs
+        for i in range(0, num_layers):
```
iizukak: `0` of range is needed?

lmnet/lmnet/blocks.py
```diff
@@ -126,3 +128,192 @@ def lmnet_block(
...
+
+        output = tf.concat([inputs, output_3x3], axis=concat_axis)
+
+        if enable_detail_summary:
+            tf.summary.histogram('output', output)
+
+    return output
+
+
+def densenet(
```
iizukak: `densnet` looks like name of the network.
In the original paper, authers called `dense block` ?

lmnet/lmnet/blocks.py
ruimashita: Thanks, I will change

lmnet/lmnet/blocks.py
ruimashita: Yes, the `densenet()` is `Dense block` in original paper.

Which one is best ?
1. `densenet()`,  `mobilenetv2()`
2. `dense_block()`, `mobilev2_block()`
3. `dense()`, `mobilev2()` 

No. 2 is not good for me, the function is obviously block, because this function is under `blocks.py`, It is duplicate.
No. 3 `dense` is confuse with `fully connect` (some library name `dense` as `fully connect`)  .
That's why I choose No.1

But just now, I think the name `densenet_group()` is better than `densenet()`. Because
most famous `Residual block` abstraction level is same as current `_dense_layer_conv_bn_act` . `Residual block` and  `_dense_layer_conv_bn_act()` are composed from some convolution and batch norm, etc. `densenet()` are composed from some the `_dense_layer_conv_bn_act()` , It's abstraction level is difference from `Residual block`.

iizukak: > No. 2 is not good for me, the function is obviously block, because this function is under blocks.py, It is duplicate.

OK. There is `lmnet_block` in this script. Thats why I commented.

`densenet_group` looks good name.

lmnet/lmnet/blocks.py
ruimashita: Fixed in https://github.com/blue-oil/blueoil/pull/212/commits/1260d4f50a1d3850cedfdc6929cabce496f3f19e

lmnet/lmnet/blocks.py
ruimashita: Fixed in https://github.com/blue-oil/blueoil/pull/212/commits/1260d4f50a1d3850cedfdc6929cabce496f3f19e

ruimashita:
@iizukak @nlpng Review please 👍

PR #217, opened 2019-04-08
Implement packed data type

primenumber:
Implement new template class for quantized packed data.

## Motivation and Context

In current implementation of master, quantized packed data type is `int32_t`, so easily confused with other types.
This PR implement template class for quantized packed data in order to prevent confusing data.

We will add some hardwares' support.
This may increase kinds of packing size(16bit, 32bit, 64bit and etc.) and data format(CHW, HWC and etc.).
If we use current implementation, we will easily confuse data format.
This is because quantized packed data types are all `QUANTIZED_PACKED`, this type hasn't data format information.
Furthermore, `QUANTIZED_PACKED` doesn't support different packing size.
This PR also make available packed data types with different packing size and data format.
Implicit type conversion between one type to another types with different packing size or data format is disabled.

I will change type for packed kernel weight in other PR (now `T_UINT`).

## Description

I implement `QuantizedPacked` template class for packed data types.

In order to reduce changes, define type alias `QUANTIZED_PACKED`.

For convenience, I implement XOR for `QuantizedPacked`.

## How has this been tested?

```shell
$ PYTHONPATH=python/dlk python python/dlk/scripts/generate_project.py -I examples/object_detection/fyolo_quantize_4_v4/minimal_graph_with_shape.pb -o tmp/ -p object_detection -hq -ts
$ cd tmp/object_detection
$ make -j8 lm_x86
$ make clean
$ make -j8 lm_arm
$ make clean
$ make -j8 lm_aarch64
$ make clean
$ make -j8 lm_fpga
``` 

## Types of changes

## Checklist:


antonionevado:
I like the idea of having more information about the datatype like the bitwidth! Thank you!

Only mention that I see the data layout more like a property of the tensor than the type itself, no? I mean, maybe having a quantized packed type which have  bitwidth as a property and also a quantized tensor type that have the data layout as a property is better?

When we define something like:
QUANTIZED_PACKED n = tensor[idx];
What is the meaning of the layout if ‘n’ is only part of a tensor?

I misunderstood the implementation?

iizukak:
@ruimashita @n-nez Please Check this PR.

dlk/python/dlk/templates/include/global.tpl.h
```diff
@@ -30,12 +30,32 @@ typedef int16_t   T_INT16;
 
 #define QUANTIZED_NOT_PACKED uint8_t
 
+enum class MemoryFormat {
+  HWC,
+};
+
+template <typename pack_type, MemoryFormat memory_format>
+class QuantizedPacked {
```
n-nez: Agreed with @antonionevado, `MemoryFormat` sounds more as an attribute of the container not packed type. Furthermore, at the current moment, you do not use it. So would suggest removing it at least for now. Then later if it happens that it is handy to tag packed type with we could easily introduce it again.

n-nez: Also would put `__attribute__ ((packed))` here to be sure that `QuantizedPacked` is always the same thing as `pack_type` in terms of in-memory representation.

dlk/python/dlk/templates/src/func/impl/arm_neon/quantized_conv2d_tiling.cpp
```diff
@@ -302,10 +302,10 @@ void QuantizedConv2DTiling(QUANTIZED_NOT_PACKED input[],
                   for (unsigned int kc = 0; kc < kw; ++kc) {
                     uint32x4_t nk = vld1q_u32(&notk[kr][kc][0]);
                     uint8x16_t nk8 = vreinterpretq_u8_u32(nk);
-                    uint32x4_t in = vdupq_n_u32(in_tile[row + kr][col + kc][0]);
+                    uint32x4_t in = vdupq_n_u32(static_cast<QUANTIZED_PACKED::T>(in_tile[row + kr][col + kc][0]));
```
n-nez: This cast all the way is a bit heavy to comprehend actually.
Wonder could we just add something like
```cpp
explicit T Raw() const { return val; }
```
Into the class.

And write such code as
```
uint32x4_t in = vdupq_n_u32(in_tile[row + kr][col + kc][0].Raw());
```
instead?

As it seems there will be a lot of it. :thinking:

dlk/python/dlk/templates/src/matrix/generic/quantized_multiplication.cpp
```diff
@@ -31,7 +31,8 @@ static int pop_count(T_UINT i) {
   return (((i + (i >> 4)) & 0x0F0F0F0F) * 0x01010101) >> 24;
 }
 
-#define CONV(i, k) r##i##k += pop_count(~(a ^ b##k##0)) + 2 * pop_count(~(a ^ b##k##1)) - 3 * (pop_count(~a));
+#define CONV(i, k) r##i##k += pop_count(~(a ^ static_cast<QUANTIZED_PACKED::T>(b##k##0))) \
+                              + 2 * pop_count(~(a ^ static_cast<QUANTIZED_PACKED::T>(b##k##1))) - 3 * (pop_count(~a));
```
n-nez: How about making `pop_count()` and `operator~()` to be a member of `QuantizedPacked` or just creating standalone function accepting it.

e.g. (member version)
```
int PopCount() const { return pop_count(val); }
QuantizedPacked<packed_type> operator~() const { return QuantizedPacked<packed_type>(~val); }
```

allows us to simplify code above to
```
#define CONV(i, k) r##i##k += (~(a ^ b##k##0)).PopCount() + 2 * (~(a ^ b##k##1)).PopCount() - 3 * (~a).PopCount()
```

dlk/python/dlk/templates/src/pack_input_to_qwords.cpp
```diff
@@ -60,7 +60,7 @@ void pack_input_to_qwords(QUANTIZED_NOT_PACKED input[],
     if (bit_count == nbit_qinput_word)
```
n-nez: As we trying to be generic and type-safe
How about making `nbit_qinput_word` to be a trait of `QuantizedPacked` class here
I.e.
```cpp
template <typename pack_type>
￼class QuantizedPacked {
...
  static constexpr BitCount = sizeof(pack_type) * 8;
```

```cpp
if (bit_count == QUANTIZED_PACKED::BitCount) {
```

So relation between number of bits in the type and the condition will be more clear

dlk/python/dlk/templates/src/func/impl/quantized_conv2d_dim2col.cpp
```diff
@@ -101,11 +101,11 @@ void binary_convolution_cpu(QUANTIZED_PACKED input_channels[],
 
     for (T_UINT idx_k = 0; idx_k < bin_kernel_nwords; idx_k++) {
       for (T_UINT in_channel = 0; in_channel < num_in_channels; in_channel++) {
-        T_UINT in_data = input_channels[idx_in];
+        QUANTIZED_PACKED in_data = input_channels[idx_in];
 
         for (T_UINT k_pe = 0; k_pe < num_kernels; k_pe++) {
           T_UINT kernel_buf = kernel[k_pe * bin_kernel_nwords + idx_k];
-          T_INT xnor_result = pop_count(~(in_data ^ kernel_buf));
+          T_INT xnor_result = pop_count(~(static_cast<QUANTIZED_PACKED::T>(in_data) ^ kernel_buf));
```
n-nez: Could we make `kernel_buf` to be of type `QUANTIZED_PACKED`
As bitsize of `kernel_buf` should always match the size of `QUANTIZED_PACKED` storing it in `QUANTIZED_PACKED` seems to make sense to me...
It also allows us to rewrite expression above as
```cpp
T_INT xnor_result = (~(in_data ^ kernel_buf)).PopCount();
```
\**assuming we added `operator~` and `PopCount()`*

dlk/python/dlk/templates/src/func/impl/arm_neon/quantized_conv2d_tiling.cpp
```diff
@@ -69,7 +69,7 @@ void pack_input_for_tiling(QUANTIZED_NOT_PACKED input[],
               + row * in_width * in_bitwidth
               + col * in_bitwidth
               + in_bit_ch;
-            output[index] |= (QUANTIZED_PACKED)bit << in_ch_low;
+            output[index] = QUANTIZED_PACKED(static_cast<QUANTIZED_PACKED::T>(output[index]) | (QUANTIZED_PACKED::T)bit << in_ch_low);
```
n-nez: Maybe we could define `operator|=` on `QuantizedPacked` as well and rewrite it somehow like
```cpp
output[index] |= QUANTIZED_PACKED(bit << in_ch_low);
```
? :thinking:

dlk/python/dlk/templates/include/global.tpl.h
```diff
@@ -30,12 +30,32 @@ typedef int16_t   T_INT16;
...
+ public:
+  using T = pack_type;
+  QuantizedPacked() = default;
+  explicit QuantizedPacked(const T val) : val(val) {}
+  explicit operator T() const { return val; }
+  explicit operator T() { return val; }
```
n-nez: having only `const` version of `operator T()` here should be enough.

```
 explicit operator T() const { return val; }
```

could be used in both `QuantizedPacked&` and `const QuantizedPacked&` cases.

primenumber:
@antonionevado @n-nez 
Thank you.
Surely, memory layout information should be hold in tensor type, not packed element type.

dlk/python/dlk/templates/include/global.tpl.h
primenumber: Removed `MemoryFormat`.

primenumber: Add attribute `packed`.

dlk/python/dlk/templates/src/pack_input_to_qwords.cpp
primenumber: Defined `BitCount`.

dlk/python/dlk/templates/src/func/impl/quantized_conv2d_dim2col.cpp
primenumber: Yes, `kernel_buf` should be `QUANTIZED_PACKED`, but this change require much implementation (python code should have information about packed or not).
I will implement in other PR.

n-nez: For now, you probably could just something like below, in this PR, and update python and other things in the other one.
```cpp
auto kernel_buf = QUANTIZED_PACKED(kernel[k_pe * bin_kernel_nwords + idx_k]);
```
:thinking:

dlk/python/dlk/templates/include/global.tpl.h
primenumber: Fixed.

dlk/python/dlk/templates/src/func/impl/arm_neon/quantized_conv2d_tiling.cpp
primenumber: Defined `operator|=`.

dlk/python/dlk/templates/src/func/impl/quantized_conv2d_dim2col.cpp
primenumber: That's right.
Fixed.

dlk/python/dlk/templates/src/matrix/generic/quantized_multiplication.cpp
primenumber: Defined `pop_count` and `operator~`.

dlk/python/dlk/templates/src/func/impl/arm_neon/quantized_conv2d_tiling.cpp
primenumber: Implemented `Raw`.

n-nez:
@primenumber Thank you. LGTM now.

PR #219, opened 2019-04-09
create lmnet config with tempfile

tsawada:
Previously, generate_lmnet_config creates a config file under cwd, which
is not always writable.


blueoil/generate_lmnet_config.py
```diff
@@ -296,10 +297,12 @@ def _save(lmnet_config):
...
 
     applied = tpl.render(lmnet_config)
-    config_file = "{}.py".format(lmnet_config['model_name'])
-    with open(config_file, 'w') as fp:
+    with NamedTemporaryFile(
+            prefix="blueoil_config_{}".format(lmnet_config['model_name']),
+            suffix=".py", delete=False, mode="w") as fp:
         fp.write(applied)
-    return config_file
+        print(fp.name)
```
ruimashita: This line `print(fp.name)` looks only for debug.
Please delete 👍

tsawada: Good catch. thanks!

iizukak:
@ruimashita Check this PR again please.

PR #226, opened 2019-04-11
Use QuantizedPacked for kernel

primenumber:
Introduce QUANTIZED_PACKED_KERNEL type for quantized packed kernel data.

## Motivation and Context

In current implementation of master, quantized packed kernel data type is T_UINT.
T_UINT is easily confused with other data types.
This PR use QuantizedPacked in order to reduce confusing types.

## Description

Input for FPGA has volatile qualifier.
But kernel for FPGA don't have to have volatile qualifier.
So, I introduced QUANTIZED_PACKED_KERNEL type alias.
For convenience, operator^ is generalized for cv qualifier.

C++ version is bumped up to C++14(to infer function return type and use std::remove_cv_t).

Default qword type is changed from Int32 to Uint32 (Because it is only used for QUANTIZED_PACKED, it should be uint32_t).

## How has this been tested?

```shell
$ PYTHONPATH=python/dlk python python/dlk/scripts/generate_project.py -i examples/object_detection/fyolo_quantize_4_v4/minimal_graph_with_shape.pb -o tmp/ -p object_detection -hq -ts
$ cd tmp/object_detection
$ make -j8 lm_x86
$ make clean
$ make -j8 lm_arm
$ make clean
$ make -j8 lm_aarch64
$ make clean
$ make -j8 lm_fpga
```
And run test of lm_* for x86, arm, FPGA (I don't have aarch64 environment).

## Types of changes

## Checklist:


iizukak:
@n-nez @antonionevado Please review 🙇

dlk/python/dlk/core/data_types.py
```diff
@@ -80,6 +82,16 @@ def nptype(cls):
         return np.int8
 
 
+class PackedInt8(Primitive, int):
```
antonionevado: We don't use this, maybe we can deleted it for now?

dlk/python/dlk/core/data_types.py
```diff
@@ -90,6 +102,16 @@ def nptype(cls):
         return np.uint8
 
 
+class PackedUint8(Primitive, int):
```
antonionevado: Not used?

dlk/python/dlk/core/data_types.py
```diff
@@ -100,6 +122,16 @@ def nptype(cls):
         return np.int16
 
 
+class PackedInt16(Primitive, int):
```
antonionevado: Not used.

dlk/python/dlk/core/data_types.py
```diff
@@ -110,6 +142,16 @@ def nptype(cls):
         return np.uint8
 
 
+class PackedUint16(Primitive, int):
```
antonionevado: Not used.

dlk/python/dlk/core/data_types.py
```diff
@@ -120,6 +162,16 @@ def nptype(cls):
         return np.int32
 
 
+class PackedInt32(Primitive, int):
```
antonionevado: Not used.

dlk/python/dlk/core/data_types.py
```diff
@@ -140,6 +202,16 @@ def nptype(cls):
         return np.int64
 
 
+class PackedInt64(Primitive, int):
```
antonionevado: More likely to be used soon in 64 bit architectures, so maybe worth keep it.

dlk/python/dlk/templates/manual/consts/input.tpl.cpp
```diff
@@ -25,13 +25,13 @@ limitations under the License.
 {% if node.transposed_data %}
 
 #if defined(RUN_ON_FPGA)
-{{ node.dtype.cpptype() }} {{ node.name }}[] = {
+Base<{{ node.dtype.cpptype() }}>::type {{ node.name }}_raw[] = {
```
n-nez: How about putting `static` here, to hide implementation? :thinking:

dlk/python/dlk/core/data_types.py
```diff
@@ -150,6 +222,16 @@ def nptype(cls):
         return np.uint64
 
 
+class PackedUint64(Primitive, int):
```
antonionevado: Similar case that PackedInt64

dlk/python/dlk/core/data_types.py
```diff
@@ -150,6 +222,16 @@ def nptype(cls):
...
+    @classmethod
+    def cpptype(cls):
+        return 'QuantizedPacked<uint64_t>'
+
+    @classmethod
+    def nptype(cls):
+        return np.uint64
+
+
```
n-nez: would prefer to not have types which we do not use

it is easy to add missed ones when we start using them and keeping unused code will increase maintenance effort / mental load while reading.

dlk/python/dlk/templates/include/global.tpl.h
```diff
@@ -35,15 +36,17 @@ template <typename pack_type>
...
  public:
   using T = pack_type;
+  using base_t = std::remove_cv_t<T>;
   static constexpr std::size_t BitCount = sizeof(pack_type) * CHAR_BIT;
   QuantizedPacked() = default;
   explicit QuantizedPacked(const T val) : val(val) {}
-  explicit operator T() const { return val; }
-  QuantizedPacked<T>& operator|=(const QuantizedPacked<T>& that) {
+  explicit operator base_t() const { return val; }
+  template <typename U, std::enable_if_t<std::is_same<std::remove_cv_t<T>, std::remove_cv_t<U>>::value, int> = 0>
```
antonionevado: Maybe we can rewrite this as:
```c++
template <typename U, std::enable_if_t<std::is_same<base_t, std::remove_cv_t<U>>::value, int> = 0>
```
to make it easier to read?

dlk/python/dlk/templates/manual/consts/input.tpl.cpp
primenumber: Add `static`.

dlk/python/dlk/core/data_types.py
primenumber: Remove these types except `PackedUint64`.

dlk/python/dlk/templates/include/global.tpl.h
primenumber: Fixed.

primenumber:
@n-nez @antonionevado Fixed these points. Plz re-review.

n-nez:
@primenumber `PackedUint8` seems to be still here. Could you remove it as well? :thinking:

primenumber:
@n-nez Removed `PackedUint8`.

iizukak:
@n-nez Can you check this PR?

n-nez:
Thank you! LGTM!

PR #252, opened 2019-04-25
Tca support prototype

antonionevado:
Add supports for TCAv2.

## Motivation and Context
TCAv2 will be released soon as part of Blueoil.

## Description

**Data layout changes**

Activations (inference: N=1) | Before | After
-- | -- | --
Rank | 3 | 4
Layout | H x W x C | C(b) x H x W x b

Kernels | Before | After
-- | -- | --
Rank | 4 | 6
Layout | O x H x W x C | O(b) x C(b) x H x W x b(O) x b(C)

where:
 - b is the famous TCA parameter.
 - H is height
 - W is width
 - C are input channels
 - O are output channels
 - C(b) stands for the input channel dimension divided in blocks of size b
 - O(b) stands for the output channel dimension divided in blocks of size b
 - b(O) stands for size b following the output channel direction
 - b(C) stands for size b following the input channel direction

**Tensorflow importer (tf.py)**
* 7 new operators to import: Gather, Unique, Cast, Minimum, StridedSlice, Prod, Shape
* 4 of these operators has 2 inputs: Gather, Minimum, Reshape and Prod
* 1 of them has four inputs: StridedSlice
* Some operators has undefined shape (to be inferred at run time)

**View.py**
* Physical addresses of each layer’s kernels are passed to the runtime (we now copy each kernel to the raw memory in advance during the initialization of the runtime)

**Data_types.py**
 * Added missing class which represents QUANTIZED_PACKED data type

**Operators.py**
* Fix reshaped operator: missing shape parameter

**Optimizer.py**
* _transpose_kernels() function is not used anymore (deleted)
* Pass_propagate_quantization_details_into_conv recognizes Lookup as a quantizer
* Pass_pack_weights now change the layout of the TensorFlow kernels to TCAv2 kernel order. It also pads in the output channel direction for the cases where the number of output channels is not a multiple of b. The codification of the kernel values is also inverted.
* pass_quantize_convolutions() now set the data type of the quantized convolutions to QUANTIZED_PACKED.

**C++ (runtime)**
* Time measurement reports for the debugging executable are now printed hierarchically
* pack_input_to_qwords() supports padding on the input channels dimension
* Network.cpp: Init() function now copy all the kernels to the raw memory during initialization time
* Added TCAConv2d() function which changes the layout of the inputs and then calls TCAv2 with de10_nano::RunTCA()
* ConcatV2 and Space2Depth operators now expect quantized inputs
* Quantizer func_QTZ_linear_mid_tread_half() function now outputs quantized values
* func_QuantizedConv2D() function now acceptsQUANTIZED_PACKED() input values

## How has this been tested?
Only for the FPGA target. Confirmed to work on DE10 Nano with TCAv2.

## Types of changes
- [ x ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [ x ] My change requires a change to the documentation.

### DIFF
--- a/dlk/python/dlk/core/data_types.py
+++ b/dlk/python/dlk/core/data_types.py
@@ -260,3 +260,23 @@ class QUANTIZED_NOT_PACKED(Primitive, int):
     @classmethod
     def nptype(cls):
         return np.int8
+
+
+class QUANTIZED_PACKED(Primitive, int):
+    @classmethod
+    def cpptype(cls):
+        return 'QUANTIZED_PACKED'
+
+    @classmethod
+    def nptype(cls):
+        return np.int32
+
+class QUANTIZED_PACKED_KERNEL(Primitive, int):
+    @classmethod
+    def cpptype(cls):
+        return 'QUANTIZED_PACKED_KERNEL'
+
+    @classmethod
+    def nptype(cls):
+        return np.int32
+
--- a/dlk/python/dlk/core/data_types.py
+++ b/dlk/python/dlk/core/data_types.py
@@ -269,7 +269,7 @@ class QUANTIZED_PACKED(Primitive, int):
 
     @classmethod
     def nptype(cls):
-        return np.int32
+        return np.uint32
 
 class QUANTIZED_PACKED_KERNEL(Primitive, int):
     @classmethod
@@ -278,5 +278,5 @@ class QUANTIZED_PACKED_KERNEL(Primitive, int):
 
     @classmethod
     def nptype(cls):
-        return np.int32
+        return np.uint32
 
--- a/dlk/python/dlk/core/optimizer.py
+++ b/dlk/python/dlk/core/optimizer.py
@@ -20,83 +20,11 @@ import numpy as np
 
 from core.graph import Graph
 from core.graph_pattern_matching import get_nodes_in_branch, sort_graph
-from core.operators import Constant, Operator, Conv
-from core.data_types import PackedUint32, QUANTIZED_NOT_PACKED
+from core.operators import Constant, Operator, Conv, Lookup
+from core.data_types import Uint32, Int32, QUANTIZED_NOT_PACKED, QUANTIZED_PACKED, PackedUint32, QUANTIZED_PACKED_KERNEL
 from typing import cast, List, Any
 from collections import defaultdict
 from modules.packer import Packer
-
-
-def _transpose_kernels(kernel_data: np.ndarray,
-                       oh: int,
-                       ow: int,
-                       od: int,
-                       kh: int,
-                       kw: int,
-                       kd: int) -> List[int]:
-    """Calculates and prepares the transposed kernel data in advance.
-
-    Parameters
-    ----------
-    kernel_data : np.ndarray
-        The input data.
-    oh : int
-        output height
-    ow : int
-        output width
-    od : int
-        output depth
-    kh : int
-        kernel height
-    kw : int
-        kernel width
-    kd : int
-        kernel depth
-    """
-    NUM_PE = 16
-    NBIT_QDYPE = 32
-    MAX_NBIT_QINPUT = 2
-    MAX_NBIT_KERNEL = 1
-    num_qinput_per_qword = int(NBIT_QDYPE / MAX_NBIT_QINPUT)
-    num_qkernel_per_qword = int(NBIT_QDYPE / MAX_NBIT_KERNEL)
-    k_c_by_word = int((kd + (num_qkernel_per_qword - 1)) / num_qkernel_per_qword)
-    k_n_aligned_with_num_pe = int(((od + (NUM_PE - 1)) / NUM_PE) * NUM_PE)
-    if od < NUM_PE:
-        k_size = k_n_aligned_with_num_pe * kh * kw * k_c_by_word
-    else:
-        k_size = od * kh * kw * k_c_by_word
-
-    flatten_value = []
-    for elem in kernel_data:
-        flatten_value.extend(elem)
-    while len(flatten_value) != k_size:
-        flatten_value.extend("0")
-
-    copy_value = [0] * k_size
-    for i in range(od * kh * kw * k_c_by_word):
-        copy_value[i] = flatten_value[i]
-
-    transposed_values = [0] * k_size
-    if (od < NUM_PE):
-        kn_out = int(k_n_aligned_with_num_pe / NUM_PE)
-    else:
-        kn_out = int(od / NUM_PE)
-    idx_src = 0
-
-    for no in range(kn_out):
-        for ni in range(NUM_PE):
-            for h in range(kh):
-                for w in range(kw):
-                    for c in range(k_c_by_word):
-                        idx_dst = no * (kh * kw * k_c_by_word * NUM_PE)
-                        idx_dst += h * (kw * k_c_by_word * NUM_PE)
-                        idx_dst += w * (k_c_by_word * NUM_PE)
-                        idx_dst += c * (NUM_PE)
-                        idx_dst += ni
-                        transposed_values[idx_dst] = copy_value[idx_src]
-                        idx_src += 1
-
-    return transposed_values
 
 
 def pass_remove_identities(graph: Graph) -> None:
@@ -251,7 +179,8 @@ def pass_propagate_quantization_details_into_conv(graph: Graph) -> None:
     qtypes = [
         'QTZ_binary_mean_scaling',
         'QTZ_linear_mid_tread_half',
-        'QTZ_binary_channel_wise_mean_scaling'
+        'QTZ_binary_channel_wise_mean_scaling',
+        'Lookup'
     ]
 
     quant_details = defaultdict(list)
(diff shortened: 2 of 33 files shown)


dlk/python/dlk/core/data_types.py
```diff
@@ -260,3 +260,23 @@ def cpptype(cls):
...
+
+
+class QUANTIZED_PACKED(Primitive, int):
+    @classmethod
+    def cpptype(cls):
+        return 'QUANTIZED_PACKED'
+
+    @classmethod
+    def nptype(cls):
+        return np.int32
```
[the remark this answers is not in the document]
antonionevado: It's true, thank you!

dlk/python/dlk/core/data_types.py
```diff
@@ -260,3 +260,23 @@ def cpptype(cls):
...
+
+class QUANTIZED_PACKED_KERNEL(Primitive, int):
+    @classmethod
+    def cpptype(cls):
+        return 'QUANTIZED_PACKED_KERNEL'
+
+    @classmethod
+    def nptype(cls):
+        return np.int32
```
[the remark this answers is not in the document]
antonionevado: It's true, thank you!

antonionevado:
@primenumber this is only a reference implementation, please feel free to adapt and improve the code as much as you want. Many parts could be optimized, the algorithms are simple for make it easy to read. Thank you!

dlk/python/dlk/templates/src/pack_input_to_qwords.cpp
```diff
@@ -12,83 +12,60 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-    }
-  }
+  for (int h = 0; h < input_height; ++h)
+      for (int w = 0; w < input_width; ++w) {
+          for (int d = 0; d < full_words_in_depth; ++d) {
+              output[current_word] = QUANTIZED_PACKED(0);
+              output[current_word + 1] = QUANTIZED_PACKED(0);
+              for (int d = 0; d < bits_per_word; ++d) {
+                for(int b = 0; b < bits_per_input; ++b)
+                  output[current_word + b] = QUANTIZED_PACKED(output[current_word + b].Raw() | ((input[input_index] & (1 << b)) >> b) << d);
```
primenumber: (input[input_index] & (1 << b)) >> b is same as (input[input_index] >> b) & 1.

dlk/python/dlk/templates/src/pack_input_to_qwords.cpp
```diff
@@ -12,83 +12,60 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+          }
+
+          if(!remainder_bits_in_depth)
+              continue;
+
+          output[current_word] = QUANTIZED_PACKED(0);
+          output[current_word + 1] = QUANTIZED_PACKED(0);
+          for (int d = 0; d < remainder_bits_in_depth; ++d) {
+             for(int b = 0; b < bits_per_input; ++b)
+               output[current_word + b] = QUANTIZED_PACKED(output[current_word + b].Raw() | ((input[input_index] & (1 << b)) >> b) << d);
```
primenumber: here, too.

dlk/python/dlk/templates/include/func/extract_image_patches.h
```diff
@@ -18,28 +18,61 @@ limitations under the License.
...
-  T_UINT output_index = 0;
+  const int bits_per_input = 2;
+  const int bits_per_word = sizeof(QUANTIZED_PACKED) * CHAR_BIT;
+  int full_words_in_depth = input_depth / bits_per_word;
+  int remainder_bits_in_depth = input_depth % bits_per_word;
+
+  T_UINT packed_input_depth = full_words_in_depth * bits_per_input + (remainder_bits_in_depth ? bits_per_input : 0);
+  T_UINT packed_output_size = packed_input_depth * kernel_size * kernel_size * out_width * out_height;
 
+  QUANTIZED_PACKED* output_buffer = new QUANTIZED_PACKED[packed_output_size];
```
primenumber: You can use std::make_unique<QUANTIZED_PACKED[]>.

dlk/python/dlk/core/optimizer.py
```diff
@@ -446,23 +376,50 @@ def pass_pack_weights(graph: Graph) -> None:
...
+        padded_data = np.copy(weight_quantizer.data)
+
+        for axis in [0, 3]:
+            pad_tensor = pad_to_multiple_of_b(padded_data, axis, b)
+            if pad_tensor is not None:
+                padded_data = np.append(padded_data, pad_tensor, axis=axis)
+
+        tca_output = np.copy(padded_data)
+        oc, kh, kw, kd = padded_data.shape[:]
+        padded_data = padded_data.flatten()
```
primenumber: Why did you flatten padded_data?

dlk/python/dlk/templates/src/func/quantized_conv2d.cpp
```diff
@@ -167,19 +286,47 @@ void func_QuantizedConv2DWithThreshold(QUANTIZED_NOT_PACKED input[],
...
                                        binary_convolution_parameters p) {
   func_QuantizedConv2DWithThreshold(input, kernel, output, scaling_factor[0],
                                     p);
 }
+
+
+void func_Lookup(float *input, QUANTIZED_PACKED_KERNEL *lsb, QUANTIZED_PACKED_KERNEL *msb, QUANTIZED_PACKED* output, int h, int w, int c) {
+
+  int b = 32;
+  int packed_depth = 2;
```
primenumber: What is this?

dlk/python/dlk/templates/include/global.tpl.h
```diff
@@ -143,6 +144,21 @@ struct Base<QuantizedPacked<T>> {
...
+
+// TCA
+
+// HPS-TO-FPGA Lightweight bridge address
+constexpr size_t HPS_TO_FPGA_LW_BASE = 0xFF200000;
+
+// Half part of phys memory
+constexpr size_t HW_BUFFERS_PHYS_ADDR_BASE = 0x20000000;
+
+void func_Lookup(float *input, QUANTIZED_PACKED_KERNEL *lsb, QUANTIZED_PACKED_KERNEL *msb, QUANTIZED_PACKED* output, int h, int w, int c);
```
primenumber: It should be written in func/lookup.h.

dlk/python/dlk/core/optimizer.py
antonionevado: Because I wanted to access with indices to the array as in here:
```
... = padded_data[idx]
```

primenumber: It seems that padded_data[g * b + d, o, h, p * b + d] (padded_data is not-flatten) is ok.

dlk/python/dlk/templates/src/func/quantized_conv2d.cpp
antonionevado: This is a hard coded part. The input image contains 3 channels, and, after the lookup, the depth will fit in 2 words of 32 bits each (total 30 channels). So:
  - Input depth becomes 30
  - Input (packed depth) becomes 2

primenumber: hardcoded constant should be constexpr value, and should be commented like `// hard coded, not configurable`

dlk/python/dlk/core/optimizer.py
antonionevado: Ah, yes, it is equivalent. For me was easier to keep it same than we used to do in the runtime.

dlk/python/dlk/templates/src/func/quantized_conv2d.cpp
antonionevado: Ah, what I mean is that shouldn't be hard coded in the first place.

dlk/python/dlk/templates/include/global.tpl.h
antonionevado: Yes, true. I moved it to another file.

dlk/python/dlk/templates/src/func/quantized_conv2d.cpp
antonionevado: I mean, it is actually configurable. It could change.

iizukak:
@primenumber @antonionevado 
I want to merge this PR to tca_support branch, Any problem?

dlk/python/dlk/templates/src/func/quantized_conv2d.cpp
primenumber: When you make it configurable? In this PR or another PR?

antonionevado: Maybe it is better to do it later, there are already a lot of changes.
Put everything together and make it work first is maybe easier.

dlk/python/dlk/core/optimizer.py
primenumber: OK.

dlk/python/dlk/templates/src/func/quantized_conv2d.cpp
primenumber: OK.

iizukak:
Thanks~

PR #258, opened 2019-05-08
Fix config overwrite by check & delete if destination exists

### LINKED ISSUE #256: The use gfile to read/write configs is not overwriting the old saved config
I tried to re-train an existing model from `v0.4.0` using `v.0.6.0` But the generated `config.py` was broken. So when I tried to run prediction, it doesn't work anymore.

I think it was caused by the use of gfile when copying a config file doesn't completely overwrite the old from a new one. as it kind of does a characterwise operation. 

https://github.com/blue-oil/blueoil/pull/223

Joeper214:
## Description

Check if destination then delete destination file if exists before gfile copy when generating config files.

- fix #256 

## Types of changes

## Checklist:

### DIFF
--- a/blueoil/blueoil_train.py
+++ b/blueoil/blueoil_train.py
@@ -42,13 +42,16 @@ def save_config_file(config_file, dest_dir):
         gfile.MkDir(dest_dir)
 
     config_file_dest = os.path.join(dest_dir, 'blueoil_config.yaml')
+
+    # HACK: This is for tensorflow bug workaround.
+    # We can remove lines 49-50 once it's been resolved in tensorflow
+    # issue link: https://github.com/tensorflow/tensorflow/issues/28508
     if gfile.Exists(config_file_dest):
         gfile.Remove(config_file_dest)
 
     return gfile.Copy(
         config_file,
-        config_file_dest,
-        overwrite=True
+        config_file_dest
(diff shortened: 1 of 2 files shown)


tsawada:
* Please add a comment that this hack can be removed once tensorflow/tensorflow#28508 is resolved.
* Please remove `overwrite=True` from `gfile.Copy`

Joeper214:
@tsawada @ruimashita I updated it thanks!

blueoil/blueoil_train.py
```diff
@@ -41,10 +41,17 @@ def save_config_file(config_file, dest_dir):
     if not gfile.Exists(dest_dir):
         gfile.MkDir(dest_dir)
 
+    config_file_dest = os.path.join(dest_dir, 'blueoil_config.yaml')
+
+    # HACK: This is for tensorflow bug workaround.
+    # We can remove lines 49-50 once it's been resolved in tensorflow
```
tsawada: It's better to simply say "following 2 lines" without referencing absolute line number like 49-50, because it'll slide if we add/remove lines above.

tsawada:
LGTM. please address one more inline comment I added.

Joeper214:
@tsawada Updated Thanks!

PR #262, opened 2019-05-09
Enable data prefetch for param tuner

nlpng:
## Description

Previously the processes would not terminate cleanly when the `DATASET.ENABLE_PREFETCH` is `True`, and cause the trial cue to halt. This is a quick fix to avoid the problem by only enable the data prefetch for the training of the parameter tuner.

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/executor/tune_ray.py
+++ b/lmnet/executor/tune_ray.py
@@ -135,7 +135,7 @@ def setup_dataset(config, subset, rank):
     dataset_class = config.DATASET_CLASS
     dataset_kwargs = dict((key.lower(), val) for key, val in config.DATASET.items())
     dataset = dataset_class(subset=subset, **dataset_kwargs)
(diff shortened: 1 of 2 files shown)


lmnet/executor/tune_ray.py
```diff
@@ -135,7 +135,7 @@ def setup_dataset(config, subset, rank):
     dataset_class = config.DATASET_CLASS
     dataset_kwargs = dict((key.lower(), val) for key, val in config.DATASET.items())
     dataset = dataset_class(subset=subset, **dataset_kwargs)
-    enable_prefetch = dataset_kwargs.pop("enable_prefetch", False)
+    enable_prefetch = dataset_kwargs.pop("enable_prefetch", False) if subset == 'train' else False
```
iizukak: Why enable_prefetch enable for training set?
Please add comment.

PR #273, opened 2019-05-15
Efficiency improvement of Docker build

odoku:
## Motivation and Context

Fixing the lmnet source code causes Docker build to be executed from the installation of the Python package, so the build takes a long time.

## Description

Modified so that differences in Docker image are not output as much as possible even if the source code is changed.

## How has this been tested?

Make sure that blueoil.sh works properly.

## Types of changes

## Checklist:


hadusam:
@odoku 
Thanks!
I checked the build times of `docker build` on my GPU server.
* Case1: First(No cache) build by `docker build --no-cache`
* Case2: Cache build after changed only lmnet sources

| Tested Branch | Case1 | Case2 |
|:----|:----|:----|
| master | 9m26.789s | 7m30.749s |
| odoku:improve-docker-build | 8m44.740s | **0m1.914s** |

So faster in the case of changed only lmnet sources! 🎉

PR #276, opened 2019-05-17
Add a dataset class for loading Tensorflow Datasets

tfujiwar:
## Motivation and Context
To be able to run training with datasets prepared in TensorFlow Datasets.

## Description
Implemented a dataset class for loading TensorFlow Datasets for classification.

## How has this been tested?
Executed several trainings with different `train_path` options.
- `gs://lmfs-backend-us-west1/home/sawada/tensorflow_datasets/mnist`
- `/home/blueoil/tensorflow_datasets/mnist`
- `/home/blueoil/tensorflow_datasets/cifar10`
- `/home/blueoil/tensorflow_datasets/cifar100`
- `/home/blueoil/tensorflow_datasets/fashion_mnist`
- `/home/blueoil/tensorflow_datasets/tf_flowers`
- `/home/blueoil/tensorflow_datasets/rock_paper_scissors`

Sample config file:
```yaml
task_type: classification

network_name: LmnetV1Quantize

dataset:
  format: TensorFlow Datasets for Classification
  train_path: gs://lmfs-backend-us-west1/home/sawada/tensorflow_datasets/mnist
  test_path:

trainer:
  batch_size: 10
  epochs: 20
  optimizer: Adam
  learning_rate_schedule: constant
  initial_learning_rate: 0.01

network:
  quantize_first_convolution: yes

common:
  image_size:
    - 32  # height
    - 32  # width

  pretrain_model: false
  dataset_prefetch: true
```

Please be sure to be able to access the GCS bucket. For example, you need to generate a credential file by `gcloud` command before executing training.

```sh
gcloud auth application-default login
python blueoil/blueoil_train.py -c /home/blueoil/config/test.yml -i test
```

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/lmnet/datasets/tensorflow_datasets.py
+++ b/lmnet/lmnet/datasets/tensorflow_datasets.py
@@ -51,20 +51,11 @@ class TensorFlowDatasetsBase(Base):
         self._init_available_splits()
         self._validate_feature_structure()
 
-        dataset = builder.as_dataset(split=self.available_splits[self.subset])
-        dataset = dataset.repeat()
-        if self.subset == "train":
-            dataset = dataset.shuffle(1024)
-
-        iterator = dataset.make_one_shot_iterator()
-        self.next_element_op = iterator.get_next()
+        self.dataset = builder.as_dataset(split=self.available_splits[self.subset])
+        self._init_features()
 
     def __getitem__(self, i, type=None):
-        # This method ignore the index "i" and returns shuffled element.
-        # Currently the this method is used only for shuffling elements
-        # but elements from TensorFlow Datasets are already shuffled.
-        element = self.session.run(self.next_element_op)
-        return self._reshape_element(element)
+        return self.features[i]
 
     def __len__(self):
         return self.num_per_epoch
@@ -106,9 +97,11 @@ class TensorFlowDatasetsBase(Base):
         """
         raise NotImplementedError()
 
-    def _reshape_element(self, element):
+    def _init_features(self):
         """
-        Reshaping a raw element from TensorFlow Datasets into internal format.
+        Initializing feature variables by using tf.Tensor from datasets.
+        For classification, self.features should be an array of tuple (image, label).
+        For object detection, self.features should be an array of tuple (images, gt_boxes).
         """
         raise NotImplementedError()
 
@@ -136,18 +129,25 @@ class TensorFlowDatasetsClassification(TensorFlowDatasetsBase):
         if not is_valid:
             raise ValueError("Datasets should have \"label\" and \"image\" features.")
 
-    def _reshape_element(self, element):
-        image = element["image"]
-        label = element["label"]
+    def _init_features(self):
+        iterator = self.dataset.make_one_shot_iterator()
+        next_element = iterator.get_next()
+        self.features = []
 
-        # Converting grayscale images into RGB images.
-        # This workaround is needed because the method PIL.Image.fromarray()
-        # in pre_prpcessor requires images array to have a shape of (h, w, 3).
-        if image.shape[2] == 1:
-            image = np.stack([image] * 3, 3)
-            image = image.reshape(image.shape[:2] + (3,))
+        with tf.Session() as sess:
+            for _ in range(self.num_per_epoch):
+                element = sess.run(next_element)
+                image = element["image"]
+                label = element["label"]
 
-        label = data_processor.binarize(label, self.num_classes)
-        label = np.reshape(label, (self.num_classes))
+                # Converting grayscale images into RGB images.
+                # This workaround is needed because the method PIL.Image.fromarray()
+                # in pre_prpcessor requires images array to have a shape of (h, w, 3).
+                if image.shape[2] == 1:
+                    image = np.stack([image] * 3, 3)
+                    image = image.reshape(image.shape[:2] + (3,))
 
-        return (image, label)
+                label = data_processor.binarize(label, self.num_classes)
+                label = np.reshape(label, (self.num_classes))
+
+                self.features.append((image, label))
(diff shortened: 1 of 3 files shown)


iizukak:
@tfujiwar Thanks!! I'll check soon.

lmnet/lmnet/datasets/tensorflow_datasets.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    @property
+    def classes(self):
+        return self.info.features["label"].names
+
+    @property
+    def num_classes(self):
+        return self.info.features["label"].num_classes
+
+    def _init_features(self):
```
ruimashita: Do `self.images` and `self.labels` have all of data on memory?
It is better to get a data for each `__getitem__` for big dataset.

lmnet/lmnet/datasets/tensorflow_datasets.py
```diff
@@ -0,0 +1,141 @@
...
+from lmnet import data_processor
+
+
+class TensorFlowDatasetsBase(Base):
+    """
+    Abstract dataset class for loading TensorFlow Datasets.
+    The parameter "extend_dir" should be a path to the prepared data of tfds.
+    "extend_dir" can be a path either in local or in Google Cloud Storage.
+    """
+    available_subsets = ["train", "train_validation_saving", "validation"]
```
ruimashita: Why do you fix `available_subsets` as `["train", "train_validation_saving", "validation"]` ? This doesn't include `test` subset.

lmnet/lmnet/datasets/tensorflow_datasets.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        Initializing feature variables by using tf.Tensor from datasets.
+        For classification, "images" and "labels" should be initialized.
+        For object detection, "images" and "gt_boxes" should be initialized.
+        """
+        raise NotImplementedError()
+
+    @property
+    def num_per_epoch(self):
+        return len(self.images)
```
ruimashita: Maybe `builder.info` has the information of data number.

lmnet/lmnet/datasets/tensorflow_datasets.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class TensorFlowDatasetsClassification(TensorFlowDatasetsBase):
```
ruimashita: Please write what the class require tfds's `features` structure (maybe `label` and `image` ), addition this, It is better to check the reuirements in code.

lmnet/lmnet/datasets/tensorflow_datasets.py
tfujiwar: Yes, they have all the data on memory. I agree that it's not good for big datasets but there is no way to access TensorFlow Datasets by index because TensorFlow Datasets work as iterators. So current implementation is a workaround. To solve this problem, we need to implement this class as a new dataset iterator.

lmnet/lmnet/datasets/tensorflow_datasets.py
tfujiwar: Thanks. I'll fix it.

lmnet/lmnet/datasets/tensorflow_datasets.py
tfujiwar: Thank you. I'll add a comment and codes to check it.

lmnet/lmnet/datasets/tensorflow_datasets.py
tkng: Another idea is tha we can ignore index in the ```__getitem__``` method.

e.g. load 10000 data into memory (from TF datasets iterator), shuffle, then always pop the 1st item in the list in ```__getitem__``` method.

That's not a straightforward implementation, but as a tentative solution, we can tolerate it.

lmnet/lmnet/datasets/tensorflow_datasets.py
tfujiwar: The number of data in pre-defined split (e.g. TRAIN, VALIDATION, TEST) is included in `builder.info`. But TRAIN split will be splitted into two sub-splits if datasets only have TRAIN split.
https://github.com/tfujiwar/blueoil/blob/828d152d7f5d04cda8be9000ff9b46e7847b3d4b/lmnet/lmnet/datasets/tensorflow_datasets.py#L77-L80

That's why `len(self.images)` is used here.

lmnet/lmnet/datasets/tensorflow_datasets.py
tfujiwar: This variable is defined as `@staticmethod` in the base class. But available subsets will be decided after the class is instantiated in case of this class because it depends on what dataset will be loaded.

So I'll define this variable as `["train", "validation"]` in this class because these two subsets are always available.

lmnet/lmnet/datasets/tensorflow_datasets.py
tfujiwar: @tkng
Thanks for your suggestion. I think it's a good idea.
Currently, the method `__getitem__` seems to be used only for shuffling in dataset_iterator.

lmnet/lmnet/datasets/tensorflow_datasets.py
ruimashita: 👍

lmnet/lmnet/datasets/tensorflow_datasets.py
ruimashita: > So I'll define this variable as ["train", "validation"] in this class because these two subsets are always available.

Nice 👍

lmnet/lmnet/datasets/tensorflow_datasets.py
tfujiwar: I tried to update `__getitem__` as we talked above. I used `tf.Session().run()` for getting next element but I found that it doesn't work with the option `enable_prefetch = True`. Maybe we can fix this problem by changing `DatasetIterator` in order to use tf.Session().run() in multi processes safely.

https://github.com/blue-oil/blueoil/blob/cf5c085933aba3b7d7f319468518fd557b1322ff/lmnet/lmnet/datasets/tensorflow_datasets.py#L62-L67

But I think we should not change many things only for introducing TensorFlowDatasets class because this class is a temporary solution anyway. I and @ruimashita talked and decided not to solve the problem this time. All the data will be loaded on initialization of the dataset class.

PR #280, opened 2019-05-21
Enable to read Blueoil config from Google Cloud Storage

tfujiwar:
## Motivation and Context
Aiming to run Blueoil training on Google Kuberenetes Engine.

## Description
Updated `generate_lmnet_config.py` to use `tf.gfile` instead of standard file open function.

## How has this been tested?
- Execute `blueoil_train.py` with config file in local.
- Execute `blueoil_train.py` with config file in GCS bucket.

## Types of changes

## Checklist:

### DIFF
--- a/blueoil/generate_lmnet_config.py
+++ b/blueoil/generate_lmnet_config.py
@@ -102,7 +102,7 @@ def _load_yaml(blueoil_config_filename):
         FileNotFoundError("File not found: {}".format(blueoil_config_filename))
 
     with gfile.GFile(blueoil_config_filename, "r") as f:
-        blueoil_config = yaml.load(f, Loader=yaml.Loader)
+        blueoil_config = yaml.load(f, Loader=yaml.SafeLoader)
 
     model_name, _ = os.path.splitext(os.path.basename(blueoil_config_filename))
 
--- a/blueoil/generate_lmnet_config.py
+++ b/blueoil/generate_lmnet_config.py
@@ -21,6 +21,7 @@ from tempfile import NamedTemporaryFile
 
 import yaml
 from jinja2 import Environment, FileSystemLoader
+from tensorflow import gfile
 
 from lmnet.utils.module_loader import load_class
 from blueoil.vars import TEMPLATE_DIR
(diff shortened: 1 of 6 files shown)


blueoil/generate_lmnet_config.py
```diff
@@ -97,11 +98,11 @@ def _load_yaml(blueoil_config_filename):
...
         blueoil_config(dict): dict of blueoil config.
     """
-    if not os.path.exists(blueoil_config_filename):
+    if not gfile.Exists(blueoil_config_filename):
         FileNotFoundError("File not found: {}".format(blueoil_config_filename))
 
-    with open(blueoil_config_filename, "r") as f:
-        blueoil_config = yaml.load(f)
+    with gfile.GFile(blueoil_config_filename, "r") as f:
+        blueoil_config = yaml.load(f, Loader=yaml.Loader)
```
iizukak: why `Loader=yaml.Loader` is needed?

tfujiwar: @iizukak Thanks for your comment.
Because I wanted to get rid of this warning.
```
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
```

And I followed the same way how to read yaml in lmnet.
https://github.com/blue-oil/blueoil/blob/60d18e041d5a6cc42293957d5a1155a2aa742b18/lmnet/lmnet/utils/config.py#L247-L249

iizukak: Thanks

ruimashita: `yaml.SafeLoader` is recommended by PyYAML, why do you use `yaml.Loader` ?

ruimashita:
@tfujiwar 
Thanks 👍 I wrote a comment, Plz check it.

blueoil/generate_lmnet_config.py
tfujiwar: I used `yaml.Loader` because this loader is the default one in PyYAML==4.2 and  I didn't want to change the behavior of `yaml.load()` from the current code.

But in this case, given yaml files should have a simple yaml format. I agree with using `yaml.SafeLoader` 👍

tfujiwar:
@ruimashita Thanks for the comment. I've replied and updated the code. Please check again 🙇

tfujiwar:
run test

tfujiwar:
run blueoil test

PR #281, opened 2019-05-22
Re-support 32bit Arm

primenumber:
Fix to re-support 32bit Arm.

## Motivation and Context

Current `tca_support` doesn't support 32bit Arm.
This PR fix it.

## Description

- Implement new convert_tensor overload
- Implement new packing function
- Several fix

## How has this been tested?

```shell
$ PYTHONPATH=python/dlk python python/dlk/scripts/generate_project.py -i examples/object_detection/widerface_v5/minimal_graph_with_shape.pb -o tmp/ -p object_detection -hq -ts
$ cd tmp/object_detection
$ make -j8 lm_arm
```

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/templates/include/tensor_convert.h
+++ b/dlk/python/dlk/templates/include/tensor_convert.h
@@ -112,36 +112,6 @@ inline void convert_tensor(const TensorView<QUANTIZED_PACKED, MemoryLayout::HWCh
           after(k, i, j, d, 0) = before(i, j, k, d, 0);
 }
 
-inline void convert_tensor(const TensorView<QUANTIZED_PACKED_KERNEL, MemoryLayout::OhIhHWOlIl>& before,
-    const TensorView<QUANTIZED_PACKED_KERNEL, MemoryLayout::NHWC>& after) {
-  const auto in_shape = before.get_shape();
-  const auto oc = in_shape[0] * in_shape[4];
-  const auto icb = in_shape[1];
-  const auto height = in_shape[2];
-  const auto width = in_shape[3];
-  for (std::size_t nh = 0; nh < in_shape[0]; ++nh)
-    for (std::size_t nl = 0; nl < in_shape[4]; ++nl)
-      for (std::size_t h = 0; h < height; ++h)
-        for (std::size_t w = 0; w < width; ++w)
-          for (std::size_t c = 0; c < icb; ++c)
-            after(nh * in_shape[4] + nl, h, w, c) = before(nh, c, h, w, nl, 0);
-}
-
-inline void convert_tensor(const TensorView<QUANTIZED_PACKED_KERNEL, MemoryLayout::OhIhHWOlIl>& before,
-    const dlk::impl::kn2row_kernel_t& after) {
-  const auto in_shape = before.get_shape();
-  const auto oc = in_shape[0] * in_shape[4];
-  const auto icb = in_shape[1];
-  const auto height = in_shape[2];
-  const auto width = in_shape[3];
-  for (std::size_t h = 0; h < height; ++h)
-    for (std::size_t w = 0; w < width; ++w)
-      for (std::size_t nh = 0; nh < in_shape[0]; ++nh)
-        for (std::size_t nl = 0; nl < in_shape[4]; ++nl)
-          for (std::size_t c = 0; c < icb; ++c)
-            after(h, w, nh * in_shape[4] + nl, c) = before(nh, c, h, w, nl, 0);
-}
-
 inline void convert_tensor(const TensorView<QUANTIZED_NOT_PACKED, MemoryLayout::NHWC>& before,
     const dlk::impl::tiling_input_t& after) {
   dlk::impl::pack_input_for_tiling(before, after);
--- a/dlk/python/dlk/templates/include/tensor_convert.h
+++ b/dlk/python/dlk/templates/include/tensor_convert.h
@@ -23,17 +23,32 @@ limitations under the License.
 #include "func/impl/quantized_conv2d_tiling.h"
 #include "func/impl/quantized_conv2d_dim2col.h"
 
-inline void convert_tensor(const TensorView<QUANTIZED_PACKED, MemoryLayout::HWChBCl>& before,
+inline void convert_tensor(const TensorView<BIN_CONV_OUTPUT, MemoryLayout::HWC>& before,
+    const TensorView<BIN_CONV_OUTPUT, MemoryLayout::ChHWCl>& after) {
+  const auto in_shape = before.get_shape();
+  const auto in_height = in_shape[0];
+  const auto in_width = in_shape[1];
+  const auto out_shape = after.get_shape();
+  const auto channel_high = out_shape[0];
+  const auto channel_low = out_shape[3];
+  for (std::size_t dh = 0; dh < channel_high; ++dh)
+    for (std::size_t r = 0; r < in_height; ++r)
+      for (std::size_t c = 0; c < in_width; ++c)
+        for (std::size_t dl = 0; dl < channel_low; ++dl)
+          after(dh, r, c, dl) = before(r, c, dh * channel_low + dl);
+}
+
+inline void convert_tensor(const TensorView<QUANTIZED_PACKED, MemoryLayout::ChHWBCl>& before,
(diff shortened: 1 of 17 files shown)


primenumber:
Output of first binary convolution doesn't match.
I have no idea...

iizukak:
@n-nez @antonionevado Do you have any idea about the problem with output doesn't match?

dlk/python/dlk/templates/manual/consts/input.tpl.cpp
```diff
@@ -28,19 +28,11 @@ const TensorView<{{ node.dtype.cpptype() }}, MemoryLayout::Atom>
...
   {{- d -}},
   {%- endfor %}
 };
-#else
-static Base<{{ node.dtype.cpptype() }}>::type {{ node.name }}_raw[] = {
-  {% for d in node.data.flatten() -%}
-  {{- d -}},
-  {%- endfor %}
-};
-#endif
```
n-nez: FPGA uses a layout of kernel data which is different from CPU version...
Do you really want to remove this one? :thinking:

dlk/python/dlk/templates/src/func/impl/pack_16bit.cpp
```diff
@@ -0,0 +1,47 @@
...
+==============================================================================*/
+
+#include "func/impl/pack_16bit.h"
+#include <cassert>
+
+namespace dlk {
+
+namespace impl {
+
+void pack_16bit(const BIN_CONV_OUTPUT input[], QUANTIZED_PACKED output[], const std::size_t length) {
```
n-nez: Other thing is that output could contain number of channels which is not multiple of 32, while your code seems to assuming that everything is multiple of 32, as otherwise you need to know not only lenght, but also how many channels you have.

For example:
If number of channel is 4, and first 8 values are 3,3,3,3,1,1,1,1, and all other values are 0 output proper output should be
```
LSB[0] = 0x0000000F // 4 channels packed
MSB[0] = 0x0000000F
LSB[1] = 0x0000000F // 4 channels packed
MSB[1] = 0x00000000
```
While your code seems to be producing something like
```
LSB[0] = 0x000000FF // 32 channels packed
MSB[0] = 0x0000000F
```
for this case... :thinking:

dlk/python/dlk/templates/manual/consts/input.tpl.cpp
primenumber: This PR code convert kernel layout in runtime from layout for FPGA to layout for CPU.

n-nez: Could not say that converting layout of statically generated data sounds like a good idea to me... :thinking:

dlk/python/dlk/templates/src/func/impl/pack_16bit.cpp
primenumber: In my code, number of channels is ceiling up to multiple of 32.

dlk/python/dlk/templates/manual/consts/input.tpl.cpp
primenumber: Kernel layout for kn2row is not same as kernel layout of tiling/im2col, if we want to generate data statically, python code should know which method is used...

n-nez: Hmm...
Doesn't seem particularly challenging to do it in python, but as workaround sounds good for me...

On the other hand, not sure do we really need `im2col`, btw. So far we did not use quantized convolutions which kn2row (read fpga) don't work for, and if we use them performance will be not very good. Thus maybe it is a good time to stop pretending to be generic. :thinking:

primenumber: OK. I'll remove `im2col` and `kn2row` and generate kernel data for CPU statically.

primenumber:
Thanks to @n-nez , output is matched!!! :100: :tada: :confetti_ball:

dlk/python/dlk/templates/manual/consts/input.tpl.cpp
primenumber: It could take several days to implement tiling for x86(done in another PR).

n-nez: Could we use `./dlk/python/dlk/templates/src/func/impl/generic/quantized_conv2d_kn2row.cpp` for x86 for now? :thinking:

primenumber: It's still good (more layout convert in python needed).

primenumber:
Fixed to generate kernel data for CPU statically.

dlk/python/dlk/templates/include/tensor_convert.h
```diff
@@ -37,6 +112,36 @@ inline void convert_tensor(const TensorView<QUANTIZED_PACKED, MemoryLayout::HWCh
...
+  const auto icb = in_shape[1];
+  const auto height = in_shape[2];
+  const auto width = in_shape[3];
+  for (std::size_t h = 0; h < height; ++h)
+    for (std::size_t w = 0; w < width; ++w)
+      for (std::size_t nh = 0; nh < in_shape[0]; ++nh)
+        for (std::size_t nl = 0; nl < in_shape[4]; ++nl)
+          for (std::size_t c = 0; c < icb; ++c)
+            after(h, w, nh * in_shape[4] + nl, c) = before(nh, c, h, w, nl, 0);
+}
```
n-nez: Do you still need this :arrow_up: functions?

dlk/python/dlk/templates/src/func/impl/apply_thresholds.cpp
```diff
@@ -54,6 +54,8 @@ void ApplyThresholds(
           new_d = 2;
         else
           new_d = 3;
+      } else if (flag == 0) { // ignore
+        new_d = 0;
```
n-nez: Why do you need this one? :thinking:

dlk/python/dlk/templates/include/tensor_convert.h
primenumber: Oh, these functions are not needed no longer.

dlk/python/dlk/templates/src/func/impl/apply_thresholds.cpp
primenumber: threshold data is only provided "original" number of channels, and rest of data are filled with zero.
In this process, rest of data should be ignored (new_d can be arbitrary number between 0 and 3).

primenumber: Uh, new_d must be 0 (if next op is conv).

n-nez: Hmm, I see.

BTW, new_d could be anything nut keeping it zero, may simplify debug... :thinking: 

If the next operator is conv, the number of output channels is already multiple of 32, i.e. there is no padding on kernels/thresholds. as the number of input, channels has to be 32. Otherwise, here is space2depth or something else in between, which must to ignore all "non" existent channels inside of packed words.

primenumber: > If the next operator is conv, the number of output channels is already multiple of 32
Oh, really?

n-nez: Yep, as so far a number of input channels must be multiple of 32.
Ok, there are some exceptions, like Lookup which produce less than 32 and pads rest with zeroes.
But in all other places, we were always relying on input channels to be a power of 32.

dlk/python/dlk/templates/include/tensor_convert.h
n-nez: Then let's remove them.

dlk/python/dlk/templates/src/func/impl/apply_thresholds.cpp
primenumber: OK, what new_d should be?

dlk/python/dlk/templates/include/tensor_convert.h
primenumber: Removed.

PR #316, opened 2019-06-13
[WIP] Save checkpoint when Sigterm Signal Received

### LINKED ISSUE #315: Save checkpoint on Sigterm wanted
I want to save the last checkpoint when `ctrl+C` was pressed during training.

Joeper214:
## Motivation and Context

- FIx #315 

The reason for implementing this feature was to catch `SIGTERM` signals. 
So it doesn't only cover `ctrl-C` 

Since we want to use Kubernetes to run training, I found out that saving checkpoints might be useful during graceful shutdowns of Pods on Kubernetes. Since it is sending [`SIGTERM`](https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html)  signals to the main process to not to lose certain checkpoints if the training was terminated.

But I'm not sure if this really works when a Kubernetes pod is being gracefully shutdown so it might need more tests in that case.

### Run the following command both on lmnet and blueoil commands

Lmnet example
$ PYTHONPATH=. python executor/train.py -c configs/example/classification.py

Blueoil example
$ ./blueoil.sh train config/cifar10.yml

> press `ctrl+c` when the training starts that shows the current step in the terminal.
> After keyboard interrupt command, it should save the current checkpoint of your training before it terminates.

I also added a unit test file to check if signal handler works.
[test_signal_handler.py](https://github.com/blue-oil/blueoil/pull/316/files#diff-022bf5172d327b83257f59429fc23ce3)

WIP: Test on k8s and post the results and how to reproduce.

## Types of changes

## Checklist:


Joeper214:
run test

iizukak:
@Joeper214 
Can you describe the details how to test this branch?
Reviewer need how to reproduct your result.

Joeper214:
> @Joeper214
> Can you describe the details how to test this branch?
> Reviewer need how to reproduct your result.

Sure I'll do that.

Joeper214:
run lmnet test

ruimashita:
@Joeper214 Please tell me why do you want to save checkpoint on Ctrl-C ?
I'm not sure when and what this feature give us ? I never have trouble of saving checkpoints.

Joeper214:
@ruimashita 

Sorry for the lack of information about this PR. I'll add this reply comment to the description.

The reason for implementing this feature was to catch `SIGTERM` signals. So it doesn't only cover `ctrl-C` 

Since we want to use Kubernetes to run training. I found out that saving checkpoints during graceful shutdowns of Pods on Kubernetes is sending `SIGTERM` signals to main process [ref](https://pracucci.com/graceful-shutdown-of-kubernetes-pods.html)

But I'm not sure if this really works when a Kubernetes pod is being gracefully shutdown so it might need more tests in that case.

ruimashita:
@Joeper214 Thanks 😄  I understand this PR motivation. According to this motivation, I think this pr should be merge after pass test with k8s.

Joeper214:
> @Joeper214 Thanks 😄 I understand this PR motivation. According to this motivation, I think this pr should be merge after pass test with k8s.

Thanks. I'll test this with k8s as mention. 
I'll keep you posted once I'm done testing. And post the results here.

lmnet/executor/train.py
```diff
@@ -351,11 +353,19 @@ def start_training(config):
...
+        if signalhandler.receivedTermSignal:
+            _save_checkpoint(saver, sess, global_step, step)
+            sys.exit(0)
+
     # training loop end.
     print("reach max step")
 
 
 def run(network, dataset, config_file, experiment_id, recreate):
+    global signalhandler
```
ruimashita: Could you avoid to use global keyword ?

Joeper214:
I can't keep track of my commits. I'll close this PR and create another one from the new master branch.

PR #313, opened 2019-06-12
Fix for test

primenumber:
## Description

- Fix type
- Fix to avoid mmap error

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature / Optimization (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #317, opened 2019-06-13
Updated Reshape operator with a default value.

ed728:
## Motivation and Context

Fixed Jenkins test failing here: https://github.com/blue-oil/blueoil/pull/313

## Description

dimension_format for Reshape had no default value and there were several
places which used it without providing one. I opted for setting
a default value rather than updating all places to supply
one of their own. Not sure why something like Reshape needs
this parameter though; seems like the superclass requires it so
we have it in Reshape as a consequence. If that is the only reason,
removing it from the Reshape constructor and using the
superclass's default would be better.

## How has this been tested?

Ran `make test-dlk` on my own DE10-Nano and all test passed.

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/core/operators.py
+++ b/dlk/python/dlk/core/operators.py
@@ -1883,7 +1883,12 @@ class Reshape(Operator):
     _input_names = ['data', 'shape']
     _output_names = ['reshaped']
 
-    def __init__(self, name: str, shape: List[int], dtype: DataType, input_ops: Ops, dimension_format: str) -> None:
+    def __init__(self,
+                 name: str,
+                 shape: List[int],
+                 dtype: DataType,
+                 input_ops: Ops,
+                 dimension_format: str = 'NHWC') -> None:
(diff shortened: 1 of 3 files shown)


yd8534976:
Yes, I think `Reshape` does not need this dimension_format. And if we pass a true `dimension_format`, `pass_transpose` will not transpose the `dimension_format` in `Reshape`. It makes test failed. Should we modify `pass_transpose`?

ed728:
Another option is to remove it from Reshape. Its superclass of Operator already has a default for `dimension_format`, so not passing anything in would be fine. Then we can just remove all cases where we pass Reshape a `dimension_format` like in tf.py.

primenumber:
Wrong `dimension_format` cause wrong code generation.
We should pass a true `dimension_format`.
I think that we should modify `pass_transpose`.

iizukak:
@ed728 @primenumber Great. thank for fixing 👍

PR #318, opened 2019-06-13
Fixed unsigned to signed integer conversion issue.

ed728:
## Motivation and Context

When running tests a lot of warning regarding `unsigned int` to `int` and vice-versa conversion appear. This makes the error log much harder to read, and in general it would be best to not have errors, by either muting(marking as not an error) them or fixing them.

## Description

Changed 2 places where we assigned `T_UINT` to `int` for some reason. Explicit usage of `unsigned int` might be desired here, but I am not sure. To be honest, I have no clue why we have `T_UINT` and the like.

Also fixed `n_bits` being `T_INT`. I assume this is to show how many bits we are using, which I believe cannot be negative.

## How has this been tested?

Local run of `make test-dlk`.

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/templates/include/func/quantized_conv2d.h
+++ b/dlk/python/dlk/templates/include/func/quantized_conv2d.h
@@ -34,13 +34,13 @@ void QuantizedConv2D(const TensorView<T, layout>& input,
   Measurement::Start("QuantizedConv2D");
 
   constexpr T_UINT TilingInTypeBitWidth = dlk::impl::tiling_input_elem_t::BitCount;
-  int kh = p.normal_conv_params.kernel_height;
-  int kw = p.normal_conv_params.kernel_width;
-  int padding = p.normal_conv_params.padding;
-  int ih = p.normal_conv_params.input_height;
-  int iw = p.normal_conv_params.input_width;
-  int ic = p.normal_conv_params.kernel_depth;
-  int oc = p.normal_conv_params.output_channels;
+  T_UINT kh = p.normal_conv_params.kernel_height;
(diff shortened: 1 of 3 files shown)


n-nez:
Most places you are changing just creating a copy of a value from some structure.

Why not just use `auto` to be sure that types on both sides of `=` are equal?

ed728:
@n-nez I think they should be kept as is though, to avoid any future problems.

Considering we had the warning, seems like no one looked into the `struct` originally, and we are doing computations on the variables(assigning that to `auto`) and passing that into other functions.

I dunno, I think if someone ever changes the `struct`, they should have to take a look at a place like this. I mean, this is close to the computational back-end. 

If you insist on this being `auto` I can change it though.

iizukak:
@n-nez @ed728 
Umm, It's ok to keep explicit  type instead of `auto`.
I'm not sure which is better.

n-nez:
@iizukak, @ed728, In general, using `auto` is good practice in modern C++
- to avoid narrowing type conversion problems
- to make sure all dependent code uses propper types when you update your structures / whatever

in this particular case of how do we use these variables, there is no any real harm from using nor from not using it, so I would prefer to merge this one as is, instead of spending energy on discussing it here. :innocent:

ed728:
@n-nez Thanks!
@iizukak  Should this be merged to `master` instead?

ed728:
@ruimashita ? Any comment at least?

n-nez:
@ruimashita, @ed728

Closed by mistake? :thinking_face:

ruimashita:
@n-nez @ed728 
This PR automatically closed when delete `tca_support` branch.

Issue #314, opened 2019-06-12
Fix generate_consts() in code_generater.py


yd8534976:
```
    def generate_consts(self) -> None:
        const_src_dir_path = path.join(self.src_dir, 'consts')
        const_header_dir_path = path.join(self.header_dir, 'consts')
        utils.make_dirs([const_src_dir_path, const_header_dir_path])

        const_src_template_path = path.join('manual', 'input', 'const.tpl.cpp')
        const_header_template_path = path.join('manual', 'input', 'const.tpl.h')

        for const in self.graph.consts:

            self.template.generate(const_src_template_path,
                                   const_src_dir_path,
                                   new_name=const.name,
                                   const=const)

            self.template.generate(const_header_template_path,
                                   const_header_dir_path,
                                   new_name=const.name,
                                   const=const)
```

I can not find these two template files, I'm not sure if it is bug?
`const_src_template_path = path.join('manual', 'input', 'const.tpl.cpp')
const_header_template_path = path.join('manual', 'input', 'const.tpl.h')`

PR #320, opened 2019-06-14
Removed unused generate_consts.

ed728:
## Motivation and Context

https://github.com/blue-oil/blueoil/issues/314

## Description

I am pretty sure `generate_consts()` is an old thing we used to handle consts, and have switched to a different method with a different folder with several files of consts.

`grep "generate_consts" -r .` produces only `./dlk/python/dlk/code_generater.py:    def generate_consts(self) -> None:`, which makes we think it is not used. There may still be the possibility that it is called via some Python magic somewhere, but Jenkins tests should be sufficient to cover that.

## How has this been tested?

Jenkins.
It's an unused function, so don't see much testing that can be done.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue) (Maybe?)

## Checklist:

### DIFF
--- a/dlk/python/dlk/code_generater.py
+++ b/dlk/python/dlk/code_generater.py
@@ -70,26 +70,6 @@ class CodeGenerater(object):
                 else:
                     shutil.copy2(src_file_path, dest_file_path)
 
-    def generate_consts(self) -> None:
-        const_src_dir_path = path.join(self.src_dir, 'consts')
-        const_header_dir_path = path.join(self.header_dir, 'consts')
-        utils.make_dirs([const_src_dir_path, const_header_dir_path])
-
-        const_src_template_path = path.join('manual', 'input', 'const.tpl.cpp')
-        const_header_template_path = path.join('manual', 'input', 'const.tpl.h')
-
-        for const in self.graph.consts:
-
-            self.template.generate(const_src_template_path,
-                                   const_src_dir_path,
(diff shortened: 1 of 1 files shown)


iizukak:
run test

yd8534976:
Thanks so much for fixing! But I think in this step, we may skip the `generate_const` step. Could you please test if dlk can compile a graph with `Const` node?
```
def generate_code_step(model: Model, config: Config) -> None:
    """Generate code for the model.

    Parameters
    ----------
    model : Model
        Model the code generation is based on

    config : Config
        Collection of configurations

    """
    graph: Graph = model.graph
    params = Params(graph, config)

    builder = CodeGenerater(graph,
                            params,
                            config)

    builder.reuse_output_buffers()
    builder.generate_files_from_template()
    builder.generate_inputs()

    if config.activate_hard_quantization:
        builder.generate_scaling_factors()

    if config.threshold_skipping:
        builder.generate_thresholds()

    if config.use_tvm:
        builder.generate_tvm_libraries()
```

yd8534976:
@ed728 , maybe it is just elder version of `generate_inputs()`...

ed728:
The old consts templates don't exist anymore, and now we have a consts folder, which is what functions like `generate_inputs` use.

iizukak:
@nlpng Can you reveiw?

iizukak:
@nlpng Please review

Issue #346, opened 2019-06-28
Pass_lookup failed to convert some cases


yd8534976:
For example, it will raise `Illegal input name` when converting below graph. (`lookup` followed by `SpaceToDepth`)
<img width="448" alt="Screen Shot 2019-06-28 at 18 13 41" src="https://user-images.githubusercontent.com/8168498/60332002-cf333e00-99d0-11e9-9d40-1fef7930897e.png">

When I change graph into this case. It will be fine.
<img width="572" alt="Screen Shot 2019-06-28 at 18 13 48" src="https://user-images.githubusercontent.com/8168498/60331972-c04c8b80-99d0-11e9-8a4c-945ac7b05488.png">

PR #347, opened 2019-07-01
Fix pass_lookup()

yd8534976:
Current implementation of `pass_lookup() ` hard-coded the next operation following `lookup`. Since input names are different between each ops. (e.g. Conv: ['X', 'W', 'B'], SpaceToDepth: ['input']) It will fail to convert some cases. #346 
I fix it to deal with more cases. (Thank @nlpng for helping.)

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/core/optimizer.py
+++ b/dlk/python/dlk/core/optimizer.py
@@ -646,7 +646,14 @@ def pass_lookup(graph: Graph) -> None:
         pe.add_outputs(quantizer.output_ops)
 
         output_op = quantizer.output_op_list[0]
-        output_op.add_input(output_op._input_names[0], pe)
+
+        target_input_name = 'X'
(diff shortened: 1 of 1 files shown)


iizukak:
@antonionevado Can you check?

yd8534976:
@antonionevado I updated the implementation. Please check it. 🙇

antonionevado:
@yd8534976 Do we have any [dlk] test that uses a protobuf file with a Lookup layer?

yd8534976:
@antonionevado Thanks! I test it with [a pb file](https://drive.google.com/open?id=1MNP_-J98ZQfqCwmoAx5USLpeGiv0SOpi). It works fine with this file.

yd8534976:
@antonionevado The failed case in #346 .

PR #348, opened 2019-07-01
Save last checkpoint on SIGTERM

Joeper214:
## Motivation and Context

[ref](https://medium.com/pixboost/save-cash-by-running-kubernetes-services-on-preemptible-vms-in-google-cloud-cca02809ae09)

## Description

The reason for implementing this feature was to catch SIGTERM signals.

Since we want to use Kubernetes to run training, I found out that saving checkpoints might be useful during graceful shutdowns of Pods on Kubernetes. Since it is sending SIGTERM signals to the main process to not to lose certain checkpoints if the training was terminated.

- create unit test for signal handler
- save checkpoint on sigterm on train.py
- Added signal Handler util class

## How has this been tested?

Can be tested on both `docker` and `kubectl` commands.

### Test on Kubernetes
1. create job
$ kubectl create -f config/mnist-job.yaml
2. Wait until the pod was created and training starts.
$ kubectl -f logs  <pod_id>
3. execute delete job command.
$ kubectl delete job <job_name>

It should show logs same as below.
```
step 603
step 604
step 605
step 606
step 607
step 608
step 609
step 610
step 611
step 612
step 613
step 614
step 615
step 616
step 617
step 618
step 619
step 620
step 621
step 622
Save ckpt. step: 623.
```

### Test on docker
1. Run training
$ ./blueoil.sh train config/{Model name}.yml
2. Wait until the training starts.
3. execute docker stop command.
$ docker stop

It should save the last checkpoint after stop command.

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/lmnet/utils/signal_handler.py
+++ b/lmnet/lmnet/utils/signal_handler.py
@@ -1,0 +1,34 @@
+
+# -*- coding: utf-8 -*-
+# Copyright 2019 The Blueoil Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
+import signal
+
+
+class SignalHandler:
+    """ Protect a piece of code from being killed by SIGINT or SIGTERM.
+    It can still be killed by a force kill.
+
+    Both functions will be executed even if a sigterm or sigkill has been received.
+    """
+
+    def __init__(self):
+        self.receivedTermSignal = False
+        signal.signal(signal.SIGTERM, self.handler)
+        signal.signal(signal.SIGINT, self.handler)
+
+    def handler(self, signum, frame):
+        self.lastSignal = signum
+        self.receivedTermSignal = True
--- a/lmnet/executor/train.py
+++ b/lmnet/executor/train.py
@@ -15,12 +15,14 @@
 # =============================================================================
 import os
 import math
+import sys
 
 import click
 import tensorflow as tf
 from tensorflow.core.util.event_pb2 import SessionLog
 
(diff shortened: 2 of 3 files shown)


Joeper214:
@iizukak @ruimashita Please review if you have time. 🙇

iizukak:
@Joeper214 Thanks. I'll test this PR 🎉

lmnet/lmnet/utils/signal_handler.py
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
+import signal
+
+
+class SignalHandler:
```
iizukak: Making `SignalHandler` class is necessary?
I think it's ok to write signal catch in `train.py`.

Joeper214: @iizukak Thanks for the comment.

I created this class to make it reusable somewhere else since it's just a utility function.
Also, I was thinking of decluttering `train.py` as it has so many things going on IMO.
What do you think?

lmnet/lmnet/utils/signal_handler.py
```diff
@@ -0,0 +1,34 @@
...
+    """ Protect a piece of code from being killed by SIGINT or SIGTERM.
+    It can still be killed by a force kill.
+
+    Both functions will be executed even if a sigterm or sigkill has been received.
+    """
+
+    def __init__(self):
+        self.receivedTermSignal = False
+        signal.signal(signal.SIGTERM, self.handler)
+        signal.signal(signal.SIGINT, self.handler)
```
iizukak: This code looks if `SIGINT` catched, then training will be terminated.
It's not same as current behavior.
Currently, we continue training, It's my misunderstand?

Joeper214: Yes, This code initializes the Class to define the signal to be caught. 

> Currently, we continue training, It's my misunderstand?
yes, we can continue training by saving the last checkpoint.

iizukak:
@Joeper214 
This PR is still want to be merged?
I think you and ruimashita decided to implement use can set the save interval and max number of checkpoints.

Joeper214:
> @Joeper214
> This PR is still want to be merged?
> I think you and ruimashita decided to implement use can set the save interval and max number of checkpoints.

Yes. That's right. I'm closing this PR in that case.

PR #354, opened 2019-07-04
Improve TFDS Dataset class and iterator

tfujiwar:
## Motivation and Context
Aiming to load larger datasets by using TFDS.

## Description
- Update TFDS dataset class not to load all data on the initialization.
- Add a reader for TFDS dataset class to dataset_iterator.

## How has this been tested?
Run `train.py` with the dataset configuration below:
With downloading:
```python
DATASET_CLASS = TFDSClassification
DATASET.TFDS_NAME = 'cifar10'
DATASET.TFDS_DATA_DIR = 'gs://lmfs-backend-us-west1/shared/tensorflow_datasets'
DATASET.TFDS_DOWNLOAD = True
```

Without downloading:
```python
DATASET_CLASS = TFDSClassification
DATASET.TFDS_NAME = 'cifar10'
DATASET.TFDS_DATA_DIR = 'gs://lmfs-backend-us-west1/shared/tensorflow_datasets'
DATASET.TFDS_DOWNLOAD = False
```

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/lmnet/datasets/tfds.py
+++ b/lmnet/lmnet/datasets/tfds.py
@@ -29,6 +29,8 @@ def _label_to_one_hot(record, depth):
 class TFDSBase(Base):
     """
     Abstract dataset class for loading TensorFlow Datasets.
+    Only images which has 3 channels (RGB) can be loaded for now.
+    TODO(fujiwara): Convert grayscale images into RGB images.
     """
     available_subsets = ["train", "validation"]
     extend_dir = None
--- a/lmnet/lmnet/datasets/tfds.py
+++ b/lmnet/lmnet/datasets/tfds.py
@@ -13,12 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # =============================================================================
-import numpy as np
 import tensorflow as tf
 import tensorflow_datasets as tfds
 
 from lmnet.datasets.base import Base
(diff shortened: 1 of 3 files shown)


lmnet/lmnet/datasets/tfds.py
```diff
@@ -39,32 +46,40 @@ def __init__(
...
+        )
 
     @property
     def num_per_epoch(self):
         split = self.available_splits[self.subset]
         return self.info.splits[split].num_examples
 
+    @property
+    def classes(self):
+        return self.info.features["label"].names
```
ruimashita: Need to change when implement object detection dataset class, because `voc2007` builder use `labels` (not `label`) as label.
 https://github.com/tensorflow/datasets/blob/7dd97e23bd7cb5c07b5da19730b5cf377873af12/tensorflow_datasets/image/voc.py#L138
https://github.com/tensorflow/datasets/blob/master/docs/datasets.md#features-32

Will we change this line when we implement Object detection dataset class ?

lmnet/lmnet/datasets/tfds.py
```diff
@@ -39,32 +46,40 @@ def __init__(
...
+
+    @property
+    def num_classes(self):
+        return self.info.features["label"].num_classes
```
ruimashita: also here

Issue #362, opened 2019-07-11
Kernel Initializer is not passed into conv2D


ananno:
# Background
I've been working with the new blueoil/lmnet code and found out that the `kernel_initializer` is not passed in `conv2D` wrapper function defined in `lmnet/lmnet/layers/layers.py`. The `kernel_initializer` argument in `darknet` block function in `lmnet/lmnet/blocks.py` provided but it is not passed into the actual `tf.layers.conv2d` function inside the wrapper function `conv2D`.

* `lmnet/lmnet/blocks.py`
```
# TODO(wakisaka): should be replace to conv_bn_act().
def darknet(name, inputs, filters, kernel_size,
            is_training=tf.constant(False), activation=None, data_format="NHWC", *args, **kwargs):
    """Darknet19 block.

    Do convolution, batch_norm, bias, leaky_relu activation.
    Ref: https://arxiv.org/pdf/1612.08242.pdf
         https://github.com/pjreddie/darknet/blob/3bf2f342c03b0ad22efd799d5be9990c9d792354/cfg/darknet19.cfg
         https://github.com/pjreddie/darknet/blob/8215a8864d4ad07e058acafd75b2c6ff6600b9e8/cfg/yolo.2.0.cfg
    """
    if data_format == "NCHW":
        channel_data_format = "channels_first"
    elif data_format == "NHWC":
        channel_data_format = "channels_last"
    else:
        raise ValueError("data format must be 'NCHW' or 'NHWC'. got {}.".format(data_format))

    with tf.variable_scope(name):
        if activation is None:
            def activation(x): return tf.nn.leaky_relu(x, alpha=0.1, name="leaky_relu")

        conv = conv2d("conv", inputs, filters=filters, kernel_size=kernel_size,
                      activation=None, use_bias=False, data_format=channel_data_format,
                      kernel_initializer=tf.contrib.layers.variance_scaling_initializer(),
                      *args, **kwargs)  # he initializer

        # TODO(wakisaka): Should be the same as darknet batrch norm.
        # https://github.com/tensorflow/tensorflow/blob/r1.1/tensorflow/contrib/layers/python/layers/layers.py
        # https://github.com/pjreddie/darknet/blob/8215a8864d4ad07e058acafd75b2c6ff6600b9e8/src/batchnorm_layer.c#L135
        batch_normed = batch_norm("bn", conv, is_training=is_training, decay=0.99, scale=True, center=True,
                                  data_format=data_format)
[...]

PR #364, opened 2019-07-11
Kernel Initializer issue

ananno:
To fix the kernel initializer issue in function `conv2d` in `lmnet/lmnet/layers.py`.

## Motivation and Context

The argument definition of `kernel_initializer` of the function `conv2d` in `lmnet/lmnet/layers/layers.py` has never been passed into `tf.layers.conv2d` function call resulting it not effective even if it is passed from the calling function. This change is necessary to pass the parameters in `tf.layers.conv2d` function by defining and passing arguments like `kernel_initializer` or `kernel_regularizer` etc.

## Description

The `kernel_initializer` argument is defined in the function argument of `conv2d` but never passed in `tf.layers.conv2d` function in the code body. This results to not passing the kernel regularizer defined or passed by the calling function. 

The fix is very simple. By removing the argument definition in `conv2d` function in `lmnet/lmnet/layers/layers.py` file. The defined argument like `kernel_initializer` or any others supported by `tf.layers.conv2d` function can be automatically passed via `*args, **kwargs`.

Details discussion on this is documented in issue [here](https://github.com/blue-oil/blueoil/issues/362).

## How has this been tested?

From the code it is obvious that the argument definition of `kernel_initializer` in `conv2d` function has never been used.

image

## Types of changes

## Checklist:


Joeper214:
@ananno Please rebase this branch from the current `blueoil/master` branch. 

So that we can only see the commits from your branch.
Currently, there are `1,004` commits which many are unrelated to this PR.

ananno:
> @ananno Please rebase this branch from the current `blueoil/master` branch.
> 
> So that we can only see the commits you did on your branch.
> Currently, there are `1,004` commits which many are unrelated to this PR.

I've resolved the issue. Please check and confirm.
Thank you.

Joeper214:
run test

iizukak:
@ananno 
Thank you for fixing the bug!!
As you comment in this PR, current lmnet implementation looks not pass `kernel_initializer` to TensorFlow.
So, currently, TensorFlow use default value `None`
https://www.tensorflow.org/api_docs/python/tf/layers/conv2d

And if your code will be merged, we use `xavier_initializer` as default (in your screenshot).

I thinks make `None` as default looks better.
How do you think?

iizukak:
@nlpng 
This PR effect many network.
It's good to be merge?

iizukak:
I checked TensorFlow use `glorot_uniform` when `kernel_initializer` is `None`.
So there is no problem about default initializer.

iizukak:
run test

Joeper214:
@ananno Please fix failed tests as well.

ananno:
@tsawada San: Can you please do Readability review on this PR.
Thank you

ananno:
@iizukak San: You approved this PR but I think according to new policy you've to approve the ownership review. Please approve as ownership review.

Thank you

iizukak:
Please check CI status.

ananno:
@iizukak @tsawada San:
I though this would be a very simple fix (passing the `kernel_initializer` through `**kwargs` but as i investigate the test failure in `yolo_v2` training it expose another significant bug. 

The `kernel_initializer` parameter passing from darknet block was never passed into `conv2D` before. But when I fixed `conv2D` it turns out the parameter passing. Which means so far, all the network inherits yolov2, i.e. `lmfyolo`, didin't use `variance_scaling_initializer` but use the default normalized xavier initializer. This should cause performance difference in YoloV2 and all it's inheritor performance. 

As you already know for relu activation it was beneficial to use `variance_scaling_initializer` as kernel_initializer but the implement was wrong. Also in TF 1.15.0 the implementation was also changed.

Anyway, I've fixed both the issue and now all the tests are passed. I think it was a good finding. Please review one more time.

tsawada:
I think @tkng san knows better so I'll leave readability review to him.

iizukak:
@tkng 
I think we can merge this PR.

ananno:
@tkng San:
Can you please review one more time.

Thank you

ananno:
@iizukak San:
Can you please review one more time and if it is ok please give a OA and LGTM on this PR.

lmnet/lmnet/layers/layers.py
```diff
@@ -50,7 +50,6 @@ def conv2d(
     strides=1,
     padding="SAME",
     activation=tf.nn.relu,
-    kernel_initializer=tf.contrib.layers.xavier_initializer(),
```
iizukak: @ananno Sorry, I have one question, before your change, we use `layers.xavier_initializer` and ignore `variance_scaling_initializer` by bug. 
Your fix change initializer `xavier_initializer` to `variance_scaling_initializer` and it's affect training accuracy?  🤔

ananno: I didn't trained with both. But theoretically it would better perform with `variance_scaling_initializer` as it works better with `ReLU` activation whereas `xavier_initializer` works better with sigmoid activation.
Ref. Xavier : http://proceedings.mlr.press/v9/glorot10a/glorot10a.pdf
Ref. Variance Scaling : https://arxiv.org/pdf/1502.01852v1.pdf

iizukak: I want @tkng san's opinion about this.

ananno: Moreover, before we didn't pass `kernel_initializer` into the `tf.layers.conv2d` function anyway, so it would use the default value always and our selection through config was not being applied.

iizukak: I see.

ananno:
@tsawada San:
Can you please review this PR for Readability Approval.

lmnet/lmnet/blocks.py
```diff
@@ -40,7 +40,7 @@ def activation(x): return tf.nn.leaky_relu(x, alpha=0.1, name="leaky_relu")
 
         conv = conv2d("conv", inputs, filters=filters, kernel_size=kernel_size,
                       activation=None, use_bias=False, data_format=channel_data_format,
-                      kernel_initializer=tf.contrib.layers.variance_scaling_initializer(),)  # he initializer
+                      kernel_initializer=tf.variance_scaling_initializer,)  # he initializer
```
tkng: According to the reference manual, `tf.contrib.layers.variance_scaling_initializer` is he initializer (MSRA initializer), but `tf.variance_scaling_initializer` is not. (Latter doesn't have an explicit short name, Caffe's default.)

- https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/contrib/layers/variance_scaling_initializer
- https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/initializers/variance_scaling

I think it's better to keep `tf.contrib.layers.variance_scaling_initializer()`, since he initializer is major than `tf.variance_scaling_initializer`'s default setting.

ananno: I see. In that case we can keep it for now. But it will be deprecated in TF2.0, until than we can use it.
 
Screenshot from 2020-01-29 16-35-16

ananno:
> Please revert this change, pull request should not include any cosmetic change without reason.
> 
> (I meant this line
> 
> https://github.com/blue-oil/blueoil/blob/984cd163bd26f80efb6989fea29ca94824615603/lmnet/lmnet/blocks.py#L43
> 
> )

@tkng San:
Thank you for the explanation. I've revert the cosmetic change. Please review one more time.

tkng:
Readability Approval
Ownership Approval

PR #366, opened 2019-07-11
Add scale augmentation for all existing tasks

yd8534976:
## Motivation and Context

Scale augmentation is an important augmentation especially for detection tasks. This PR implements Scale class to support scale augmentation for all existing tasks.

## Description

- Remove affine_scale()
- Implement `Scale` class.

## How has this been tested?

It is tested well using below notebook.

https://colab.research.google.com/drive/1MvRCKS92LylyhDsx0nIF9Drtt5f97OxZ#scrollTo=K-0KmaAYAkqb

## Types of changes

- [X] New feature / Optimization (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/lmnet/lmnet/data_augmentor.py
+++ b/lmnet/lmnet/data_augmentor.py
@@ -893,22 +893,69 @@ def color_filter(img):
     return new_image
 
 
-# TODO(wakisaka): implement class
-def affine_scale(img, scale, fill_color="white"):
-    """Resize image to scale size keeping the aspect ratio and place it in center of fill color image."""
-    image = Image.fromarray(np.uint8(img))
-    original_width = image.size[0]
-    original_height = image.size[1]
+class Scale(data_processor.Processor):
+    """Scale.
 
-    outer_width = original_width - (original_width * scale)
-    outer_height = original_height - (original_height * scale)
-    new_image = Image.new(
-        'RGB',
-        (original_width, original_height),
-        fill_color,
-    )
+    Args:
+        scale_range (tuple): Scale range.
+        fill_color (str): 'black' or 'white'.
+    """
 
-    scaled = image.resize((int(original_width * scale), int(original_height * scale)))
-    new_image.paste(scaled, (int(outer_width / 2), int(outer_height / 2)))
+    def __init__(self, scale_range=(0.7, 1.35), fill_color='black'):
 
-    return np.array(new_image)
+        min_scale, max_scale = scale_range
+
+        assert min_scale >= 0.5
+        assert max_scale <= 1.5
+
+        self.min_scale = min_scale
+        self.max_scale = max_scale
+        self.fill_color = fill_color
+
+    def __call__(self, image, mask=None, gt_boxes=None, *args, **kwargs):
+
+        scale = random.uniform(self.min_scale, self.max_scale)
+
+        image = Image.fromarray(np.uint8(image))
+        original_width = image.size[0]
+        original_height = image.size[1]
+
+        outer_width = original_width - (original_width * scale)
+        outer_height = original_height - (original_height * scale)
+        new_image = Image.new('RGB',
+                              (original_width, original_height),
+                              self.fill_color)
+
+        scaled = image.resize((int(original_width * scale), int(original_height * scale)))
+        new_image.paste(scaled, (int(outer_width / 2), int(outer_height / 2)))
+
+        if mask is not None:
+            mask = Image.fromarray(np.uint8(mask))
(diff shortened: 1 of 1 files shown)


yd8534976:
I have not addressed a very small boxes in edge, maybe I should drop them directly. 🤔

yd8534976:
run test

yd8534976:
> Did you any experiment with Scale augmentation?
I want result of training using and not using training result.

@iizukak , Thanks for reviewing! I will fix them. Since I'm training another task.. I'll test this PR later.

Since it will have similar issues as https://github.com/blue-oil/blueoil/issues/377 . I worry that they can not support various tasks in future. Actually I want to rewrite them using `cv2`. 🤔

iizukak:
opencv-python is big and heavy library.
I don't know it's good idea to use OpenCV for augmentation...

tkng:
> Since it will have similar issues as #377 . I worry that they can not support various tasks in future. Actually I want to rewrite them using cv2. 🤔

Now we don't have dependency to `cv2` module except some demo scripts. Since OpenCV is very large software, we should think carefully before adding that dependency.

yd8534976:
@tkng 
> Now we don't have dependency to cv2 module except some demo scripts. Since OpenCV is very large software, we should think carefully before adding that dependency.

Thanks for answering! I think data augmentation will only be used in training environment, does it matter only add them into `training dependency`?

yd8534976:
@tkng 

Or, we can add some methods like this.. 

https://github.com/dmlc/gluon-cv/blob/c11643d8e81ad08a8c89555a5ad8d4860bb86b19/gluoncv/utils/filesystem.py#L43


```
def try_import_cv2():
    """Try import cv2 at runtime.
    Returns
    -------
    cv2 module if found. Raise ImportError otherwise
    """
    msg = "cv2 is required, you can install by package manager, e.g. 'apt-get', \
        or `pip install opencv-python --user` (note that this is unofficial PYPI package)."
    return try_import('cv2', msg)
```

tkng:
> I think data augmentation will only be used in training environment, does it matter only add them into training dependency?

Considering that adding a dependency to OpenCV will increase several hundreds of mega bytes (or maybe over 1 giga byte) of our docker image size, adding the dependency just for some data augmentation methods seems not a reasonable choice, even only for training environment.

As we already depends on TensorFlow, how about using `tf.contrib.image.rotate`?

yd8534976:
> > I think data augmentation will only be used in training environment, does it matter only add them into training dependency?
> 
> Considering that adding a dependency to OpenCV will increase several hundreds of mega bytes (or maybe over 1 giga bytes) of our docker image size, adding the dependency just for some data augmentation methods seems not a reasonable choice, even only for training environment.
> 
> As we already depends on TensorFlow, how about using `tf.contrib.image.rotate`?

Sorry, I have never tried `tf.contrib.image.rotate`... I just worry if it will make graph more complicated and we may need to define a clean `name_scope` to make it easy to compile.

Well, although rotate augmentation may be helpful, it is not necessary. We may write it by hand or just ignore it.

yd8534976:
> > > I think data augmentation will only be used in training environment, does it matter only add them into training dependency?
> > 
> > 
> > Considering that adding a dependency to OpenCV will increase several hundreds of mega bytes (or maybe over 1 giga bytes) of our docker image size, adding the dependency just for some data augmentation methods seems not a reasonable choice, even only for training environment.
> > As we already depends on TensorFlow, how about using `tf.contrib.image.rotate`?
> 
> Sorry, I have never tried `tf.contrib.image.rotate`... I just worry if it will make graph more complicated and we may need to define a clean `name_scope` to make it easy to compile.
> 
> Well, although rotate augmentation may be helpful, it is not necessary. We may write it by hand or just ignore it. Or rotate them channel by channel...

tkng:
> I just worry if it will make graph more complicated and we may need to define a clean name_scope to make it easy to compile.

I think we can use `tf.contrib.image.rotate` in our data augumentation, without affecting our network definition.

BTW, the core of what I wanted to tell here was, before adding a new dependency to large package, we should consider carefully. Otherwise our software will bloat and be hard to use. `tf.contrib.image.rotate` is just an idea to prevent the bloat, you do not have to be stuck with it.

> Well, although rotate augmentation may be helpful, it is not necessary. We may write it by hand or rotate them channel by channel.

Rotate them channel by channel (if channel number != 3) looks a good idea, because it's simple.

yd8534976:
@tkng Okay! I will pay attention to this point. Thanks for your patience!

PR #372, opened 2019-07-18
Update trainer parameters for checkpoint saving

Joeper214:
## Motivation and Context

Currently, Blueoil saves checkpoints based on lmnet's default config settings.
 This PR gives blueoil users the option to configure how to save the checkpoints from training.
There are 2 trainer parameters that are changed which are:

`SAVE_CHECKPOINT_STEPS` - This parameter determines which step interval should the model be saved.
`KEEP_CHECKPOINT_MAX` - This is the max number of checkpoints that need to be saved throughout the training.

## Description

- Change SAVE_STEPS to SAVE_CHECKPOINT_STEPS and add KEEP_CHECKPOINT_MAX parameters on lmnet and blueoil

## How has this been tested?

This should work by running normal training tasks.

for example:

- In blueoil
`$ ./blueoil.sh train config/object_detection.yml`

- In Lmnet
`$ python lmnet/executor/train.py -c lmnet/configs/example/object_detection.py`

## Types of changes

## Checklist:


Joeper214:
run dlk test

ruimashita:
@Joeper214 
Did you check these cases ?
in blueoil yaml
1. `keep_checkpoint_max` is `0`
2. `keep_checkpoint_max` is `null`

Joeper214:
> @Joeper214
> Did you check these cases ?
> in blueoil yaml
> 
> 1. `keep_checkpoint_max` is `0`
> 2. `keep_checkpoint_max` is `null`

Yes, I test it with these cases. Normally it should.

```
If None or 0, no checkpoints are deleted from 
the filesystem but only the last one is kept in the checkpoint file.
```

As this was the definition stated in tensorflow train.Saver class.

ref: https://www.tensorflow.org/api_docs/python/tf/train/Saver#class_saver

Update in lmnet [doc](https://github.com/blue-oil/blueoil/pull/372/files#diff-ab8772dde073a55dd7aecdb016ffdf37R215):

PR #393, opened 2019-08-16
automatically enable Horovod when multiple GPU IDs are specified

tkng:
This pull request is still work in progress, shutting down the process when training completed is not working correctly. (= We cannot exit the finished processes.)

## Motivation and Context

To train a large neural network, we'd like to use multiple GPUs. This pull request enables us to use multiple GPUs at the training phase of Blueoil.

## Description

Following changes are included:

- Introduced Horovod into the Blueoil Docker image
- If multiple GPUs are specified by `CUDA_VISIBLE_DEVICES`, automaticlly enables Horovod.
- Updated Horovod version to 0.16.4
- Updated OpenMPI version to 4.0.1

## How has this been tested?

There's no unit test for Horovod and it's a little bit hard to write. Hence I tested following situation.

- `CUDA_VISIBLE_DEVICES="0"`
- `CUDA_VISIBLE_DEVICES="1,2"`

In both cases, training was progressed expectedly.

None

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.


tkng:
I think now it's ready to merge, please review the PR.

iizukak:
@nlpng 
Please review this PR.
I know you like Horovod 🙄

lmnet/executor/train.py
```diff
@@ -112,7 +99,7 @@ def start_training(config):
         else:
             loss = model.loss(output, labels_placeholder)
         opt = model.optimizer(global_step)
```
yd8534976: Although I've never tried that. I heard that distributed training had a larger batch size. So the gradient is more accurate. But the update frequency is lower. To get a better performance, we usually need to enlarge learning rate to get a better performance. It should be something like `opt = tf.train.MomentumOptimizer(0.01 * hvd.size())`.

tkng: I has some experiments about that. The technique you mentioned is quite effective.

It would be a good idea to recommend to increase the learning rate in proportion to the batch size. However, automatically changing the learning rate which is set by a user, might confuse the user.

blueoil/blueoil_train.py
```diff
@@ -20,17 +20,23 @@
...
 
 
 def run(blueoil_config_file, experiment_id):
     """Train from blueoil config."""
 
-    # Copy bueoil config yaml.
-    output_dir = os.environ.get('OUTPUT_DIR', 'saved')
-    experiment_dir = os.path.join(output_dir, experiment_id)
-    save_config_file(blueoil_config_file, experiment_dir)
+    if horovod_util.is_enabled():
```
yd8534976: I think this line is duplicated, you will call a `setup()` in below `is_rank0()`.

tkng: Thanks for the comment, there must be a room to improvement.

There're some cases we need to consider:

1. horovod is not installed and user doesn't use horovod
2. horovod is installed but user doesn't use horovod
3. horovod is not installed but use try to use horovod (should be an error in this case)
4. horovod is installed and user try to use horovod

Current code is not well considered to handle these 4 cases, I guess. Also, another problem is that `is_rank0` should not call `setup`.

tkng: I changed the code to exit immediately in the case of 3, since recover is impossible. I think now all cases are handled well.

yd8534976: LGTM.

PR #399, opened 2019-08-22
Modify BDD100K Dataset file add class for segmentation

ananno:
## Motivation and Context

BDD100K dataset file currently designed only for object detection. Dataset class needs to be designed for segmentation also.

## Description

* Modifying label and annotation reading to work with new version of bdd100k
* Added Dataset class for segmentation

## How has this been tested?

Tested using blueoil test framework:
```
$ tox -e pytest
```

## Types of changes

## Checklist:


ananno:
@iizukak San, @ytfksw San :
As you requested, I've added BDD100K dataset for both Object Detection and Segmentation. Please check.

ananno:
@iizukak San:
For some reason it doesn't showing all the changes. Please check the source branch from my [fork](https://github.com/ananno/blueoil/tree/BDD100K_dataset_objdet_seg).

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -1,16 +1,24 @@
 # -*- coding: utf-8 -*-
-import functools
-
-from os.path import join, splitext, basename
+import math
+import os
+import pymp
+import time
```
ytfksw: pymp is not a dependence module of blueoil/lmnet

iizukak:
@ananno 
We need unit test like `test_camvid.py`
https://github.com/blue-oil/blueoil/blob/master/lmnet/tests/lmnet_tests/datasets_tests/test_camvid.py

.gitignore
```diff
@@ -105,3 +105,4 @@ ENV/
 /dataset/
 /config/
 /tmp/
+/.idea/
```
iizukak: @ananno Can you add to your local `~/.gitignore` ?

ananno: OK

ananno:
> @ananno
> We need unit test like `test_camvid.py`
> https://github.com/blue-oil/blueoil/blob/master/lmnet/tests/lmnet_tests/datasets_tests/test_camvid.py

@iizukak San:
I'm on it...

ananno:
Gathering annotation info in Server 5 (8 core CPU):
* With PyMP : 351.13 sec
* Witout PyMP : 1206.71 sec

lmnet/lmnet/datasets/bdd100k.py
tkng: > You can suggest a better way if you have something in mind.

You can use `multiprocessing` module directly, which is a standard library, well tested and don't require extra dependency.

BTW, I measured the performance of parsing annotation files on our 22th server, once json files are on the disk cache, I got the following result:

```
elapsed time: 16.161115646362305  for  70000  images
elapsed time: 2.5325820446014404  for  10000  images
```
Compared to your 1206sec, these numbers are very small 🤔Could you run the experiment twice, to check the effectiveness of disk cache?

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +82,88 @@ def __init__(self,
...
+        if os.path.exists(self.save_file):
+            with open(self.save_file, "rb") as f:
+                loaded_data = pickle.load(f)
+                if len(loaded_data[0]) > 0:
+                    self.paths, self.bboxs = loaded_data
+                else:
+                    os.remove(self.save_file)
+                    self._init_files_and_annotations()
+        else:
+            self._init_files_and_annotations()
```
ytfksw: Please don't save the cache file.

lmnet/lmnet/datasets/bdd100k.py
ananno: @tkng San: 
I talked with @iizukak san and I'll remove the `pymp` package dependency. And you're right once the json is in the cache it will be significantly faster but it also depends on the system. I've tested in server 5, which may be slower than `srv22`

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +82,88 @@ def __init__(self,
...
+        counts = pymp.shared.array(shape=[count], dtype=np.int64)
+        with pymp.Parallel(num_threads=max(1, int(cpu_count()))) as P:
+            # for i, item in enumerate(anno_data):
+            for i in P.range(len(anno_data)):
+                item =  anno_data[i]
+                for label in item['labels']:
+                    class_name = label['category'].replace(' ', '_')
+                    if not item['name'] in img_index:
+                        # print("Label not in images : ", item)
+                        continue
```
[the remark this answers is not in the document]
ananno: I'll remove it in the next commit.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +82,88 @@ def __init__(self,
...
+                    if not class_name in self.classes:
+                        # print("Category not in Selected classes : ", class_name)
+                        continue
```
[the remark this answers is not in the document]
ananno: I'll remove it in the next commit.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -1,16 +1,24 @@
...
+from PIL import Image
+from os.path import join, splitext, basename
 
-from lmnet.datasets.base import ObjectDetectionBase
+from lmnet.datasets.base import ObjectDetectionBase, SegmentationBase
 
 
-class BDD100K(ObjectDetectionBase):
+class BDD100KObjDet(ObjectDetectionBase):
     """BDD100K Dataset for Object Detection (Car Camera)
```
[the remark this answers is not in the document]
ananno: I'll use the full name. Please suggest if you have a better idea.

lmnet/lmnet/datasets/bdd100k.py
ananno: The reason I saved the cached data as `pickle` is for convenience. Every time you start the training with BDD100K the annotation preparation takes about 10min (on server5). 

Anyway, If the caching is not recommended, I can remove the pickle saving and every time training run it will take around 10min (on server 5) to prepare the lists.

lmnet/lmnet/datasets/bdd100k.py
ananno: @tkng:
I've run the training on `lm-srv-05` with the following results:
* 1st run : 
  * 70000 images : 997.87 sec
  * 10000 images : 24.51 sec
* 2nd run :
  * 70000 images : 1001.92 sec  
  * 10000 images : 25.61 sec
* 3rd run [from pickle files] :
  * 70000 images : < 1 sec
  * 10000 images : < 1 sec

lmnet/lmnet/datasets/bdd100k.py
tkng: I wrapped whole lines of `_init_files_and_annotations` by `time.time()`, measured the performance. Here's my result on  the server 5.

```
elapsed time for  70000  images: 8.224702596664429 sec
elapsed time for  10000  images: 2.0470070838928223 sec
```

This performance seems that the data is already on the disk cache, hence it would be the best case. However, even if the data is not on the disk cache, since server 5's storage is SSD, the performance would not be changed so drastically. 10 mins vs 10 secs are very big difference, I want to know what causes this difference. Could you tell me how to reproduce your result?

lmnet/lmnet/datasets/bdd100k.py
tkng: Could you tell me how to reproduce the result? I cannot understand why the result of the 1st and 2nd trial are almost same... 2nd trial should be much faster, as we expect the speed up gain by disk cache...

In my case, you can reproduce by the following steps:
1. login to lmsvr05
2. cd to /home/tokunaga/blueoil/lmnet
3. `PYTHONPATH=. CUDA_VISIBLE_DEVICES="3" python3.6 executor/train.py -c configs/core/object_detection/lm_fyolo_bdd100k.py -i bdd100k_test`

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +96,71 @@ def __init__(self,
...
+        total_count = len(img_paths)
+        print("\rGathering annotation data ... ", end="")
+        counts = 0
+        for i, item in enumerate(anno_data):
+            counts += 1
+            for label in item['labels']:
+                class_name = label['category'].replace(' ', '_')
+                # Skip if Label not in images
+                if not item['name'] in img_index:
+                    continue
```
ytfksw: Since img_index is a (large) list, item['name'] is compared to elements of img_index one by one (about 70,000 times compared for each image and labels). It seems better to use 'set'.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +96,71 @@ def __init__(self,
...
+                # Skip if Classname/Category not in Selected classes
+                if not class_name in self.classes:
+                    continue
+
```
ytfksw: self.classes is not big, but It seems better to use 'set' too.

lmnet/lmnet/datasets/bdd100k.py
ananno: @tkng San:
The timing looks like it is still reading from the pickle file. Did you remove the pickle files?
Please delete the pickle files:
```
# from blueoil/lmnet
$ rm -rf dataset/bdd100k_*.pickle
```
Anyway, I've modified to use dict instead of list as suggested by Hanamoto San. Now without parallel implementation the dataset prepared within 1min (in server 5)

lmnet/lmnet/datasets/bdd100k.py
ananno: @tkng San:
:open_mouth: I'm confused.
Server 5 has 2 GPUs, How can you run with `CUDA_VISIBLE_DEVICES="3"` without an error.

Anyway, Please check the pickle file at `blueoil/lmnet/dataset/bdd100k_{train/val}.pickle`. Please remove them.

ananno:
`pytest` succeeded.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +97,60 @@ def __init__(self,
...
+                x = x1
+                y = y1
+                w = x2 - x1
+                h = y2 - y1
+
+                index = item['name']
+                bbox[index] += [[x, y, w, h, cls_idx]]
+
+            print("\rGathering annotation data ... %d%% [%s/%s] " %
```
a-hanamoto: You don't need to print the status, because this process is very fast if you write codes correctly.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +97,60 @@ def __init__(self,
...
+                y1 = int(round(label["box2d"]["y1"]))
+                y2 = int(round(label["box2d"]["y2"]))
+
+                x = x1
+                y = y1
+                w = x2 - x1
+                h = y2 - y1
+
+                index = item['name']
+                bbox[index] += [[x, y, w, h, cls_idx]]
```
a-hanamoto: Let `N` be the length of `bbox`. What is the computational complexity of this code?

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +97,60 @@ def __init__(self,
...
+
+        total_count = len(img_paths)
+        img_names = list(img_paths.keys())
+
+        print("\rGathering annotation data ... ", end="")
+        counts = 0
+        for item in anno_data:
+            counts += 1
+            # Skip if Label not in images
+            if not item['name'] in img_names:
```
a-hanamoto: Let `N` be the length of `img_names`. What is the computational complexity of this code?

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +97,60 @@ def __init__(self,
...
+                continue
+            for label in item['labels']:
+                class_name = label['category'].replace(' ', '_')
+                # Skip if Classname/Category not in Selected classes
+                if not class_name in self.classes:
```
a-hanamoto: Let `N` be the length of `classes`. What is the computational complexity of this code?

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +97,60 @@ def __init__(self,
...
-                    y2 = int(round(obj["box2d"]["y2"]))
-                    x = x1
-                    y = y1
-                    w = x2 - x1
-                    h = y2 - y1
-
-                    bbox.append([x, y, w, h, cls_idx])
-            bbox = np.array(bbox, dtype=np.int32)
-            self.bboxs.append(bbox)
+        img_paths = OrderedDict([(os.path.basename(path), path) for path in glob.glob(os.path.join(self.img_dir, "*.jpg"))])
```
a-hanamoto: Why do you use `OrderedDict` ?

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +97,60 @@ def __init__(self,
...
+        anno_data = json.load(open(self.anno_dir))
+
+        bbox = [[] for _ in range(len(img_paths))]
+        bbox = OrderedDict(zip(list(img_paths.keys()), bbox))
```
a-hanamoto: Why do you use `OrderedDict` ?

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +97,60 @@ def __init__(self,
...
+
+        total_count = len(img_paths)
+        img_names = list(img_paths.keys())
+
+        print("\rGathering annotation data ... ", end="")
+        counts = 0
+        for item in anno_data:
```
a-hanamoto: What is the computational complexity of this code block?

lmnet/lmnet/datasets/bdd100k.py
ananno: I removed it in next commit.

lmnet/lmnet/datasets/bdd100k.py
ananno: `bbox` being a python dict:
Max Computational Complexity = O(N)
Avg. Computational Complexity = O(1)

lmnet/lmnet/datasets/bdd100k.py
ananno: Max Time Complexity : O(N)
Avg. Time Complexity : O(1)

lmnet/lmnet/datasets/bdd100k.py
ananno: `classes` being a PythonList:
Max Time Complexity : O(N)

lmnet/lmnet/datasets/bdd100k.py
ananno: At the end of the function I assigned `self.path` and `self.bboxs` as:
```        
self.paths = list(img_paths.values())
self.bboxs = list(bbox.values())
```
By using OrderedDict I can ensure the order of the `img_paths.values()` and `bbox.values()`. Normal Dict is not Ordered (as per my knowledge).

lmnet/lmnet/datasets/bdd100k.py
ananno: Same as above

lmnet/lmnet/datasets/bdd100k.py
ananno: For a `for` loop: O(N) where N being the size of `anno_data`

lmnet/lmnet/datasets/bdd100k.py
antonionevado: @tkng Only mention one detail that maybe helps. 
On our servers we usually have the next configuration:
```
/  is mounted on SSD
/storage  is mounted on a traditional HDD
/storage2  is mounted on a traditional HDD
```
Normally, we store the datasets on `/storage` or `/storage2` so we are always reading from HDD drives during training. Always wondered if performance will improve if we store the dataset on the SSD. Or we actually keep most of the dataset in memory during training?

Not sure how it works, actually.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -26,7 +28,26 @@ class BDD100K(ObjectDetectionBase):
                "truck"]
     num_classes = len(classes)
     available_subsets = ["train", "validation"]
-    extend_dir = "bdd100k"
+    extend_dir = "BDD100K/bdd100k"
```
tkng: I guess this change requires extra work to make a new directory `BDD100K`, user would not be able to understand what they should do to make this dataset class works, without reading the source.

tkng:
We cannot add image files from BDD100K without copyright notice.
Could you check the license of BDD100K and follow the license term?

lmnet/lmnet/datasets/bdd100k.py
ananno: @tkng San:
I thought about that too. But when you unzip the `BDD100K.zip` data file after downloading from website, it will create a `bdd100k` folder. So, as before may be using only `bdd100k` is better. What do you think?

ananno:
> We cannot add image files from BDD100K without copyright notice.
> Could you check the license of BDD100K and follow the license term?

I've checked the licence agreement of `BDD100K` dataset. As I understand, it is granted to use, copy and distribute for non-commercial use. Here is the full term:
```
Copyright ©2018. The Regents of the University of California (Regents). All Rights Reserved. 

Permission to use, copy, modify, and distribute this software and its documentation for educational, research, and not-for-profit purposes, without fee and without a signed licensing agreement; and permission use, copy, modify and distribute this software for commercial purposes (such rights not subject to transfer) to BDD member and its affiliates, is hereby granted, provided that the above copyright notice, this paragraph and the following two paragraphs appear in all copies, modifications, and distributions. Contact The Office of Technology Licensing, UC Berkeley, 2150 Shattuck Avenue, Suite 510, Berkeley, CA 94720-1620, (510) 643-7201, otl@berkeley.edu, http://ipira.berkeley.edu/industry-info for commercial licensing opportunities.

IN NO EVENT SHALL REGENTS BE LIABLE TO ANY PARTY FOR DIRECT, INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES, INCLUDING LOST PROFITS, ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN IF REGENTS HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.

REGENTS SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE. THE SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED "AS IS". REGENTS HAS NO OBLIGATION TO PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS. 
```

lmnet/lmnet/datasets/bdd100k.py
ananno: as `img_index` is not need to preserve order `set` can be used for performance gain. I'll update.

lmnet/lmnet/datasets/bdd100k.py
ananno: @tkng San:

> NOTE : I appologize for long discussion.

I've tried to reproduce your case with the current master branch of blueoil on server 5. Here is my result.
* The training seems to be running but the dataset clearly not populated with data. As in the error, for both training and validation dataset_iterator gives the following error:
```
Exception in thread Thread-1:
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/hasibur/blueoil/lmnet/lmnet/datasets/dataset_iterator.py", line 163, in run
    self.loop_body()
  File "/home/hasibur/blueoil/lmnet/lmnet/datasets/dataset_iterator.py", line 139, in loop_body
    task_list = self.gen_task(self.dataset.batch_size * 8)
  File "/home/hasibur/blueoil/lmnet/lmnet/datasets/dataset_iterator.py", line 123, in gen_task
    data_id = self.data_ids.pop()
IndexError: pop from empty list

Exception in thread Thread-5:
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/hasibur/blueoil/lmnet/lmnet/datasets/dataset_iterator.py", line 163, in run
    self.loop_body()
  File "/home/hasibur/blueoil/lmnet/lmnet/datasets/dataset_iterator.py", line 139, in loop_body
    task_list = self.gen_task(self.dataset.batch_size * 8)
  File "/home/hasibur/blueoil/lmnet/lmnet/datasets/dataset_iterator.py", line 123, in gen_task
    data_id = self.data_ids.pop()
IndexError: pop from empty list
```
* The reason of the error is in the `blueoil/lmnet/dataset/bdd100k.py` dataset file the `_init_files_and_annotations` function expect the annotation in json file for each image with the same filename but if you download the data from BDD website and extract, instead of getting individual json for each image you'll get 1.5GB json files, one for training and one for validation. So, the `bdd100k.py` dataset file in current blueoil master branch no longer works.

But we already identified the bottleneck and improved performance. Please checkout the current update of my pull request, now it takes less than 1 min. The biggest improvement we've got is as follows:
* **Changing `List` -> `Dict`** : From Python 3.5 dict is significantly improved in performance
* **Changing `List` -> `Set`** : In case where order is not required set gives more performance by using hash matching.

lmnet/lmnet/datasets/bdd100k.py
ananno: This problem is already solved. 
Please see this comment : https://github.com/blue-oil/blueoil/pull/399#discussion_r318364484

lmnet/lmnet/datasets/bdd100k.py
ananno: After testing the effect is insignificant. So we can use `self.classes` to ensure the order for consistent  `class_id`.

lmnet/lmnet/datasets/bdd100k.py
ytfksw: OK! Thank you!

lmnet/lmnet/datasets/bdd100k.py
ytfksw: I got it.

iizukak:
@ananno 
Please add license description of the fixture datasets.

ananno:
> @ananno
> Please add license description of the fixture datasets.

@iizukak San:
The license applies to all the assets of bdd100k dataset. So I put it on the parent directory `tests/fixtures/BDD100K`

Please suggest if the location is not good.

iizukak:
@ananno OK Thanks.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +94,54 @@ def __init__(self,
...
+                y2 = int(round(label["box2d"]["y2"]))
+
+                x = x1
+                y = y1
+                w = x2 - x1
+                h = y2 - y1
+
+                index = item['name']
+                bbox[index] += [[x, y, w, h, cls_idx]]
+        print()
```
ytfksw: Now, this print function is not necessary. Could you remove this part?

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +94,54 @@ def __init__(self,
...
-                    bbox.append([x, y, w, h, cls_idx])
-            bbox = np.array(bbox, dtype=np.int32)
-            self.bboxs.append(bbox)
+        img_paths = OrderedDict([(os.path.basename(path), path)
+                                 for path in glob.glob(os.path.join(self.img_dir, "*.jpg"))])
+        img_names = set(img_paths.keys())
+
+        anno_data = json.load(open(self.anno_dir))
+        bbox = [[] for _ in range(len(img_paths))]
+        bbox = OrderedDict(zip(list(img_paths.keys()), bbox))
```
ytfksw: Find a better algorithm without ordered dictionary.

ananno: Implemented without `OrderedDict` in commit 643ec047bbd649365df0bcbb42dee57387d4f651

lmnet/lmnet/datasets/bdd100k.py
ananno: Removed

lmnet/lmnet/datasets/bdd100k.py
ananno: It is resolved in commit 643ec047bbd649365df0bcbb42dee57387d4f651

lmnet/lmnet/datasets/bdd100k.py
ananno: Resolved in commit : 643ec047bbd649365df0bcbb42dee57387d4f651

lmnet/lmnet/datasets/bdd100k.py
ananno: Finally I've used `extended_dir = "bdd100k"` in last commit 643ec047bbd649365df0bcbb42dee57387d4f651

ananno: Resolved in commit 643ec047bbd649365df0bcbb42dee57387d4f651

ananno:
@ytfksw San: Can you please review again? It passed all check.

ananno:
@ytfksw San: Can you review again please. If everything is OK I would like to remove the WIP tag from the title.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -126,13 +165,135 @@ def __len__(self):
...
+    def __len__(self):
+        return len(self.files_and_annotations()[0])
+
+    @property
+    def num_per_epoch(self):
+        return len(self.files_and_annotations()[0])
+
+
 def check_dataset():
-    train = BDD100K(subset="train")
```
ytfksw: @ananno Could you remove the debug function?

ananno: I removed in commit : 3abf09428f3e38d3f19ecfbac8351a814d87b0a6

ytfksw:
I got an error when I executed the sample detection config.
```
  File "executor/train.py", line 432, in <module>
    main()
  File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/clic
k/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/cli$k/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/cli$k/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/cli$k/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "executor/train.py", line 428, in main
    run(network, dataset, config_file, experiment_id, recreate)
  File "executor/train.py", line 393, in run
    start_training(config)
  File "executor/train.py", line 350, in start_training
    summary, _ = sess.run([summary_op, metrics_update_op], feed_dict=feed_dict)
  File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/ten$orflow/python/client/session.py", line 929, in run
    run_metadata_ptr)
  File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/ten$orflow/python/client/session.py", line 1128, in _run
    str(subfeed_t.get_shape())))
ValueError: Cannot feed value of shape (18, 66, 5) for Tensor 'labels_placeholder:0', w$ich has shape '(18, 91, 5)'
```

ananno:
> I got an error when I executed the sample detection config.
> 
> ```
> Traceback (most recent call last):
>   File "executor/train.py", line 432, in <module>
>     main()
>   File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/clic
> k/core.py", line 722, in __call__
>     return self.main(*args, **kwargs)
>   File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/cli$k/core.py", line 697, in main
>     rv = self.invoke(ctx)
>   File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/cli$k/core.py", line 895, in invoke
>     return ctx.invoke(self.callback, **ctx.params)
>   File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/cli$k/core.py", line 535, in invoke
>     return callback(*args, **kwargs)
>   File "executor/train.py", line 428, in main
>     run(network, dataset, config_file, experiment_id, recreate)
>   File "executor/train.py", line 393, in run
>     start_training(config)
>   File "executor/train.py", line 350, in start_training
>     summary, _ = sess.run([summary_op, metrics_update_op], feed_dict=feed_dict)
>   File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/ten$orflow/python/client/session.py", line 929, in run
>     run_metadata_ptr)
>   File "/storage/fukasawa/.pyenv/versions/bdd100k_check/lib/python3.6/site-packages/ten$orflow/python/client/session.py", line 1128, in _run
>     str(subfeed_t.get_shape())))
> ValueError: Cannot feed value of shape (18, 66, 5) for Tensor 'labels_placeholder:0', w$ich has shape '(18, 91, 5)'
> ```

I've fixed the issue in commit : 2c58c21
Please review one more time.

ytfksw:
@ananno 
Since these config files will be published, can you adjust the config for object detection and segmentation?
A gradient explosion happens in the current object detection settings.

ananno:
> @ananno
> Since these config files will be published, can you adjust the config for object detection and segmentation? A gradient explosion happens in the current object detection setting.

@ytfksw San:
I'm running training with BDD100K object detection in `lmdgx01`. I'll update as it progresses.

ananno:
@ytfksw San:

# Training Observations
* First Test Step took ~45min with 1000 test images

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -126,13 +161,127 @@ def __len__(self):
...
+        label_paths.sort()
+
+        return image_paths, label_paths
+
+    def __getitem__(self, i):
+        imgs, labels = self.files_and_annotations()
+        img = load_image(imgs[i])
+        label = load_image(labels[i])
+
+        return np.array(img), np.array(label)
```
ytfksw: np.array is not necessary because load_image retrun np.array

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -28,6 +27,25 @@ class BDD100K(ObjectDetectionBase):
...
+
+        return np.array([
+            bike, bus, car, motor,
+            person, rider, traffic_light,
+            traffic_sign, train, truck
+        ])
+
     @classmethod
     @functools.lru_cache(maxsize=None)
     def count_max_boxes(cls, base_path=None):
```
ytfksw: Now, this function is not necessary.

lmnet/lmnet/datasets/bdd100k.py
ananno: Yes. But what happened if someone want to change the `load_image` function implementation? Also, I think `np.array` of an numpy array doesn't do anything but just return the array.

If you think we should remove it, I can remove it.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -47,7 +65,7 @@ def count_max_boxes(cls, base_path=None):
     def num_max_boxes(self):
         # from dataset:
         # train - max 91 boxes, val - max 66 boxes
-        return self.max_boxes
```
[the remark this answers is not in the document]
ananno: This is old code. I already updated with hard coded max value `91` here.

lmnet/lmnet/datasets/bdd100k.py
ananno: The function `count_max_boxes` is there for conventional purposes. As you dig deeper sometimes, this function is called from the base class e.g. in `pascalvoc_2007` dataset. I think we should keep this to ensure operation of the BDD100K dataset class throughout the project. If you go to other object detection dataset all of them have this function.

lmnet/lmnet/datasets/bdd100k.py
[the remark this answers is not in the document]
ytfksw: self.max_boxes and max_boxes are defined and are used elsewhere. It is looks odd for me: 

https://github.com/blue-oil/blueoil/blob/3bdb78dec90e16360e5acb5751277cc0b40f7e46/lmnet/lmnet/datasets/bdd100k.py#L90-L92

https://github.com/blue-oil/blueoil/blob/3bdb78dec90e16360e5acb5751277cc0b40f7e46/lmnet/lmnet/datasets/bdd100k.py#L145-L147

https://github.com/blue-oil/blueoil/blob/3bdb78dec90e16360e5acb5751277cc0b40f7e46/lmnet/lmnet/datasets/bdd100k.py#L77-L79

lmnet/lmnet/datasets/bdd100k.py
ytfksw: I think this is not only a problem of BDD100K.  Just leave it as you say.

lmnet/lmnet/datasets/bdd100k.py
ytfksw: Docstring says that  "Returns numpy array of an image".
You don’t have to worry about it, because you created a test code.
Please remove np.array.

lmnet/lmnet/datasets/bdd100k.py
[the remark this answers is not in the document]
ananno: I don't understand your confusion. It is very logical if you think about the history and the issues of the blueoil development. Let me explain it for you:

First of all, in the `__init__` function the argument `max_boxes` has default value `100`. This is defined in old implementation of `bdd100k` dataset file, if you like you can check the commit : cc0e69a51a2782d4cf9880296a697c11cf5ed27f.

Previously, by default the max boxes was not in many object detection dataset class and the property `num_max_boxes` which actually returns the maximum number of boxes in one image by using `count _max_boxes` function to set the limit of the `numpy` array dimension which results in an uniform data structure for the training and validation purposes, which must have a fixed dimension. 

Now `count_max_boxes` function loads the Dataset classes for both `train` and `validation` set to determine the maximum number of boxes. In normal case this is OK. But in case of very large dataset like BDD100K, where we had performance issue before it was a waste of time to load the whole dataset twice just to determine the `num_max_boxes` property. Hence following the example of `widerface` dataset implementation a new variable `self.max_boxes` has been created which have the default value of 100 to keep the convension and can be modified from outside during initialization of the dataset class. The true number of `self.max_boxes` is being calculated in new implementation `bdd100k` dataset in the initialization phase using the code block:
```
num_boxes = len(bbox)
            if num_boxes > 0:
                self.paths.append(img_paths[img_name])
                self.bboxs.append(bbox)
                if num_boxes > max_boxes:
                    max_boxes = num_boxes
```
in `_init_files_and_annotations` function.

This has 1 problem, as during training there are 2 dataset in play, `train` and `validation`. By counting the number with the above code block does not consider both dataset at once, as using `count_max_boxes` function dose. So to avoid having wrong dataset array dimension for  one of the dataset the variable `self.max_boxes` is hard coded to `91` which is the max number of boxes in the dataset found by the code block mentioned above. This was also counted in previous implementation of `bdd100k` dataset and commented in the `num_max_boxes` property as you can check in commit : cc0e69a51a2782d4cf9880296a697c11cf5ed27f. The previous implementation set the default value 100 for the `self.max_boxes` which is adequate for both train and validation set.

But as we're using the maximum allowed image size, in order to optimize the memory footprint I hard coded the value to `91` while keeping the code block which calculate actual number of boxes for convenient purposes.

As you know, now we don't have the performance issue with the `BDD100K` dataset the original `count_max_boxes` function can be used, but it will load both the `train` and `validation` dataset twice, it will cost memory in the beginning of the training and might claim a large amount of `swap` space depending on available RAM. To avoid this issue the hard coded value `91` is kept and also the function `count_max_boxes` for keep the convension as it is being used or can be used for user side implementation in future.

I hope now it would be clear to you.

lmnet/lmnet/datasets/bdd100k.py
ananno: Fixed in commit : 1269d5665bed5a68882d91710b8e5cc579281ae5

lmnet/lmnet/datasets/bdd100k.py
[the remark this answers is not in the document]
ytfksw: I knew that the maximum number between validation and train was different because you explained. Although there are conventions as you say, in the current implementation the default value of `max_boxes` and `self.max_boxes` are unused variables. I think that the default value of max_boxes and self.max_boxes should be removed and can be fixed at 91. On the other hand, count_max_boxes is defined in the parent class, so it is unnatural if not implemented.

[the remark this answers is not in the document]
ytfksw: I knew that the maximum number between validation and train was different because you explained. I understand there are conventions. But in the current implementation, the default value of max_boxes and self.max_boxes are unused variables. It looks odd to me. I think that the default value of max_boxes and self.max_boxes should be removed and can be fixed at 91. On the other hand, count_max_boxes is defined in the parent class, so it is unnatural if not implemented.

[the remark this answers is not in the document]
ananno: @tkng San:
I agree with you. I also think `count_max_boxes` function can be removed from the `ObjectDetectionBase` class defined in `lmnet/lmnet/datasets/base.py` file, so that this function can be ignored if needed.

I also agree with you that we should not try to resolve multiple problem within one pull request. This pull-request is getting too much changes.

As a side note, from your suggestion of using self.max_boxes assigned in `_init_file_and_annotations` function but as I explained in my previous comment, by this way two datasets, `train` and `validation` dataset will have different number of array dimension. E.g. for this BDD100K dataset `train` set has maximum boxes of 91 and `validation` dataset will have 66 max boxes. This will create discrepancy while feeding the data into the label `placeholder`.

I think we should hardcode the `self.max_boxes` per dataset basis, as this is a constant value per dataset.

ananno:
@ytfksw @a-hanamoto San
This pull request got very dirty and contains lots of changes. Which is very annoying and misleading to review. To clean this up I propose the followings:

* Divide the changes of this pull request to 3 or 4 smaller pull request with relevant changes
* Remove code changes that moved to other pull request from this pull request
* Carefully merge all pull request by maintaining dependency order

Please suggest any better way to clean up and in corporate changes to the blueoil repo. If this idea is OK please add tokunaga san or someone from blueoil team to overview.

a-hanamoto:
@ananno If the remained tasks are small, please do them in this PR. If not, you can divide.

ananno:
> @ananno If the remained tasks are small, please do them in this PR. If not, you can divide.
@a-hanamoto San:
There are no new task in this PR. In fact there are few changes which are out of scope. Anyway, as you suggested after testing the code I'll remove WIP status.

Thank you

ananno:
@ytfksw San:
As suggested by @a-hanamoto san, the task related to this pull request is done and it also passed all check. Please review one more time and if all is OK put a comment `LGTM`.

ytfksw:
@ananno LGTM!

ananno:
@iizukak San:
@ytfksw San already approved the pull request. Please review one more time.

Thank you.

ananno:
@iizukak San, @suttang San:
This pull request is waiting for a while (about 2 weeks) for review. Please review in your earliest convenient time.

Thank you.

iizukak:
@ananno Sorry for delaying the review.

This PR looks too large for review.
Can you split this PR for several smaller PRs?

PR1. Fixture dataset
PR2. Dataset Loader and test for Object Detection
PR3. Config file for Object Detection
PR4. Dataset Loader and test for Semantic Segmentation
PR5. Config file for Semantic Segmentation

2&3 and 4&5 may need to put together.

Most of the code is LTGM. Thanks.

ananno:
> @ananno Sorry for delaying the review.
> 
> This PR looks too large for review.
> Can you split this PR for several smaller PRs?
> 
> PR1. Fixture dataset
> PR2. Dataset Loader and test for Object Detection
> PR3. Config file for Object Detection
> PR4. Dataset Loader and test for Semantic Segmentation
> PR5. Config file for Semantic Segmentation
> 
> 2&3 and 4&5 may need to put together.
> 
> Most of the code is LTGM. Thanks.

This is exactly what I wanted to do as I suggested [here](https://github.com/blue-oil/blueoil/pull/399#issuecomment-546875334).

Anyway, I'll do that.

iizukak:
@ananno 
Sorry, I missed the discussion.

One question. can you teach me where is the original BDD100k license file?

ananno:
> @ananno
> Sorry, I missed the discussion.
> 
> One question. can you teach me where is the original BDD100k license file?

@iizukak San:
The original license file is available during the download option. You've to have an account (free to create) to download bdd100k dataset files. Here is the screenshot of that:
Screen Shot 2019-11-18 at 14 38 55

iizukak:
@ananno Thanks. I understood.

ananno:
This PR is not necessary anymore. It is divided into smaller PRs and all are merged.
PRs: #628 #630, #631, #632, #633

Issue #397, opened 2019-08-20
Fix round


yd8534976:
- In `linear.py', 'linear_mid_tread_half_quantizer()` uses `quantized = tf.round(shifted * n) / n`.
- In `quantizer.h`, `func_QTZ_binary_channel_wise_mean_scaling` uses `tmp = roundf(tmp * n) / n`.

Where, `tf.round()` will round half to even. So `tf.round(2.5) = 2.0`. Yet `roundf(2.5) = 3.0`.

primenumber:
We can use std::nearbyint/nearbyintf.
Behavior of these functions depends on rounding mode.
If current rounding mode is FE_TONEAREST, it is recommended to round half to even and glibc is so.

PR #400, opened 2019-08-23
Fix round

yd8534976:
https://github.com/blue-oil/blueoil/issues/397

Inconsistency of `tf.round()` and `roundf()` may cause unexpected computation errors.

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/core/operators.py
+++ b/dlk/python/dlk/core/operators.py
@@ -1446,7 +1446,7 @@ class QTZ_linear_mid_tread_half(Quantizer):
 
         n = 2 ** bit - 1
         np.clip(in_data, 0, max_value, out=in_data)
-        kwargs['data'] = np.round(in_data * n / max_value).astype(np.int32)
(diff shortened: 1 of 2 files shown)


dlk/python/dlk/core/operators.py
```diff
@@ -1446,7 +1446,7 @@ def run(self, **kwargs) -> Dict:
 
         n = 2 ** bit - 1
         np.clip(in_data, 0, max_value, out=in_data)
-        kwargs['data'] = np.round(in_data * n / max_value).astype(np.int32)
+        kwargs['data'] = np.floor(in_data * n / max_value + 0.5).astype(np.int32)
```
nlpng: @yd8534976 
I am not sure, I thought the problem is the `tf.round`, why you need to fix this, as it is using `np.round`.

yd8534976: @nlpng Both `np.round` and `tf.round` round half to even, so I change both of them. And in `pass_lookup()`, this `run() func` will be called. Although in `pass_lookup()` it may have no obvious effect. We need to change it for consistency.

PR #401, opened 2019-08-23
[WIP] Add encode_image decorator

yd8534976:
- Add encode_image() decorator 

how to use it?

```
class LMFYolo(YoloV2):

    @encode_image
    def base(self, images, is_training):
        pass
```

## Types of changes

## Checklist:


tkng:
@yd8534976 
Thanks for implementing this feature! However, from the following reasons, it looks better to implement `encode_image` as just a simple function, not as a decorator. (Or, there might be a better idea, but I don't have such an idea at the moment.) Could you modify the PR?

- In the future, we might want to receieve 2 inputs (like video frame and audio frame) at the same time. 
- In my opinion, in general, decorator should only be used when it is very convenient. [Google's style guide](http://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators) also recommend to "Use decorators judiciously".

yd8534976:
@tkng 
Thanks for comments! I think this implementation is the best way to avoid changing every `base() ` method for now. But it does have some problems. Sorry, it is an attempt out of curiosity.. I will change it later.

PR #402, opened 2019-08-29
Port Packer class to Python

tkng:
Port Packer class to Python from C++

## Motivation and Context

Currently, Packer class is written in C++, therefore the code under the Packer class is somewhat complicated to call C++ code from Python. Also, build process is not simple because we need to buila an extention module.

On the other hands, what the Packer class do is very simple, just packing the array of some small integer values into an integer array. For example, the class packs input array `[1,0,1,1]` to an integer array `[13]` (13 = 0b1101)

I'm not very happy to maintain over 200 lines of C ++ code for this simple feature.

## Description

- new Packer class is written in Python
- Old Packer related C++ code is removed
- Trade off is that now the conversion is 2-5 seconds slower than old C++ code

## How has this been tested?

unit test

None

## Types of changes

## Checklist:


tsawada:
I think this can be cleaner... but I don't have concrete ideas.
Python Readability Approval.

PR #403, opened 2019-08-30
Add unit tests for Packer class

tkng:
Add unit tests for Packer class

## Motivation and Context

before merging #402, add some test cases to ensure that the new Packer class is working correctly.

## Description

- add some test cases for Packer class

## How has this been tested?

This PR itself is adding of unit tests.

## Types of changes

## Checklist:
- [ ] My code follows the code style of this project.
-->


dlk/tests/test_packer.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_bw1_dividable_by_wordsize(self):
+        """Test for when the input tensor size is able to divide by wordsize (1 bit version)."""
+        packer = Packer(1, 32)
+
+        test_input = np.zeros([32], dtype=np.float32)
+        test_input[0:6] = [0, 1, 0, 1, 0, 1]
+
+        test_output = packer.run(test_input)
+        print(test_output)
```
tsawada: please remove

dlk/tests/test_packer.py
```diff
@@ -0,0 +1,66 @@
...
+        packer = Packer(2, 32)
+
+        test_input = np.zeros([37], dtype=np.float32)
+        test_input[0:-1:2] = 1
+        test_input[0:-1:4] = 2
+
+        test_output = packer.run(test_input)[0]
+        expectged_output = np.array([1145324612, 286331153, 4294967268, 4294967265], dtype=np.uint32)
+
+        self.assertTrue((test_output == expectged_output).all())
```
tsawada: Maybe it's better to use `assert_array_equal`?
https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_array_equal.html#numpy.testing.assert_array_equal

dlk/tests/test_packer.py
tkng: done!

tkng:
> do we not need non-divisible cases?

added.

dlk/tests/test_packer.py
tkng: I choose `.all()` here, to be consisitent to other tests.

iizukak:
@tkng This PR has some comflict.

tkng:
I changed some parts, could you review again?

- I found that old (C++) implementation has a problem, handling of extra bits seems not working correctly.
  - As a consequence, some unit tests proposed in this PR were failed.
  - Also, in the `optimizer.py`, Lookup operator was filtering higher bits of the result of `packer.run` to avoid the weired behavior.  (it's just my guess.)

I fixed these problems, updated some unit tests. Also, `packer.py` was simplified a bit.

iizukak:
I checked 74493870720d195d0ba9c34bebfae5e24fa86ef1 and cda28690dc34ed569835bd138bcfc7a726de756a .

I think this commits are OK to merge.

BTW, It's difficult to guess why programmer write mask `0x000003FF` for both `msb` and `lsb` 
 for me...

tkng:
> I think these commits are OK to merge.

thanks!

> BTW, It's difficult to guess why programmer write mask `0x000003FF` for both `msb` and `lsb`
> for me...

I guess that the programmer found that higher bits of result of `packer.run` are broken, but old packer code was so complicated, so he or she gave up fundamental solution, applied a mask to filter out the broken higher bits as a temporary repair. It's just my guess, though...

iizukak:
Understood...

PR #440, opened 2019-09-18
update TF version to 1.14.0, Horovod version to 0.18.1

tkng:
## Motivation and Context

TensorFlow 1.14.0 warn more deprecated functions than 0.13.1. We should fix such warnings before upgrading to TF 2.0.

## Description
- upgrade TF version to 1.14.0
- A new dependency `gast==0.2.2` is necessary for TF 1.14.0. (recently released `gast` `0.3.x` is not compatible with `0.2.x`, that cause some unreasonable warning messages.)
- Horovod is also updated to the latest version, since TF 1.14.0 seems sometimes not compatible with Horovod 0.16.4. (For more detail, see https://github.com/horovod/horovod/issues/1232)
 
## How has this been tested?

- unit test
- by running some training scripts

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/cpu.requirements.txt
+++ b/lmnet/cpu.requirements.txt
@@ -1,2 +1,3 @@
 -r dev.requirements.txt
-tensorflow==1.13.1
+tensorflow==1.14.0
+gast==0.2.2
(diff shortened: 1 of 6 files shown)


lmnet/cpu.requirements.txt
```diff
@@ -1,2 +1,3 @@
 -r dev.requirements.txt
-tensorflow==1.13.1
+tensorflow==1.14.0
+gast==0.2.2
```
iizukak: Why tensorflow automatically solve dependency of `gast`...
Anyway LGTM.

tkng: I guess TF 1.14.0 just uses the latest gast, because TF people doesn't expected the incompatible update of `gast`.

Issue #420, opened 2019-09-12
Too many duplicated load_image(filename) functions


yd8534976:
A bunch of load_image(filename) with almost same functionality are implemented everywhere. We should have a module to store this kind of functions.

```
./output_template/python/run.py:    img = Image.open(input_image).convert("RGB")
./output_template/python/lmnet/utils/output.py:                mask_image = PIL.Image.open(BytesIO(mask_data))
./dlk/utils/run_inference.py:    img = Image.open(input_image_path).convert('RGB')
./dlk/utils/run_test.py:        with Image.open(image_file).convert("RGB") as img:
./dlk/docs/python/shared_library.md:    img = Image.open(input_image_path).convert('RGB')
./lmnet/tests/lmnet_tests/test_data_augmentor.py:    image = PIL.Image.open("tests/fixtures/sample_images/cat.jpg")
./lmnet/tests/lmnet_tests/datasets_tests/test_camvid.py:    raw_image = Image.open(raw_image_path).convert("L")
./lmnet/tests/lmnet_tests/util_tests/test_json.py:            mask_pil_image = PIL.Image.open(BytesIO(mask_data))
./lmnet/executor/predict.py:            tmp_image = PIL.Image.open(filename)
./lmnet/executor/export.py:    tmp_image = PIL.Image.open(filename)
./lmnet/executor/measure_latency.py:    tmp_image = PIL.Image.open(filename)
./lmnet/lmnet/datasets/open_images_v4.py:        image = PIL.Image.open(filename)
./lmnet/lmnet/datasets/cityscapes.py:        img = Image.open(imgs[i])
./lmnet/lmnet/datasets/cityscapes.py:        label = Image.open(labels[i])
./lmnet/lmnet/datasets/camvid.py:    image = Image.open(filename)
./lmnet/lmnet/datasets/image_folder.py:        image = PIL.Image.open(filename)
./lmnet/lmnet/datasets/mscoco.py:        image = PIL.Image.open(filename)
./lmnet/lmnet/datasets/ilsvrc_2012.py:        image = PIL.Image.open(filename)
./lmnet/lmnet/datasets/delta_mark.py:        image = PIL.Image.open(target_file)
./lmnet/lmnet/datasets/base.py:        image = PIL.Image.open(target_file)
./lmnet/lmnet/datasets/base.py:        image = PIL.Image.open(target_file)
./lmnet/lmnet/utils/json.py:                mask_image = PIL.Image.open(BytesIO(mask_data))

```

PR #449, opened 2019-09-20
Unify load_image() in lmnet

yd8534976:
As #420 , to reduce duplicated code and manage them easily. I unify all load_image()-like functions into a `load_image()` in `utils/image.py`. I only changed obvious RGB format cases.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:


ruimashita:
@yd8534976 
Thanks 👍
In `pre_processor.py`,  we should implement class that can be used with` PRE_PROCESSOR` in config file, I think.
How about `load_image()` in `utils/image.py` ?

yd8534976:
@ruimashita  Thanks! If so, I may have to add an extra  link in `output_template` to `utils/image.py`. I'm not sure which one is better. 🤔

ruimashita:
> I may have to add an extra link in output_template to utils/image.py.

yes, we need to link in `output_template/utils/image.py` to `lmnet/lmnet/utils/image.py` for `output_template/python/run.py`

output_template/python/lmnet/utils/image.py
```diff
@@ -0,0 +1 @@
+lmnet/lmnet/utils/image.py
```
ruimashita: I think this path should be relative path from  `output_template/python/lmnet/utils/image.py`, Maybe `../../../../lmnet/lmnet/utils/image.py` is correct path 🤔

yd8534976: @ruimashita Thanks! Fixed.

PR #454, opened 2019-09-20
Well managed file descriptor and mmap-ed memory

primenumber:
Implement class for manage file descriptor / mmap-ed memory for dlk.

## Motivation and Context

In current implementation, file descriptor and mmap-ed memory are not managed, so resource leakage occurs.
This PR fixes it.

## Description

To manage file descriptor, FDManager
To manage mmap-ed memory, SimpleMappedMem

I considered to use std::unique_ptr:
Type of file descriptor is not pointer, so we have to implement custom class for store file descriptor which can be compare with `std::nullptr_t`.
`munmap()` require mapped size, so we have to implement custom class for store pointer with mapped size.
These increase implement cost.

## How has this been tested?

```shell
$ PYTHONPATH=python/dlk python python/dlk/scripts/generate_project.py -i minimal_graph_with_shape.pb -o tmp/ -p object_detection -hq -ts -cache
$ cd tmp/object_detection.prj
$ mkdir build
$ cd build
$ cmake .. -DTOOLCHAIN_NAME=linux_arm -DUSE_NEON=1 -DRUN_ON_FPGA=1
$ make -j8 lm
```

## Types of changes

## Checklist:


dlk/python/dlk/templates/include/memdriver.h
```diff
@@ -21,11 +21,55 @@ limitations under the License.
 #include <stdint.h>
 #include <fcntl.h>
 #include <unistd.h>
+#include <memory>
+#include <system_error>
+
+class FDManager {
```
tkng: For me, this `FDManager` class represents file descriptor itself, rather than a manager. How about renaming this class to just `FD` or `FileDescriptor` or something?

If you are uncomfortable using such a common name, you can also use namespace.

dlk/python/dlk/templates/include/memdriver.h
```diff
@@ -21,11 +21,55 @@ limitations under the License.
...
+    if (fd >= 0) {
+      close(fd);
+    }
+  }
+  operator int() const { return fd; }
+ private:
+  int fd;
+};
+
+class SimpleMappedMem {
```
tkng: Since `MappedMem` class seems not be used in the current code, how about removing current `MappedMem` class and renaming this `SimpleMappedMem` class to `MappedMem` class?

dlk/python/dlk/templates/include/memdriver.h
primenumber: OK, fixed.

dlk/python/dlk/templates/include/memdriver.h
primenumber: Fixed.

PR #461, opened 2019-09-24
Remove LmThingsOnATable class

### LINKED ISSUE #433: remove LmThingsOnATable class
`LmThingsOnATable` class is currently used only from test files (test_yolo_v1.py, test_yolo_v2.py, etc.) To reduce maintenance cost, it would be better if we can remove this class.

Joeper214:
## Motivation and Context

- Fix #433 
- Remove LmThingsOnATable class and replace it with PascalVOC2007
- Pass existing tests

## How has this been tested?

- This should be tested with `pytest`

## Types of changes

- [x] Refactoring

## Checklist:


Joeper214:
@iizukak @tkng Please review if you have time. 🙇

iizukak:
@Joeper214 
Of course 👌

tkng:
> `+ lmnet/tests/fixtures/datasets/open_images_v4/images/val/000026e7ee790996.jpg`

Do you need to add this file?

Joeper214:
> > `+ lmnet/tests/fixtures/datasets/open_images_v4/images/val/000026e7ee790996.jpg`
> 
> Do you need to add this file?

It's not necessary, I'll update in the next commit.

lmnet/tests/lmnet_tests/datasets_tests/test_image_folder.py
```diff
@@ -75,7 +75,7 @@ def test_image_folder():
 
 class DummyHasValidation(ImageFolderBase):
     extend_dir = "dummy_classification"
-    validation_extend_dir = "lm_things_on_a_table"
+    validation_extend_dir = "open_images_v4"
```
tkng: `OpenImageV4` class doesn't derived from `ImageFolderBase` class.  I'm not sure whether this change is OK.

Joeper214: yeah. That makes sense. both `lm_things_on_a_table` and `open_images_v4` is not related to `ImageFolderBase` AFAIK. Maybe it's better to extend to `dummy_classification` as well right? Since there is no `ImageFolder` dataset fixture.

tkng: I check the whole implementation of this file. Current implementation just checks some property of dataset (say, `num_per_epoch`, `extend_dir`, etc) but doesn't check that whether `__getitem__` method work, which is the most important feature for dataset class. It looks better to add some new unit tests, could you add a new github issue?

For this pull request, let's leave this as-is.

Joeper214: @tkng Issue created here:
https://github.com/blue-oil/blueoil/issues/509

tkng: Thanks, other than that, LGTM.

PR #463, opened 2019-09-24
Allocate proper array size

primenumber:
Minimize memory usage for buffer

## Motivation and Context

Current dlk converter is allocating 32x bigger size if buffer type is quantized type.
This PR fix it.

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/core/view.py
+++ b/dlk/python/dlk/core/view.py
@@ -30,7 +30,10 @@ class View(object):
 
     @property
     def shape(self):
-        return '*'.join(map(lambda x: str(x), self.op.shape))
+        if self.op.dtype == Float32():
+            return '*'.join(map(lambda x: str(x), self.op.shape))
+        else:
+            return '*'.join(map(lambda x: str(x), self.op.shape[:-1]))
 
     @property
     def shape_list(self):
--- a/dlk/python/dlk/core/view.py
+++ b/dlk/python/dlk/core/view.py
@@ -30,10 +30,10 @@ class View(object):
 
     @property
     def shape(self):
-        if self.op.dtype == Float32():
-            return '*'.join(map(lambda x: str(x), self.op.shape))
+        if self.op.dtype == QUANTIZED_PACKED():
+            return '*'.join(map(lambda x: str(x), self.op.shape)) + f' / (sizeof({self.op.dtype.cpptype()}) * CHAR_BIT)'
         else:
-            return '*'.join(map(lambda x: str(x), self.op.shape[:-1]))
+            return '*'.join(map(lambda x: str(x), self.op.shape))
 
     @property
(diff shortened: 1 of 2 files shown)


tkng:
Thanks for sending this pull request, this might be the most shocking pull request in this year...

(For recording): In verbal communication with @primenumber, I understand that the problem was that 32 (word size) was included at the end of shape, and it should not be included in the computation in the case of quantized convolution.

For example, in the case of LmnetV1 (dataset is CIFAR-10), the shape of `conv4_conv2d_Conv2D` is `[2, 16, 16, 2, 32]`. We don't need the last `32` in the case of quantized conv. Also, the shape of `conv3_QTZ_linear_mid_tread_half` is `[16, 16, 4, 2, 32]`.

Based on discussion, I re-read the code. For me, it looks that the real problem is that someone is inserting word size wrongly into the shape information. Hence, I investigated who do so, by inserting `traceback.print_stack` to `Operator.update_shape`.

The problem seems caused by [this line](https://github.com/blue-oil/blueoil/blob/master/dlk/python/dlk/core/optimizer.py#L496) and [this line](https://github.com/blue-oil/blueoil/blob/master/dlk/python/dlk/core/optimizer.py#L508). For me, it looks better to change these lines. What do you think?

primenumber:
Can we always assume word size = b parameter?
If so, we can remove last dimension data from shape list for quantized packed types.

tkng:
Hmm, I still misunderstood... As you pointed out, `b` is *not* word size, but parameter of our accelerator. Though it might be difficult to define what is word size, but IMO we should distinguish `b` and word size. For example, we can consider the case of `b = 16`, but word size should be 32 in my intuition.

tkng:
I think now I understand the problem completely.

Currently, data format of conv node is `ChHWBCl`. `dtype` of the conv node is `QUANTIZED_PACKED`, which is actually `QuantizedPacked<Uint32>`. Hence, we allocate `CHWB * 32 bit` for a conv node. On the other hand, how much we we should allocate is `CHWB bit`, thus current code is allocating 32 times larger buffer.

Accodring to my understanding, I think current `View.shape` method returns the number of required element to keep output information. Hense, what should be treated as a special case is `dype == QUANTIZED_PACKED`, not `dtype == Float32()`, because we might want to add `dtype == Int32()` in the future. Also, we should devide the return value by `sizeof(default_qword_dtype) * 8`.

I propose to change the pull request like following:

```
        if self.op.dtype == QUANTIZED_PACKED():
            return '*'.join(map(lambda x: str(x), self.op.shape) + '/ sizeof(default_qword_dtype) * 8')
        else:
            return '*'.join(map(lambda x: str(x), self.op.shape))
```

primenumber:
@tkng That's right. I fixed.

PR #472, opened 2019-09-30
Layout-generalized concat_on_depth

primenumber:
Support mixed layout (e.g. ChHWBCl and HWChBCl) concat_on_depth.

## Motivation and Context

There are several memory layout for Quantized and packed types.
And these layouts may be mixed in a network, even in inputs for an ops.
That cause compile error in concat_on_depth.
This PR fix it.
If you want to add new layout, you have to just implement index_{height,width,channels_high} and (access_ChHWBCl or access_NHWC).

## Description

- Re-implement func_ConcatOnDepth

## Types of changes

## Checklist:


dlk/python/dlk/templates/include/func/concat_on_depth.h
```diff
@@ -16,90 +16,153 @@ limitations under the License.
...
+    case MemoryLayout::ChHWBCl: return 2;
+    case MemoryLayout::HWChBCl: return 1;
+    case MemoryLayout::NHWC: return 2;
+    default: return 0;
+  }
+}
 
-  Measurement::Stop();
+template<class T>
+T& access_ChHWBCl(const TensorView<T, MemoryLayout::ChHWBCl>& tensor,
```
tkng: This function looks existing to access quantized tensor element, but the actual memory layout depends on the type  of the argument `tensor`. How about changing the function name to `access_quantized_tensor_elem` or something?

Just `access` or `access_tensor_element` might be enough.

dlk/python/dlk/templates/include/func/concat_on_depth.h
```diff
@@ -16,90 +16,153 @@ limitations under the License.
...
+    const std::size_t digit) {
+  return tensor(h, w, ch, digit, 0);
+}
 
-  T_UINT depths[32];
-  for (T_UINT n = 0; n < n_inputs; ++n) {
-    depths[n] = inputs[n].get_shape()[0];
-  }
+template<class T>
+T& access_NHWC(const TensorView<T, MemoryLayout::NHWC>& tensor,
```
tkng: I propose to change the function name for the same reason as above. Just `access `or `access_tensor_element might` be enough.

tkng:
The new code looks complex, it took a while for me to understand that the new code is calling the template class recursively to concat arbitrary number of tensors. However, I have no good counter proposal.

primenumber:
I agree that new code is complex.
If we introduce C++17, we can use `if constexpr` or `fold expression` to reduce complexity...

dlk/python/dlk/templates/include/func/concat_on_depth.h
primenumber: OK, fixed.

dlk/python/dlk/templates/include/func/concat_on_depth.h
primenumber: also fixed.

primenumber:
run lmnet test

tkng:
Now I'm fixing test fail problem, that will be merged soon. After that, you can re-run the test.

> If we introduce C++17, we can use if constexpr or fold expression to reduce complexity...

Considering that even out docker image is based on Ubuntu 16.04, we cannot use C++ 17 for now.
Maybe it's a good idea to leave a comment like "TODO: after compiler update to C++17, rewrite this function with if constexpr of fold expression."

PR #504, opened 2019-10-04
Fix lm_resnet skip connection for unusual dimension 

nlpng:
## Motivation and Context

Since @joelN123 keeps pointing out the bugs that would happen in certain condition, this PR fixes those conditions.

## Description

The conditions such as:
1. when the output channel and input channel size are not equal when no downsampling.
2. when difference of input and output channel is odd.

## How has this been tested?

manual test by changing the input and output channels

## Types of changes

## Checklist:


nlpng:
@iizukak 
I added the test for `lm_resnet` itself, I will try to include the test for the `basicblock`.

PR #506, opened 2019-10-04
Fix matrix multiplication for Arm

primenumber:
Fix matrix multiplication for Arm.

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/templates/src/matrix/multiplication.cpp
+++ b/dlk/python/dlk/templates/src/matrix/multiplication.cpp
@@ -78,9 +78,9 @@ void matrix_multiplication_impl(
   const auto B_col_blocks = (B.cols() + regblock_m - 1) / regblock_m;
   float *B_buf_ptr = B_buf.get();
   for (std::size_t j = 0; j < B.cols(); j += regblock_m) {
-    if (B.cols() - j >= regblock_m) {
+    if (j + regblock_m <= B.cols()) {
       std::size_t k = 0;
-      for (; k + regblock_m - 1 < B.rows(); k += regblock_m) {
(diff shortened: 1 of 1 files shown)


dlk/python/dlk/templates/src/matrix/multiplication.cpp
```diff
@@ -80,7 +80,7 @@ void matrix_multiplication_impl(
   for (std::size_t j = 0; j < B.cols(); j += regblock_m) {
     if (B.cols() - j >= regblock_m) {
```
tkng: This might be out of scope of this PR, is there a reason to write the condition in this way?

For me, `j + regblock_m < B.cols()` looks natural and consistent to line 80.

dlk/python/dlk/templates/src/matrix/multiplication.cpp
```diff
@@ -80,7 +80,7 @@ void matrix_multiplication_impl(
...
       std::size_t k = 0;
-      for (; k < B.rows(); k += regblock_m) {
+      for (; k + regblock_m - 1 < B.rows(); k += regblock_m) {
```
tkng: How about changing this condition to `k + regblock_m <= B.rows()`? It looks a little bit simple. (Though generated assembler code might be the same...)

dlk/python/dlk/templates/src/matrix/multiplication.cpp
primenumber: Fixed.

dlk/python/dlk/templates/src/matrix/multiplication.cpp
primenumber: Fixed.

PR #490, opened 2019-10-03
Fix DLK code generation test cases

hadusam:
Step 1 of #477 

I changed test cases to inherit base-file definition, use inputs of table testing and separated test files of each architecture for running each test separately.

| Task | Model | x86_64 | ARM | ARM_FPGA |
| --- | --- | --- | --- | --- |
| classification | lmnet_quantize_cifar10_space_to_depth | o | o | o |
| classification | resnet_quantize_cifar10 | o | o | o |
| object_detection | fyolo_quantize_4_v4 | o | o | o |
| segmentation | lm_segnet_v1_quantize_camvid | o | o | o |

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature / Optimization (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #477, opened 2019-10-02
DLK test cases are complicated


hadusam:
I would like to add new tests for AArch64 (related #335 )
But current DLK tests are very complicated.

First of all, we have many test cases of [test_code_generation.py](https://github.com/blue-oil/blueoil/blob/c3689d4e63b88d12c9a27291ee7a21cbe204bedf/dlk/tests/test_code_generation.py)
These tests create a binary file and run a script with binary on each x86_64, ARM and ARM_FPGA environment.
It is too difficult to understand what test cases are run, and too hard to add new devices for similar test cases.

Currently, these test cases are run.

| Task | Model | x86_64 | ARM | ARM_FPGA |
| --- | --- | --- | --- | --- |
| classification | lmnet_v1_group_conv | o | - | - |
| classification | lmnet_quantize_cifar10 | o | - | - |
| classification | lmnet_quantize_cifar10_space_to_depth | o | o | o |
| classification | resnet_quantize_cifar10 | o | o | o |
| object_detection | fyolo_quantize_4_v4 | o | o | o |
| object_detection | widerface_320 | - | o | - |
| object_detection | widerface_v5 | - | o | - |
| segmentation | lm_segnet_v1_quantize_camvid | o | - | - |

Wow, segmentation task is not tested on ARM, FPGA 😭 

These Models are found on [dlk/examples](https://github.com/blue-oil/blueoil/tree/master/dlk/examples) directory, and these `.pb` files are too old, so we should update and replace them!

I am planning to make the latest `.pb` files of current stable networks supported by blueoil (ref: [blueoil/blueoil_init.py](https://github.com/blue-oil/blueoil/blob/master/blueoil/blueoil_init.py#L33-L56)).
So the test cases will be as follows.
 
| Task | Model | x86_64 | ARM | ARM_FPGA |
| --- | --- | --- | --- | --- |
| classification | LmnetV1Quantize | o | o | o |
| classification | ResNetQuantize | o | o | o |
| object_detection | LMFYoloQuantize | o | o | o |
| segmentation | LmSegnetV1Quantize | o | o | o |

iizukak:
Thanks @hadusam. I agreed your suggested plan.

hadusam:
To make small changes and easily reviewing, I would like to solve this issue following some steps.
1. Refactor and align the same test cases on each device with current Models.

| Task | Model | x86_64 | ARM | ARM_FPGA |
| --- | --- | --- | --- | --- |
[...]

PR #518, opened 2019-10-15
Refactoring dlk/tests/test_code_generation.py

hadusam:
Related Step1 of #477
Separated PR of #490 for reviewability

## Types of changes

## Checklist:


hadusam:
@iizukak Thanks for reviewing! 😄

ruimashita:
This changes are getting wrong for readability, I think.
When we read test code, we focus what test do. The many nested test case generation function is difficult to understand what test do.
Old version is simpler and better for understand.

hadusam:
@ruimashita 
Thanks for your comment!
I don't think the old version is better for understanding because it was like a patchwork of test cases and not structured.
But I agree with your opinion, we should make it explicit what test cases run in test code and we should not use dynamical function for test cases.
I created another PR to improve test case readability 👍 
Please check it!

PR #520, opened 2019-10-15
Add fpga_setup flag for setUpClass

hadusam:
Related Step1 of #477
Separated PR of #490 for reviewability

## Motivation and Context

Currently, in the test case of x86_64, FPGA will be set up. However, that is unnecessary.
Add fpga_setup flag for setUpClass to avoid unnecessary setup.
## Types of changes

## Checklist:

### DIFF
--- a/dlk/tests/test_code_generation.py
+++ b/dlk/tests/test_code_generation.py
@@ -392,6 +392,8 @@ class TestCodeGenerationX8664(TestCodeGenerationBase):
 class TestCodeGenerationArm(TestCodeGenerationBase):
     """Test class for code generation testing."""
 
+    fpga_setup = True
+
     @params(*get_configurations_arm())
     def test_code_generation(self, i, configuration) -> None:
         self.run_test_all_configuration(i, configuration)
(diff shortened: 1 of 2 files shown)


dlk/tests/test_code_generation.py
```diff
@@ -389,15 +389,15 @@ def test_code_generation(self, i, configuration) -> None:
         self.run_test_all_configuration(i, configuration)
 
 
-class TestCodeGenerationArm(TestCodeGenerationBase):
+class TestCodeGenerationArm(TestCodeGenerationBase, TestCaseFPGABase):
```
tsawada: I haven't checked these base classes, but it is better to avoid multiple inheritance because it adds complexity. Is there a chance we could merge them?

hadusam: @tsawada Thanks for reviewing!
OK, I see. I changed them to use a class variable `setup_fpga` as a flag for checking whether the setup is needed instead of multiple inheritances with splitting class.
Please review it again! 🙇

Issue #523, opened 2019-10-15
If set optional classes on MscocoObjectDetection then fail to instantiate.


takecore:
at version: `v0.12.0`

I want to use optional classes on MscocoObjectDetection like below code.

https://github.com/blue-oil/blueoil/blob/master/lmnet/lmnet/datasets/mscoco.py#L140

```
class MscocoObjectDetection(ObjectDetectionBase):
-     classes = DEFAULT_CLASSES
---
+     classes = ["person"]
```

So I checked it with test like below code.

```py
# lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
from lmnet.datasets.mscoco import MscocoObjectDetection, DEFAULT_CLASSES


def test_mscoco_object_detection_optional_classes():
    classes = [DEFAULT_CLASSES[0]]
    MscocoObjectDetection.classes = classes
    dataset = MscocoObjectDetection(subset='train')
```

And got error like below.

```
$ pytest tests/lmnet_tests/datasets_tests/test_mscoco.py

________________________________________________________________ test_mscoco_object_detection_optional_classes ________________________________________________________________

    def test_mscoco_object_detection_optional_classes():
        classes = [DEFAULT_CLASSES[0]]
        MscocoObjectDetection.classes = classes
>       dataset = MscocoObjectDetection(subset='train')

tests/lmnet_tests/datasets_tests/test_mscoco.py:25:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
lmnet/datasets/mscoco.py:164: in __init__
    self._init_files_and_annotations()
lmnet/datasets/mscoco.py:249: in _init_files_and_annotations
    self.files, self.annotations = self._files_and_annotations()
lmnet/datasets/mscoco.py:244: in _files_and_annotations
    gt_boxes_list = [self._gt_boxes_from_image_id(image_id) for image_id in image_ids]
lmnet/datasets/mscoco.py:244: in <listcomp>
    gt_boxes_list = [self._gt_boxes_from_image_id(image_id) for image_id in image_ids]
[...]

PR #525, opened 2019-10-15
MscocoObjectDetection can set optional classes and instantiate.

takecore:
## Motivation and Context

See #523

## Description

in `_gt_boxes_from_image_id`, only to use `self.classes` related annotation data.

I add `lmnet/test/fixtures/datasets/MSCOCO`. please check this out.
I add blank 4 images to MSCOCO/train2014, MSCOCO/val2014.
like this image 10x10 black image.

COCO_train2014_000000000001

## How has this been tested?

```sh
# on blueoil/lmnet
$ PYTHONPATH=. pytest tests/lmnet_tests/datasets_tests/test_mscoco.py
```

## Types of changes

## Checklist:


lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
```diff
@@ -0,0 +1,31 @@
...
+
+pytestmark = pytest.mark.usefixtures("set_test_environment")
+
+
+def test_mscoco_object_detection_optional_classes_instantiate():
+    classes = [DEFAULT_CLASSES[0]]
+    MscocoObjectDetection.classes = classes
+    try:
+        MscocoObjectDetection()
+        assert True
```
iizukak: @takecore 
Thank you for your contribution.
Is it possible to add test for `MscocoObjectDetection` instance?

takecore: @iizukak 
This test only check for instantiation because this is the PR for #523. 

If there are any problems, I would like to solve it in another PR!! 🦀

takecore:
I fixed buildkite test. please check again if you have time.

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
```diff
@@ -0,0 +1,31 @@
...
+    except Exception:
+        assert False
```
Joeper214: I think it's better to assert more clear expectations in this test case like setting the number of classes the `MscocoObjectDetection()` method returns. not just by setting a try-catch that causes unmet expectations.

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
```diff
@@ -0,0 +1,31 @@
...
+import pytest
+
+from lmnet.datasets.mscoco import MscocoObjectDetection, DEFAULT_CLASSES
+
+
+pytestmark = pytest.mark.usefixtures("set_test_environment")
+
+
+def test_mscoco_object_detection_optional_classes_instantiate():
+    classes = [DEFAULT_CLASSES[0]]
```
Joeper214: Please specify what class name you want to get from the default classes.

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
Joeper214: Also, I can't recommend using try-catch in unit tests.

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
takecore: @Joeper214 
I think better to specify `classes =  DEFAULT_CLASSES[0]` than `classes =  ['person']`
How do you think?

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
iizukak: I see

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
Joeper214: Yes. 👍

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
Joeper214: Maybe you can refer to some test cases here:

https://github.com/blue-oil/blueoil/blob/master/lmnet/tests/lmnet_tests/datasets_tests/test_cityscapes.py

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
takecore: I fixed it.

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
takecore: @Joeper214 I add another test. please check this out 🐶

lmnet/tests/fixtures/datasets/MSCOCO/annotations/instances_train2014.json
```diff
@@ -0,0 +1,534 @@
+{
+  "info": {
```
tsawada: these needs to be replaced with dummy as well

lmnet/tests/fixtures/datasets/MSCOCO/annotations/instances_val2014.json
```diff
@@ -0,0 +1,534 @@
+{
```
tsawada: same

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
```diff
@@ -0,0 +1,59 @@
...
+import pytest
+
+from lmnet.datasets.mscoco import MscocoObjectDetection
+
+
+pytestmark = pytest.mark.usefixtures("set_test_environment")
+
+
+# TODO
+def test_mscoco_segmentation():
```
tsawada: This gives a false sense that we test segmentation (when somebody looks at list of tests)
So just have a todo comment without a function.
`# TODO(takecore): test segmentation`

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
```diff
@@ -0,0 +1,59 @@
...
+def test_mscoco_object_detection(subset, num_classes, files, annotations):
+    dataset = MscocoObjectDetection(subset=subset)
+
+    assert dataset.num_classes == num_classes
+    assert (files, annotations) == dataset._files_and_annotations()
+
+
+def test_mscoco_object_detection_optional_classes_instantiate():
+    MscocoObjectDetection.classes = ['person']
+    try:
```
tsawada: Why a `try` block? I think the test fails when the function throws exception anyway (I might be wrong)

lmnet/lmnet/datasets/mscoco.py
```diff
@@ -225,10 +225,16 @@ def coco_category_id_to_lmnet_class_id(self, cat_id):
     @functools.lru_cache(maxsize=None)
     def _gt_boxes_from_image_id(self, image_id):
         """Return gt boxes list ([[x, y, w, h, class_id]]) of a image."""
+        classes = [class_name for class_name in self.classes if class_name is not "__background__"]
+        class_ids = self.coco.getCatIds(catNms=classes)
```
tsawada: how many ids do we have in `class_ids` ?
if this is only used in line 235, we should convert this to tuple.

takecore: `class_ids` has `90`.

tuple OK, thanks.

lmnet/tests/lmnet_tests/datasets_tests/test_mscoco.py
takecore: @tsawada 

This try block mean instantiate test.

> I think the test fails when the function throws exception anyway

I think so. but test fails problem is how class designed is. (I think it is out of scope.)
This PR just fix override `classes` and can instantiate.

```py
# ok pattern
MscocoObjectDetection()

# error occured pattern, but this PR will can run
MscocoObjectDetection.classes = ['person']
MscocoObjectDetection()
```

lmnet/lmnet/datasets/mscoco.py
tsawada: Oops I was wrong, we should use `set()` not `tuple()` for `class_ids`. `classes` can be a list.

lmnet/tests/fixtures/datasets/MSCOCO/annotations/instances_val2014.json
```diff
@@ -0,0 +1,534 @@
...
+      "id": 89,
+      "name": "hair drier"
+    },
+    {
+      "supercategory": "indoor",
+      "id": 90,
+      "name": "toothbrush"
+    }
+  ]
+}
```
tsawada: always pay attention to this "🚫⏎" mark on every file. We need a newline character at the end of (almost) any text files. Same for the other json file.

takecore:
@tsawada I fixed it! please check again.

lmnet/lmnet/datasets/mscoco.py
```diff
@@ -225,10 +225,16 @@ def coco_category_id_to_lmnet_class_id(self, cat_id):
     @functools.lru_cache(maxsize=None)
     def _gt_boxes_from_image_id(self, image_id):
         """Return gt boxes list ([[x, y, w, h, class_id]]) of a image."""
+        classes = set(class_name for class_name in self.classes if class_name is not "__background__")
```
tsawada: `class_ids` should be a set BUT`classes` should be a list.

takecore: sorry, you right. I fixed this.

lmnet/lmnet/datasets/mscoco.py
```diff
@@ -198,7 +198,7 @@ def coco(self):
     def _image_ids(self):
         """Return all files and gt_boxes list."""
         classes = [class_name for class_name in self.classes if class_name is not "__background__"]
-        target_class_ids = self.coco.getCatIds(catNms=classes)
+        target_class_ids = set(self.coco.getCatIds(catNms=classes))
```
tsawada: why do you use `set()` here?

takecore: @tsawada I have 2 reasons.

When another person looks at the code in the future, it can be recognized that the same description is written.
If they can be recognized as the same, it is easy to fixed or abstracted.
That's why I used `set`.

It's hard to tell if you're using `self.classes` in this class or if you're using looped `classes`.
I want to distinguish each one.
But, since the scope of the fixes is getting wider and wider, I have only described `set`.

tsawada: I think we don't need set() here. Please leave it as a list.

PR #539, opened 2019-10-18
Remove unused files

primenumber:
im2col quantized convolution is no longer used.

## Description

Remove 2 files about im2col conv.

## Types of changes

## Checklist:


tkng:
This will be a great pull request that can remove amost 5% lines of C++ code from out repository, however, it would need a little more work to pass the unit tests.

- the line `#include "func/impl/quantized_conv2d_dim2col.h"` should be removed from `tensor_convert.h`
- accordingly, some `convert_tensor` functions should be removed from `tensor_convert.h`

PR #540, opened 2019-10-18
Raise TypeError in view.run() if op_type is not supported

yd8534976:
Raise TypeError in view.run() if op_type is not supported.

When we add a new op into dlk, we have to add related codes in `view.py` to generate code.
This PR will inform us if we miss this piece of code.

## Types of changes

## Checklist:


dlk/python/dlk/core/view.py
```diff
@@ -744,6 +744,8 @@ def run(self):
 
             return self.format_string(f"""func_Lookup({inputs_string}, {op.name});""")
 
+        raise TypeError(f"{self.op.op_type} is not supported in view.py!")
```
tkng: This error message is not telling a lie, but "not supported in View.run" whould be more kind.

Also, the exclamation mark emphasis should not be used casually.

yd8534976: I have fixed it.

PR #544, opened 2019-10-21
make global_step available from all networks

tkng:
# Motivation and Context

Currently, `global_step` variable which describes how many training steps were executed, is only available in object detection networks.

However, sometimes I want to use this variable in other kind of networks like classification networks.

## Description

This pull request changes how global_step variable is passed to network object.

- Current: passed by argument of `loss()` method and `train()` method
- This PR: passed by argument of network object constructor

## How has this been tested?

unit test

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/executor/train.py
+++ b/lmnet/executor/train.py
@@ -89,35 +89,35 @@ def start_training(config):
 
     graph = tf.Graph()
     with graph.as_default():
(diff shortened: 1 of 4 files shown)


tkng:
Since `global_step` is only needed for training, I realized that passing the variable for `train` method is better design. Hence I close this issue, will make another pull request.

PR #386, opened 2019-08-14
Add Single-person pose estimation task

yd8534976:
This PR provides a full pipeline for single-person pose estimation.

Current camera demo is not very stable, I will make other PRs (Improve base model or more data augmentations) in the near future.

## Description

It contains:
- MSCOCO_2017 dataset loader customized for single-person pose estimation.
- Pre process and post process.
- An example quantized network.
- Unit tests.
- Training visualization and metrics for tensorboard.
- Usb camera demo related code.

## How has this been tested?

@Joeper214 - san has run a usb_camera_demo successfully on a de10_nano. Thanks!

### How to test:

- Environment variables in below example:
    - `OUTPUT_DIR=/storage2/home/yang/lmnet/saved`
    - `DATA_DIR=/storage2/dataset/MSCOCO_2017`

- Train a quantized model (under `blueoil/lmnet`):
`docker-compose run --rm tensorflow python executor/train.py -c configs/core/keypoints_detection/lm_single_pose_v1_quantize_mscoco.py -i single_pose_example`

- Export model (under `blueoil/lmnet`):
`PYTHONPATH=. python executor/export.py -i single_pose_example`

- Generate code (under `blueoil/dlk`):
`PYTHONPATH=python/dlk python python/dlk/scripts/generate_project.py -i /storage2/home/yang/lmnet/saved/single_pose_example/export/save.ckpt-XXX/256x320/minimal_graph_with_shape.pb -o ~/tmp -p single_pose_example -hq -ts`

- Make (under `~/tmp/single_pose_example`):
`make -j8 lib_fpga`

- Copy related files to de10-nano:
    - `~/tmp/lib_fpga.so`
    - `blueoil/output_template`
    - `/storage2/home/yang/lmnet/saved/single_pose_example/meta.yaml`

- Run demo:
`ssh -X root@de10-nano`
`cd output_template/python`
`PYTHONPATH=. python usb_camera_demo.py -m ~/lib_fpga.so -c ~/meta.yaml`

## Screenshots (if appropriate):

image

image

image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature / Optimization (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #546, opened 2019-10-21
Add FormatJoints post_processor for single-person Pose estimation

yd8534976:
This PR is a part of #386 , post processor part.

- FormatJoints class.
- Unit test (to test the consistency with pre processor).

## Types of changes

## Checklist:


lmnet/lmnet/post_processor.py
```diff
@@ -419,3 +419,74 @@ class Softmax(Processor):
...
+        Args:
+            num_dimensions: int, it only supports 2 for now.
+            stride: int, stride = image_height / heatmap_height.
+            confidence_threshold: float, [0, 1], ratio of threshold_value and max_value.
+        """
+        self.num_dimensions = num_dimensions
+        self.stride = stride
+        self.confidence_threshold = confidence_threshold
+
+    def __call__(self, outputs, *args, **kwargs):
```
tkng: I feel uncomfortable using "outputs" as a variable name for the input argument. Do you have another idea for this variable name?

lmnet/lmnet/post_processor.py
```diff
@@ -419,3 +419,74 @@ class Softmax(Processor):
...
+                                                   num_dimensions=self.num_dimensions,
+                                                   stride=self.stride,
+                                                   confidence_threshold=self.confidence_threshold)
+
+        return dict({'outputs': joints}, **kwargs)
+
+
+def gaussian_heatmap_to_joints(heatmap, num_dimensions=2, stride=2, confidence_threshold=0.1):
+    """
+    Extract joints from gaussian heatmap. Current version only supports 2D pose estimation.
```
tkng: Since `FormatJoints` class is a public interface (but `gaussian_heatmap_to_joints` is not, I guess), it would be better to move this kind of comment to the class description.

lmnet/lmnet/post_processor.py
```diff
@@ -419,3 +419,74 @@ class Softmax(Processor):
...
+        confidence_threshold: float, [0, 1], ratio of threshold_value and max_value.
+
+    Returns:
+        joints: a numpy array of shape (num_joints, num_dimensions + 1).
+
+    """
+
+    height, width, num_joints = heatmap.shape
+
+    threshold_value = 10 * confidence_threshold
```
tkng: This `10` looks a little sudden. How about using 10 times larger confidence threshold?

lmnet/lmnet/post_processor.py
```diff
@@ -419,3 +419,74 @@ class Softmax(Processor):
     def __call__(self, outputs, **kwargs):
         results = _softmax(outputs)
         return dict({'outputs': results}, **kwargs)
+
+
+class FormatJoints(Processor):
```
tkng: How about changing this class name to` GaussianHeatmapToJoints` or something like that?

I'm a little bit unconfortable with this class name, since it starts with "Format" (verb) but it's not consistent with other class name, say, `FormatYoloV2`. With the latter name, it is very obvious for me, I can easily understand that this post processor will format the output of YoloV2 network. On the other hand, what the `FormatJoints` means is not obvious for me. Actually, at first, I misunderstood that this class expects Joint information as input, and output something which could be computed by the joints.

lmnet/lmnet/post_processor.py
yd8534976: I agree with you. This name is not comfortable.
But since this line of demo code:  https://github.com/blue-oil/blueoil/blob/4928b8ed05023b3fafe1f874c93d2e587c637a6b/output_template/python/lmnet/utils/demo.py#L159
I can not decide this variable name without refactoring.
Current implementation makes we can combine this post processors using `Sequence` in any order. But it does sacrifice the readability. In fact, we may don't need this feature in post process cases..

lmnet/lmnet/post_processor.py
yd8534976: It is sudden... I got `10` just from tuning... I set it to 10 by default in pre_processor.
https://github.com/blue-oil/blueoil/blob/4928b8ed05023b3fafe1f874c93d2e587c637a6b/lmnet/lmnet/pre_processor.py#L451
Two solutions here, I can hard-code both or add a `max_value` argument here as pre_processor.

Which one do you think is better? 🤔

lmnet/lmnet/post_processor.py
yd8534976: This name was trying to imitate `FormatYoloV2`.
`GaussianHeatmapToJoints` should be better here as you said. 👍

lmnet/lmnet/post_processor.py
tkng: Thanks, I understood better than before. I recommend to omit this `10` (and `max_value` also.) A user need to optmize threshold value (if needed) anyhow, hence adding a hyperparameter that only affects the threshold value linearly only increases the effort for users.

Another option is that you can extract the max_value automatically from a heat map, but personally I don't prefer to add such a feature which might not be trivial from user's perspective.

lmnet/lmnet/post_processor.py
tkng: Though I'm not sure, but it looks OK to just remove `outputs=` from the line 159 of `demo.py`. Could you try?

yd8534976: It seems ok to remove it. How about changing them in another PR? 
Since all arg of post processors is 'outputs' now.

tkng: OK, let's leave them as is now.

lmnet/lmnet/post_processor.py
yd8534976: Thanks. Fixed.

lmnet/lmnet/post_processor.py
```diff
@@ -419,3 +419,81 @@ class Softmax(Processor):
...
+    # max value of gaussian heatmap.
+    max_value = 10
+    threshold_value = max_value * confidence_threshold
+
+    joints = np.zeros((num_joints, num_dimensions + 1), dtype=np.float32)
+
+    for i in range(num_joints):
+        argm = np.argmax(heatmap[:, :, i])
+        y, x = np.unravel_index(argm, (height, width))
+        max_value = heatmap[y, x, i]
```
tkng: this `max_value` seems not maximum value of a heatmap.

tkng: Sorry, that's my misunderstanding, please ignore.

lmnet/lmnet/post_processor.py
tkng: There seems a miscommunication... 😢 I meant all `max_value` (except the line 492&493) should be rtemoved. I'll explain why I'm thinking so.

- From the docstring, it would be hard to understand what will happen if a user change the `confindence_threshold`, since none of `threshold_value` and `max_value` is explained.
- `max_value` in the line 484 doesn't help a user. The result is exactly the same as multiplying the `confidence_threshold` by 10. User will be confused because they cannot know that actual threshold value is multiplied by 10, until they read the code.

Please keep in mind, we should write code that the user can guess the behavior by looking at the API. I'm thinking that this point is very important for your growth as a software engineer. If you feel that your understanding is not enough, please ask me a question offline at any time 😉

yd8534976: It was confusing that I used `max_value` twice.. I have fixed it. Thanks! 😄

tkng: It's not a problem of using `max_value` twice. Could you explain why do you want multiply `10` to heatmap?

yd8534976: Since I found that quantized networks are hard to regress a heatmap directly ranged from 0 to 1. So I multiply it by `10` and it works. `10` is a tuned result.

tkng: Then, `max_value` would not be an appropriate variable name. How about `heatmap_scaling_coefficient` or `heatmap_scaling_factor`?

yd8534976: Yes! it is. I'm sorry for this bad name. I was confused to choose this name! `heatmap_scaling_factor` should be better I think. Do I need to add a redundant variable or just change the comments.

tkng:
I'm still not comfortable with multiplying 10, because there seems to be another better way... However, I don't have clear good answer so far, I accept this pull request.

yd8534976:
@tkng , Thanks!
I can understand your feeling. Me too. In fact, I have tried many values from `1` to `255`. `10` performed best so far. Maybe I could dismiss this `10`. after tuning hyper parameters more.
There are some implementations about this `scaling_factor`.

They use `1` in fact:
https://github.com/dmlc/gluon-cv/blob/3e2a26e81f08019b71ef87fcde2b5b989fd3bc53/gluoncv/data/transforms/pose.py#L468

They use `255` in fact:
https://github.com/chenyilun95/tf-cpn/blob/1ed29435262fa8239672050d87799467b5e8db26/models/COCO.res101.384x288.CPN/dataset.py#L156

PR #548, opened 2019-10-23
Add global_step for all loss function tasks

### LINKED ISSUE #458: add `global_step` variable as an argument for all loss function
Currently, only object detection model receives the variable `global_step` as the third argument. Consequently, in `lmnet/executor/train.py`, we need an if statement to pass `global_step` only for object detection model.

IMO, conditional branching by task type should be avoided whenever possible. To this end, we can pass `global_step` for all network definitions.

Joeper214:
## Motivation and Context

- fix #458 

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/executor/train.py
+++ b/lmnet/executor/train.py
@@ -109,10 +109,8 @@ def start_training(config):
         images_placeholder, labels_placeholder = model.placeholders()
 
         output = model.inference(images_placeholder, is_training_placeholder)
-        if config.TASK == Tasks.OBJECT_DETECTION:
-            loss = model.loss(output, labels_placeholder, global_step)
-        else:
-            loss = model.loss(output, labels_placeholder)
+        loss = model.loss(output, labels_placeholder, global_step)
+
         opt = model.optimizer(global_step)
         if use_horovod:
             # add Horovod Distributed Optimizer


tkng:
I made a similar pull request a few days ago, but closed the pull request. (Please check #544.)

Now I'm thinking that there is room for improvement in current network API design, it's not just a problem of `global_step` variable. The purpose of `train.py` is just training. What the script need to know about the network should be following:

- how to initialize the network
- how to execute a training step
- how to execute a validation step

`train.py` need not to know how to run inference, how to compute a loss. Rather, it is better not to know.

Could you tell us your opnion about the idea above?

Joeper214:
> I made a similar pull request a few days ago, but closed the pull request. (Please check #544.)

Oh. Sorry I didn't notice that. It seems like this PR makes no sense at all since I was only thinking of a temporary fix. I'll be closing this PR in that case.

### Regarding 
* how to initialize the network
* how to execute a training step
* how to execute a validation step

I agree as it will make train.py simpler and straightforward. 

However, I'm not sure how we can generate the gradients, summary, matrices and apply the loss function each step, and generate checkpoints in that case. Will it be in a separate API or executor?

PR #561, opened 2019-10-25
Move requirements.txt in lmnet to the top

iizukak:
## Motivation and Context

To create blueoil python package in the furture, We want to move requirements.txt to the top directory.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.


docker/Dockerfile
```diff
@@ -91,8 +91,8 @@ RUN apt-get update && apt-get install -y g++-5-arm-linux-gnueabihf && \
     ln -s /usr/bin/arm-linux-gnueabihf-g++-5 /usr/bin/arm-linux-gnueabihf-g++
 
 # Install requirements
-COPY lmnet/*requirements.txt /tmp/requirements/
-RUN pip install -r /tmp/requirements/gpu.requirements.txt
+COPY ./*requirements.txt /home/blueoil/
```
suttang: I think we don't need to copy requirement to `/home/blueoil`.
requirements.txt is not required for production use.

hadusam: +1

iizukak: @suttang 
We need `requirements.txt` in Docker image because we need to install requirements.
Which path is good ?

suttang: @iizukak 
/tmp/requirements/ would be better.

iizukak:
@suttang Oh, I simply misunderstood. Path was fixed.

lmnet/tox.ini
```diff
@@ -7,7 +7,7 @@ skipsdist = True
 [testenv]
 recreate = True
 deps =
-    -r{toxinidir}/gpu.requirements.txt
+    -r{toxinidir}/../gpu.requirements.txt
```
iizukak: @suttang @hadusam 
In docker image, we have requreiments.txt in `/tmp/requirements/`, When tox executed, these files cannot find requirements.txt. Do you have any idea to find path in tox using docker?

suttang: IMO tox command via docker will not be used much.
If you want to run via docker, you just use -v.

iizukak: >  tox command via docker will not be used much.

`make test-lmnet` use tox. So this line cause CI failure.
Thinking time...

iizukak:
Talking with @suttang and @hadusam, We decide to put `*requirements.txt` in `/home/blueoil` for tox issue.

PR #565, opened 2019-10-25
Refactor dlk/tests/test_code_generation.py to improve test case readabiity

hadusam:
In #518 , @ruimashita commented as following,
> When we read test code, we focus what test do. The many nested test case generation function is difficult to understand what test do.

I agree with this opinion, but previous test cases were very complicated and difficult to understand because it was many duplicated test case settings and relied on implicit inherit default setting.
So I refactor again to write the explicit test cases to avoid implicitly generated test cases by function by using shared base configurations.

## Types of changes

## Checklist:

### DIFF
--- a/dlk/tests/test_code_generation.py
+++ b/dlk/tests/test_code_generation.py
@@ -90,83 +90,145 @@ def dict_codegen_segmentation(cpu_name) -> dict:
             }
 
 
-def get_configurations_by_test_cases(test_cases, configuration):
-
-    return [updated_dict(configuration,test_case) for test_case in test_cases]
-
-
-def get_configurations_by_architecture(test_cases, cpu_name):
-    configurations = []
-    configurations.extend(get_configurations_by_test_cases(test_cases, dict_codegen_classification(cpu_name)))
-    configurations.extend(get_configurations_by_test_cases(test_cases, dict_codegen_classification_resnet(cpu_name)))
-    configurations.extend(get_configurations_by_test_cases(test_cases, dict_codegen_object_detection(cpu_name)))
-    configurations.extend(get_configurations_by_test_cases(test_cases, dict_codegen_segmentation(cpu_name)))
-
-    return configurations
-
-
 def get_configurations_x86_64():
     cpu_name = "x86_64"
-    test_cases = [
-        {'use_avx': True, 'hard_quantize': True, 'threshold_skipping': True},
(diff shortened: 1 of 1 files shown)


ruimashita:
@hadusam Nice :+1:

ruimashita:
[x] owner approval

dlk/tests/test_code_generation.py
```diff
@@ -90,83 +90,145 @@ def dict_codegen_segmentation(cpu_name) -> dict:
...
-    additional_test_cases = [
-        {'use_avx': True, 'input_name': 'raw_image.png', 'test_level': TEST_LEVEL_FUTURE_TARGET},
-        {'use_avx': True, 'input_name': 'preprocessed_image.npy', 'from_npy': True},
-        {'use_avx': False, 'input_name': 'raw_image.png', 'test_level': TEST_LEVEL_FUTURE_TARGET},
-        {'use_avx': False, 'input_name': 'preprocessed_image.npy', 'from_npy': True},
+    cls_base = dict_codegen_classification(cpu_name)
+    cls_resnet_base = dict_codegen_classification_resnet(cpu_name)
+    det_base = dict_codegen_object_detection(cpu_name)
+    seg_base = dict_codegen_segmentation(cpu_name)
+    configurations = [
```
tsawada: This look overly complex.
What about
```
sub_patterns = [{'use_avx': True, ...}, ...]
configurations = []
configurations.extend(get_configurations_by_test_cases(cls_base, sub_patterns)
...

return list(enumerate(configurations))
```

Well, this also look complex :(
do you have better idea?

hadusam:
Umm...
Now, there are three patterns as follows,
1. Return test cases by architecture and testcase_pattern(Nested Function): currently the master branch
2. Return test cases by base_config and testcase_pattern: @tsawada commented
3. No Function for test cases: this PR

|No|Easy to add test as the same setting as other cases|Easy to understand what test will do|Constraints|
| - | - | - | - |
|1|○|×| Test should be the same by each architecture and by each network |
|2|○|△| Test should be the same by each network |
|3|×|○| Nothing |

Why I created this PR, I am following the thinking that test cases should be explicit as much as possible.
On the other hand, test cases will increase by multiplication of parameters, so this will become not to be able to show all test cases because of too many lines in the near future.
I think there is no best answer to this issue, but we should decide which is better.
What do you think? @tsawada @ruimashita 
I should get more opinions from other blueoil members?

PR #566, opened 2019-10-25
Can access tensorboard using docker-compose

ruimashita:
## Motivation and Context

When I use docker-compose run tensorbaord, I can't access tensorboard.
I change the network config of docker to use host network. 

## How has this been tested?

```
$ DATA_DIR=/storage/dataset/ docker-compose run blueoil tensorboard --logdir saved/
TensorBoard 1.14.0 at http://leapmind-dgx-01:6006/ (Press CTRL+C to quit)
```
then I can access http://leapmind-dgx-01:6006/

## Types of changes

## Checklist:


kchygoe:
I'm curious this change should be default or not. coz the change is based on the idea that access tensorboard from outside of host but I suppose it is optional. 
Let me check if it can be optional as docker-compose option

ruimashita:
@kchygoe Thanks :+1: 
This `docker-compose` is only for development. On development, we run tensorbaord and see tensorbaord from outside always. That's why I think this change should be default .

suttang:
I can access tensorboard using the below command.

```
DATA_DIR=/storage/dataset docker-compose run -p 6007:6006 blueoil tensorboard --logdir=saved
```

Is the purpose of this PR to use the default port binding?

Issue #545, opened 2019-10-21
Fix comment of __len__ in dataset class


suttang:
Currently, `dataset.num_per_epoch` and `dataset.__len__` should return same value.
But both of comments has different meanings.

```python
    def num_per_epoch(self):
        """Returns the number of datas in the data subset."""
```

```python
    def __len__(self):
        """returns the number of items in the dataset."""
````

The comment of `__len__` looks like # of files in all of the subsets.

And we should print a message 
And we need to inform that `num_per_epoch` will be deprecated.

PR #581, opened 2019-10-30
added deprecation warning ISSUE #545

twille00:
I added the deprecation warning and changed the comment in __len__ function.


lmnet/lmnet/datasets/base.py
```diff
@@ -95,7 +95,7 @@ def __getitem__(self, i, type=None):
     @property
     @abstractmethod
     def __len__(self):
-        """returns the number of items in the dataset."""
+        """returns the number of datas in the dataset."""
```
tkng: Original explanation is correct, could you revert this line?

BTW, `data` is already plural form, hence we cannot say `datas`. (This means that line 86 is wrong... 😢)

twille00: I revert this, done. :)

Issue #569, opened 2019-10-28
Wrong memoryLayout in func_QTZ_binary_channel_wise_mean_scaling


yd8534976:
NHWC is not a layout for kernel. It should be something like OHWC or HWCO.

But It seems `func_QTZ_binary_channel_wise_mean_scaling` may never be generated..

https://github.com/blue-oil/blueoil/blob/56f6e4451e1dc48b3206aab28dc7444d373b5994/dlk/python/dlk/templates/include/quantizer.h#L25

primenumber:
`NHWC` or `HWNC` for kernel layout is widely used in dlk, it is better to rewrite.

primenumber:
I think `func_QTZ_binary_channel_wise_mean_scaling` should be removed.

PR #584, opened 2019-11-01
Fix dimension format for kernel

primenumber:
related #569.
Fix dimension format for kernel, HWNC -> HWOC, NHWC -> OHWC.

## Motivation and Context

In dlk, `HWNC` for kernel means Height, Width, Output Channels, Input Channels
But `N` is often confused with `N` for Batch size.

## Types of changes

## Checklist:


tkng:
Great PR, this problem is very confusing for developers.

BTW, how about changing the `C` to `I`?

I googled an found that other implementation like TVM, TensorFlow and ideep looks using `I` instead of `C`. I think we can make the code even less confusing by following their conventions.

- https://discuss.tvm.ai/t/tf-kernel-layout-hwio-vs-hwoi/3895
- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/conv_2d.h#L203
- https://github.com/pytorch/pytorch/blob/master/caffe2/ideep/operators/conv_transpose_op.cc#L33

primenumber:
Renamed `C` to `I`.

PR #586, opened 2019-11-01
Remove func_QTZ_binary_channel_wise_mean_scaling

primenumber:
## Motivation and Context

This function is not used anymore.

## Types of changes

## Checklist:


primenumber:
related #569 .

primenumber:
@tkng How about is this PR?

tkng:
Let me clarify my understanding. `QTZ_binary_channel_wise_mean_scaling` is always optimized out by optimizer, hence we don't need this implementation.

I'm worried that removing this implementation won't make it harder to test, but I don't have concrete evidence that it shouldn't be removed.

tkng:
Since restoring implementation is not difficult, now I'm thinking it's OK to merge.

Issue #473, opened 2019-10-01
Smarter memory allocation and reuse for DLK runtime


primenumber:
- Reduce memory consumption
- Erase useless memory copy
In dlk runtime, buffer size are known in compile time, we can optimize memory usage and memory transfer.

primenumber:
Some operations require temporary buffer to compute.
In current implementation, temporary buffer are managed by global `std::unique_ptr<T[]>` to avoid memory allocation in inference time.
It prevent concurrent inference and increase memory consumption.
We should consider to manage these buffer in `Network`.

primenumber:
Temporary buffer is now allocated in `Network::init` except RunTCA.

PR #595, opened 2019-11-07
The debug binary now show how much memory it will try to allocate on the heap

antonionevado:
Currently there is no way of knowing how much memory a network model requires for activations on deployment.

## Motivation and Context
It is possible to know the required memory from the Tensorflow model but will not match with the actual required amount of memory due to buffer reuse and quantization. This PR will make the debug binary (`lm_xxx.elf`) to print a message that show how much memory will be allocated on the heap **before** calling `Network::init()`. This allows you to have an idea of how much memory will be used and, debug possibly related memory issues.

## Description
- Added a new method to `Network` class that returns the amount of memory in bytes.
- Use this method from `mains/main.cpp` and print how many megabytes will be required by the model.

## How has this been tested?
Tested with 2 of the provided examples:
- classification: lmnet_quantized_cifar10
- object detection: widerface_v5

For object detection this will show:
```
This network will try to allocate 10.2753 MB on the heap
```

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/templates/include/network.tpl.h
+++ b/dlk/python/dlk/templates/include/network.tpl.h
@@ -28,6 +28,7 @@ public:
     Network();
     ~Network();
 
+    int memory();
     bool init();
 
     int get_input_rank();
--- a/dlk/python/dlk/templates/src/network.tpl.cpp
+++ b/dlk/python/dlk/templates/src/network.tpl.cpp
@@ -168,6 +168,32 @@ Network::~Network()
 #endif
 }
 
+int Network::memory() {
+  int total_mem_bytes = 0;
+  {% for node in graph.non_variables -%}
+  {% if node.available_buffer == '' -%}
+  {% for out_k in node.output_ops.keys() -%}
+  {% if node.output_ops.keys()|length > 1 -%}
+  total_mem_bytes += sizeof({{ node.dtype.cpptype() }}) * ({{ node.view.size_in_words_as_cpp }}); // {{ node.name + '_' + out_k }}_raw
+  {% else -%}
+  total_mem_bytes += sizeof({{ node.dtype.cpptype() }}) * ({{ node.view.size_in_words_as_cpp }}); // {{ node.name }}_raw
+  {%- endif %}
+  {%- endfor %}
+  {% elif node.available_buffer != '' and node.output_ops.keys()|length > 1 -%}
+  {% for out_k in node.output_ops.keys() -%}
+  {% if out_k != node.output_ops.keys()|list|first -%}
+  total_mem_bytes += sizeof({{ node.dtype.cpptype() }}) * ({{ node.view.size_in_words_as_cpp }}); // {{ node.name + '_' + out_k }}_raw
+  {%- endif %}
+  {%- endfor %}
+  {%- endif %}
+  {%- endfor %}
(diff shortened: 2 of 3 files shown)


dlk/python/dlk/templates/src/network.tpl.cpp
```diff
@@ -168,6 +168,32 @@ Network::~Network()
...
+  total_mem_bytes += sizeof({{ node.dtype.cpptype() }}) * ({{ node.view.size_in_words_as_cpp }}); // {{ node.name + '_' + out_k }}_raw
+  {%- endif %}
+  {%- endfor %}
+  {%- endif %}
+  {%- endfor %}
+#if !defined RUN_ON_FPGA
+  total_mem_bytes += sizeof(QUANTIZED_PACKED) * max_device_input_elems;
+  total_mem_bytes += sizeof(BIN_CONV_OUTPUT) * max_device_output_elems;
+#endif
+  return total_mem_bytes;
```
primenumber: There are several temporary buffer allocated by some operators (ex. kn2row_buf in `src/func/impl/generic/quantized_conv2d_kn2row.cpp`), but these are not counted yet.
I'm working to allocate these hidden temporary buffers in `Network::init()` ( #473 ), but it may take a while...
Until it is resolved, we should count these by hand...

antonionevado: Ah thank you! Didn't know about these buffers.

antonionevado: We can wait until they are allocated from `Network::init()`, if you want.
Or we can add it later and count by hand as you said :smile:

primenumber: These buffers are allocated by `std::make_unique`.

I found such buffers in following files:
- `src/func/conv2d.cpp`
- `src/func/*/batch_normalization.cpp`
- `src/func/impl/generic/quantized_conv2d_kn2row.cpp`
- `src/func/impl/arm_neon/quantized_conv2d_tiling.cpp`
- `src/func/impl/x86_avx/quantized_conv2d_tiling.cpp`
- `src/matrix/multiplication.cpp`

antonionevado: Ah, yes, seems that for changing layout we need these temporaries...

primenumber: When #473 is fixed, buffers may be unified into single buffer, total buffer size is obvious.
It seems waiting for fixed is reasonable way.

dlk/python/dlk/templates/include/network.tpl.h
```diff
@@ -28,6 +28,7 @@ class SYM_PUBLIC Network
     Network();
     ~Network();
 
+    int memory();
```
tkng: To keep consistency of API, it looks better to use verb for the method name. Considering that the responsibility of this function should output debug information, how about renaming to `get_debug_info`?

tkng: Also, I propose to change this API to return string, (or char*), so that we can see the buffer size layer by layer. (Is it possible to change the behavior of this API?)

antonionevado: Yes, it's true. The name can be better and we should return more detailed information.

PR #604, opened 2019-11-08
Change DLK test runner from nose2 to pytest

### LINKED ISSUE #602: Make dlk test executable with pytest command
Currently, we are running the dlk test with the setup.py command.

```
PYTHONPATH=python/dlk python setup.py test
```

We want to make dlk test executable with pytest command.

```
pytest
```

hadusam:
Close #602


dlk/tests/test_code_generation.py
```diff
@@ -416,7 +417,7 @@ def codegen_cpu(self,
 class TestCodeGenerationX8664(TestCodeGenerationBase):
     """Test class for code generation testing."""
 
-    @params(*get_configurations_x86_64())
+    @parameterized.expand(get_configurations_x86_64())
```
suttang: Can't use `@pytest.mark.parametrize`?

hadusam: I've tried it first, but unfortunately, `@pytest.mark.parametrize` does not work in `unittest.TestCase` subclasses.
https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses
`import parameterized` is an easy way to remove nose2 dependency without large structural changes(e.g. removing inheritance of unittest.TestCase).

suttang: OH... Thank you for your explanation!!

tkng:
LGTM, please merge after the conflict is resolved.

PR #587, opened 2019-11-01
Fix deprecated warnings

Joeper214:
## Motivation and Context

- fix #507 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature / Optimization (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #616, opened 2019-11-14
Fix module tensorflow._api.v1.io.gfile has no attribute `DeleteRecursively`

yd8534976:
After #587 #609 , there are still some lines missed.. This PR fixes them.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:


Joeper214:
Maybe we should test these methods as well? Seems like unit tests pass even if there's an error.

PR #630, opened 2019-11-21
BDD100K : ObjDet class update

ananno:
## Motivation and Context

BDD100K Object Detection Dataset class needs to be updated due to change in annotation file format.
## Description

Originally, the dataset annotation for object detection comes with xml files per image. The dataset class annotation reader function was implemented accordingly. But recently, BDD changes the annotation file format into JSON file each for train and test.
So, the annotation reader and dataset preparation needs to be updated

This implementation updates the dataset annotation reader and dataset preparation. This function is also optimized for optimum performance.

## How has this been tested?

`tox -e pytest`
## Types of changes

## Checklist:


lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -1,16 +1,15 @@
 # -*- coding: utf-8 -*-
-import functools
```
tkng: Could you apply `isort` to sort import orders?

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +92,58 @@ def __init__(self,
...
-            with open(f) as fp:
-                raw = json.load(fp)
-            objs = raw["frames"][0]["objects"]
+        img_paths = dict([(os.path.basename(path), path)
+                          for path in glob.glob(os.path.join(self.img_dir, "*.jpg"))])
+        img_names = set(img_paths.keys())
+
+        anno_data = json.load(open(self.anno_dir))
+
+        counts = 0
```
tkng: This variable looks better to remove.

lmnet/lmnet/datasets/bdd100k.py
```diff
@@ -74,41 +92,58 @@ def __init__(self,
...
+        max_boxes = 0
```
tkng: SInce you changed to return constant (`91`) from `num_max_boxes`, it looks better to remove this variable.

ananno:
@tkng, @iizukak San:
I've applied `isort` as requested. Please review one more time.

lmnet/lmnet/datasets/bdd100k.py
ananno: Done

lmnet/lmnet/datasets/bdd100k.py
ananno: Done

lmnet/lmnet/datasets/bdd100k.py
ananno: If you meant `counts` I can not remove it as it gives error `Variable not defined` but if you meant `max_boxes` I've removed that.

ananno:
@tkng San:
I've resolved your CR. Please review one more time.
---
Thank you.

lmnet/lmnet/datasets/bdd100k.py
tkng: In my understanding, the variable `counts` is only used in line 110 and line 114, it never return the information to the outside of this method, hence it could be removed.

ananno: I'm sorry. It is a shame I missed that. It is removed.

PR #637, opened 2019-11-25
Add runtime option to docker-compose

suttang:
## Motivation and Context

When we run `docker` or `docker-compose` on the machine which has not set default runtime to `nvidia`, we need to set runtime option.

So the purpose of this PR is to change default runtime to `nvidia` when use docker-compose.

## Description

- Change docker-compose version `3.6` to `2.4`
- Add runtime option to docker-compose

## How has this been tested?

```
DATA_DIR=/storage/dataset CUDA_VISIBLE_DEVICES=4 docker-compose run --rm blueoil python blueoil/cmd/main.py train -c docker-compose24-test.yml -e exp
```

## Types of changes

## Checklist:


hadusam:
@iizukak 
I am not so familiar with that but I found the following comment that v2 and v3 are for different use cases and v2 is not older than v3.
https://github.com/docker/compose/issues/6691#issuecomment-557133239
So, I think this change does not means downgrading but using a suitable version for us. 😄

iizukak:
@hadusam 
>  v2 is not older than v3.
Oh, that's... ok .

PR #640, opened 2019-11-27
Use more proper DATA_DIR in conftest.py

yd8534976:
In `init_env()`, `DATA_DIR` has been set to `blueoil_dir`. It is not proper. Since all dataset loaders except those who inherits `StoragePathCustomizable` will be broken. (They don't deal with `train_path` and `test_path` in config file.)
This PR makes an ordinary dataset loader can be used in `e2e` tests.
https://github.com/blue-oil/blueoil/blob/b45d0806a7daf8bff4af10e258c1e0b9d008953b/tests/e2e/conftest.py#L27-L33

## Types of changes

- [X] New feature / Optimization (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -25,7 +25,7 @@ def init_env():
 
     # TODO: Remove this setting after lmnet.environment has been refactored.
     envs = {
-        "DATA_DIR": blueoil_dir,
+        "DATA_DIR": os.path.join(blueoil_dir, "lmnet", "tests", "fixtures", "datasets"),
         "OUTPUT_DIR": train_output_dir.name,
         "_EXPERIMENT_DIR": os.path.join(train_output_dir.name, "{experiment_id}"),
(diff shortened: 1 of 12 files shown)


tests/e2e/conftest.py
```diff
@@ -25,7 +25,7 @@ def init_env():
 
     # TODO: Remove this setting after lmnet.environment has been refactored.
     envs = {
-        "DATA_DIR": blueoil_dir,
+        "DATA_DIR": os.path.join(blueoil_dir, "lmnet", "tests", "fixtures", "datasets"),
```
iizukak:
```suggestion
        "DATA_DIR": os.path.join(blueoil_dir, "lmnet/tests/fixtures/datasets"),
```

tsawada: I think we should do `os.path.join("a", "b", "c")` instead of `("a", "b/c")` for better cross-platform compatibility

iizukak: I see.

PR #643, opened 2019-11-29
Fix generate lmnet config flags (quantize 1st layer and dataset prefetch)

### LINKED ISSUE #612: generate dataset iterator prefetch flag from Blueoil prefetch flag
## tl;dr :
A straightforward fix is to include the prefetch flag in the _blueoil_to_lmnet function of blueoil/generate_lmnet_config.py

## detailed explanation :

For the DatasetIterator, there is a flag called enable_prefetch, which specifies whether to use prefetch or not. The value fed to it is a flag called prefetch, from the LMNet config. The LMNet config is generated from the Blueoil config by the _blueoil_to_lmnet function in the file blueoil/generate_lmnet_config.py

However, the _blueoil_to_lmnet function does not mention prefetch. When running python blueoil/cmd/main.py train, I also find the generated config.py file does not contain a value for the prefetch flag.

There exists a corresponding flag in the Blueoil config, called dataset_prefetch, inside of common. But, it looks like this flag is not used to generate a prefetch flag for the LMNet config.

So, I think this creates two problems : 1. prefetch defaults to not being used, but most users would probably prefer to have prefetch enabled. 2. the value of dataset_prefetch flag in the Blueoil config yml file does not reflect what is done in training.

Joeper214:
## Motivation and Context

Generate lmnet config from blueoil properly when setting `quantization at first layer` and `dataset prefech` flags.
- fix #612 

## Types of changes

## Checklist:


Joeper214:
Oh. I realized the quantized first convolution flag is only used in object detection. I'll omit this flag in other tasks.

blueoil/templates/lmnet/classification.tpl.py
```diff
@@ -85,6 +85,7 @@
 }
 NETWORK.WEIGHT_QUANTIZER = binary_mean_scaling_quantizer
 NETWORK.WEIGHT_QUANTIZER_KWARGS = {}
+NETWORK.QUANTIZE_FIRST_CONVOLUTION = {{ quantize_first_convolution }}
```
Joeper214: Oh. I realized the quantized first convolution flag is only used in object detection. I'll omit this flag in other tasks.

PR #648, opened 2019-12-04
Fix netlify setting

hadusam:
Currently, netlify build job is failing, I think something happened in Netlify with [build image selection](https://docs.netlify.com/configure-builds/get-started/?#build-image-selection).
Our setting is not following the latest official doc's setting. Following [official doc of netlify](https://docs.netlify.com/configure-builds/manage-dependencies/#python), we should specify the python version by using runtime.txt on the base directory for Netlify.


hadusam:
I found that Netlify seems to support the only python v2.7/3.7...🤔 It is wondering but now our build is fixed.
Please review this PR 👍

PR #658, opened 2019-12-11
Optimize batchnorm

primenumber:
## Motivation and Context

This PR optimize batch normalization operator.
Optimized operator calculate just multiply and add, and quit to use additional temporary buffer.

## Description

- Introduce `BatchNormalizationOptimized` operator
- convert `BatchNormalization` operator into optimized one if it is not merged into previous convolution
- Update `func_BatchNormalization`
- Remove temporary buffer for `func_BatchNormalization`

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/templates/src/func/generic/batch_normalization.cpp
+++ b/dlk/python/dlk/templates/src/func/generic/batch_normalization.cpp
@@ -20,7 +20,7 @@ limitations under the License.
 #include "func/batch_normalization.h"
 #include "time_measurement.h"
 
-void func_BatchNormalization(const TensorView<T_FLOAT, MemoryLayout::NHWC>& input,
+void func_BatchNormalizationOptimized(const TensorView<T_FLOAT, MemoryLayout::NHWC>& input,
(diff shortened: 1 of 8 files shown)


dlk/python/dlk/templates/src/func/generic/batch_normalization.cpp
```diff
@@ -20,32 +20,20 @@ limitations under the License.
 #include "func/batch_normalization.h"
 #include "time_measurement.h"
 
-static const auto scale = std::make_unique<float[]>(MAX_IN_C);
-static const auto shift = std::make_unique<float[]>(MAX_IN_C);
-
 void func_BatchNormalization(const TensorView<T_FLOAT, MemoryLayout::NHWC>& input,
```
tkng: For consistency, it might be better to rename this function to `func_BatchNormalizationOptimized`. What do you think?

primenumber: Renamed.

tkng: thanks!

PR #664, opened 2019-12-12
Allocate temporary buffers in Network::init

primenumber:
This PR is premise to solve #473 .
Make easier to manage memory.

## Motivation and Context

This PR allocate temporary buffers which is used operators in `Network::init`.
Memory allocation only occurs in `Network` and debug routines (such as time_measurement).

## Description

- Define unique_ptr in `Network` instead of global
- Stop to use `std::vector` in generic qconv operator.
    - This vector is used to manage threads, so changed to use OpenMP instead native threads.

## Types of changes

## Checklist:


tkng:
- [x] Readability Approval
- [x] Ownership Approval

PR #686, opened 2019-12-17
add a script to generate inference-pipline.yaml for buildkite

tfujiwar:
## Motivation and Context
We are trying to construct an automated benchmarking system with Buildkite. The benchmarking system consists of two main parts, one is a training part and the other is an inference part. These two parts will be defined as Buildkite Pipelines. This PR defines the inference pipeline.

## Description
This PR includes a script to generate the inference pipeline steps. This script is loaded and executed in the first step of the [inference pipeline](https://buildkite.com/blueoil/wip-benchmark-inference) like this.
(Please refer to the [official document](https://buildkite.com/docs/pipelines/defining-steps#dynamic-pipelines ) for detail.)
```yaml
steps:
  - label: "inference: upload pipeline"
    command: ".buildkite/inference-benchmark-pipeline.sh | buildkite-agent pipeline upload"
```

The first step above generates the following steps dynamically.
```yaml
steps:
  - label: "inference: setup"
    key: "setup"
    timeout_in_minutes: "30"
    agents:
      - "env=benchmark"
      - "agent-type=gcloudsdk"

    artifact_paths:
      - "convert-result.tgz"

    command: |
      gsutil -m cp "gs://path/to/result.tgz" ./


  - label: "inference: de10nano (lib/lib_arm.so)"
    depends_on: "setup"
    timeout_in_minutes: "30"

    agents:
      - "env=benchmark"
      - "agent-type=de10nano"

    artifact_paths:
      - "export/*/*/output/python/output/output.json"

    command: |
      buildkite-agent artifact download "convert-result.tgz" ./
      tar xvf convert-result.tgz
      cd export/*/*/output/python
      python run.py -i ../../inference_test_data/raw_image.png -c ../models/meta.yaml -m ../models/lib/lib_arm.so

  # 
  # All other cases come here...
  # 

  - label: "inference: raspberry-pi (lib/lib_aarch64.so)"
    depends_on: "setup"
    timeout_in_minutes: "30"

    agents:
      - "env=benchmark"
      - "agent-type=raspberry-pi"

    artifact_paths:
      - "export/*/*/output/python/output/output.json"

    command: |
      buildkite-agent artifact download "convert-result.tgz" ./
      tar xvf convert-result.tgz
      cd export/*/*/output/python
      python run.py -i ../../inference_test_data/raw_image.png -c ../models/meta.yaml -m ../models/lib/lib_aarch64.so
```


.buildkite/inference-benchmark-pipeline.sh
```diff
@@ -0,0 +1,80 @@
...
+#
+# Usage: CONVERT_RESULT_PATH=gs://path/to/result.tgz ./inference-benchmark-pipeline.sh
+
+# test cases to run inference
+# "<buildkite agent-type> <model filename>"
+TEST_CASES=(
+    "de10nano       lib/lib_arm.so"
+    "de10nano       lib/lib_fpga.so"
+
+    "jetson-nano    lib/lib_aarch64.so"
```
iizukak: aarch64 for Jetson is in the list of benchmark?
Raspberry is in the list.

.buildkite/inference-benchmark-pipeline.sh
```diff
@@ -0,0 +1,80 @@
...
+      - "export/*/*/output/python/output/output.json"
+
+    command: |
+      buildkite-agent artifact download "convert-result.tgz" ./
+      tar xvf convert-result.tgz
+      cd export/*/*/output/python
+EOS
+
+    # Generate commands to run inference (with and without "sudo")
+    if [ "${AGENT}" = "de10nano" ] && [ "${MODEL}" = "lib/lib_fpga.so" ] ; then
```
iizukak: Why only fpga inference need sudo?
lib_arm.so dose need?

iizukak:
Umm, inference pipeline can execute shell script...?

.buildkite/inference-benchmark-pipeline.sh
tfujiwar: Sorry, this was my mistake. Only GPU is enough for Jetson devices. But aarch64 for Raspberry Pi is needed.

iizukak: I see.

.buildkite/inference-benchmark-pipeline.sh
tfujiwar: I got an error message `permission denied, segmentation fault` without sudo.

This is because our binary for FPGA requires to access `/dev/mem`.
https://github.com/blue-oil/blueoil/blob/ce15ef3628af89a891dcf8a98f2af0174c7f0f18/dlk/python/dlk/templates/include/memdriver.h#L43-L54

And the code above is called only in the binary for FPGA.
https://github.com/blue-oil/blueoil/blob/ce15ef3628af89a891dcf8a98f2af0174c7f0f18/dlk/python/dlk/templates/src/network.tpl.cpp#L266-L267

tfujiwar:
@iizukak 
> How builkite execute this script?
> Human excite this shell script and then buidkite execute yaml?

In Buildkite, we can define pipelines as YAML file. We also can generate pipelines dynamically using shell scripts.

This script is executed in the first step of the inference pipeline. The stdout from the script is a YAML string that defines the following inference steps (We don't need to run this script manually).

Of course, we can define the inference pipeline as a static YAML file but it will be long and it contains many duplicated lines because inference steps are almost the same for all devices.

That's why I decided to prepare a script to generate inference steps.

.buildkite/inference-benchmark-pipeline.sh
iizukak: I see.

.buildkite/inference-benchmark-pipeline.sh
```diff
@@ -0,0 +1,75 @@
...
+    timeout_in_minutes: "30"
+    agents:
+      - "env=benchmark"
+      - "agent-type=gcloudsdk"
+
+    artifact_paths:
+      - "convert-result.tgz"
+
+    command: |
+      gsutil -m cp "${CONVERT_RESULT_PATH}" ./
```
hadusam: I think we should add `wait` after this step to wait for this setup before running inference steps.
https://buildkite.com/docs/pipelines/wait-step

.buildkite/inference-benchmark-pipeline.sh
```diff
@@ -0,0 +1,75 @@
...
+
+    agents:
+      - "env=benchmark"
+      - "agent-type=${AGENT}"
+
+    artifact_paths:
+      - "export/*/*/output/python/output/output.json"
+
+    command: |
+      buildkite-agent artifact download "convert-result.tgz" ./
```
hadusam: I think we should specify `--build $BUILDKITE_BUILD_ID` to download correct artifacts uploaded by the previous step.
See: https://buildkite.com/docs/agent/v3/cli-artifact#downloading-artifacts

.buildkite/inference-benchmark-pipeline.sh
tfujiwar: I set `depends_on: "setup"` to inference steps.
https://buildkite.com/docs/pipelines/dependencies

hadusam: Oh! So nice! 👍

.buildkite/inference-benchmark-pipeline.sh
tfujiwar: Thank you!! Fixed.

.buildkite/inference-benchmark-pipeline.sh
```diff
@@ -0,0 +1,75 @@
...
+
+    agents:
+      - "env=benchmark"
+      - "agent-type=${AGENT}"
+
+    artifact_paths:
+      - "export/*/*/output/python/output/output.json"
+
+    command: |
+      buildkite-agent artifact download "convert-result.tgz" ./ --build $BUILDKITE_BUILD_ID
```
hadusam:
```suggestion
      buildkite-agent artifact download "convert-result.tgz" ./ --build \$\${BUILDKITE_BUILD_ID}
```
I think it is better escaping `BUILDKITE_BUILD_ID` to run without `BUILDKITE_BUILD_ID` for testing this script locally.
In this situation, there are three times environment variable substitutions.

| How to write | Substitution Situation | stdout of shell | uploaded pipeline |
| --- | --- | --- | --- |
| `${BUILDKITE_BUILD_ID}` | uploading step (inference-benchmark-pipeline.sh) | `999` | `999` |
| `\${BUILDKITE_BUILD_ID}` | uploading step (buildkite-agent upload) | `${BUILDKITE_BUILD_ID}` | `999` |
| `\$\${BUILDKITE_BUILD_ID}` | inference step | `$${BUILDKITE_BUILD_ID}` | `${BUILDKITE_BUILD_ID}` |

Ref: https://buildkite.com/docs/agent/v3/cli-pipeline#environment-variable-substitution

Fortunately, `BUILDKITE BUILD_ID` is the same on each step of the same build, but I think it is better to use the env value of the inference step to avoid future potential bugs caused by referred to this script.

tfujiwar: Thank you so much!! I didn't think about this.

PR #696, opened 2019-12-20
WIP: Modifying cell size via config file

ananno:
## Motivation and Context

The default cell size for YOLOv2 is hardcoded to 32x32 px. But many object detection dataset has object labels less than 32px in height or weight. Those images are less likely to be matched with the ground truth anchors with a decent IoU. To mitigate this issue smaller cell size for YOLOv2 anchor grid is a good option.

## Description

This pull request gives user and options `CELL_SIZE` to be set in the config file. Using this configuration user can easily set the YOLOv2 cell size for the anchor grid. By having this parameter in the configuration will be make it easy to generate different config files with different cell size to run parametric training in batch processing using kubernetes.

## How has this been tested?

The implementation was verified with : `tox -e pytest`

## Types of changes

## Checklist:


iizukak:
@ananno 
Thanks. for WIP PR.

@yd8534976 
How do you think to add `CELL_SIZE` option to the config file?
I think you are training some small object detection.

tkng:
In my understanding, we do not have the flexibility to customize `CELL_SIZE` with a configuration file, because down sampling rate of an object detection network depends on the network itself.

Also, I'm skeptical about variable name `CELL_SIZE`. This variable name should be `DOWNSAMPLING_RATE` or something. Otherwise, line 112 (```self.num_cell = self.image_size[0] // cell_size, self.image_size[1] // cell_size```) is not understandable.

yd8534976:
> In my understanding, we do not have the flexibility to customize `CELL_SIZE` with a configuration file, because down sampling rate of an object detection network depends on the network itself.
> 
> Also, I'm skeptical about variable name `CELL_SIZE`. This variable name should be `DOWNSAMPLING_RATE` or something. Otherwise, line 112 (`self.num_cell = self.image_size[0] // cell_size, self.image_size[1] // cell_size`) is not understandable.

Yes! `STRIDE` should be less confusing for me.

ananno:
> > In my understanding, we do not have the flexibility to customize `CELL_SIZE` with a configuration file, because down sampling rate of an object detection network depends on the network itself.
> > Also, I'm skeptical about variable name `CELL_SIZE`. This variable name should be `DOWNSAMPLING_RATE` or something. Otherwise, line 112 (`self.num_cell = self.image_size[0] // cell_size, self.image_size[1] // cell_size`) is not understandable.
> 
> Yes! `STRIDE` should be less confusing for me.

@tkng San:
1. About the variable name. I choose this name to be consistant. In `yolo_v2.py` line [109](https://github.com/blue-oil/blueoil/blob/349b9eca4e37309599db90b70b718fa641704b74/lmnet/lmnet/networks/object_detection/yolo_v2.py#L109) and [111](https://github.com/blue-oil/blueoil/blob/349b9eca4e37309599db90b70b718fa641704b74/lmnet/lmnet/networks/object_detection/yolo_v2.py#L111) it is already calculating `self.num_cell` which is basically a tuple of the number of sub-division or regions YOLOv2 make on an image, each called a cell by convention. Hense the variable defining the size of the cell `CELL_SIZE`. But of course if this create confusion please suggest a better name for it.

@yd8534976 San:
Well, strides can also be used.

tkng:
> About the variable name. I choose this name to be consistant. In yolo_v2.py line 109 and 111 it is already calculating self.num_cell which is basically a tuple of the number of sub-division or regions YOLOv2 make on an image, each called a cell by convention. Hense the variable defining the size of the cell CELL_SIZE. But of course if this create confusion please suggest a better name for it.

It is still not clear for me what you indicated with the variable name `CELL_SIZE`, even after I read the line 109 and line 111. In my understanding, a cell doesn't have size. If you want to indicate "the size of the region in the original image which is allocated to one cell", it is difficult for me to imagine this meaning by this simple variable name.

More important issue is that the downsampling rate is network specific and cannot be changed without modification of network architecture. In other words, just changing the configuration file will not work. I am still not sure about when this pull request is effective. What do you think? @ananno

ananno:
> > About the variable name. I choose this name to be consistant. In yolo_v2.py line 109 and 111 it is already calculating self.num_cell which is basically a tuple of the number of sub-division or regions YOLOv2 make on an image, each called a cell by convention. Hense the variable defining the size of the cell CELL_SIZE. But of course if this create confusion please suggest a better name for it.
> 
> It is still not clear for me what you indicated with the variable name `CELL_SIZE`, even after I read the line 109 and line 111. In my understanding, a cell doesn't have size. If you want to indicate "the size of the region in the original image which is allocated to one cell", it is difficult for me to imagine this meaning by this simple variable name.
> 
> More important issue is that the downsampling rate is network specific and cannot be changed without modification of network architecture. In other words, just changing the configuration file will not work. I am still not sure about when this pull request is effective. What do you think? @ananno

@tkng San:
I'm sorry to hear that my explanation was inadequate. Well, the cell is exactly what you indicated by "region in the original image" which is sub-dividing the original image in spatial dimension for the anchors. Anyway, as YOLOv2 code usage a rectangular cell by default with 32px for height and width of the cell, using `CELL_HEIGHT` or `CELL_WIDTH` was not an option and in my opinion would create more confusion. Hence for the ambiguous name for both `CELL_HEIGHT` and `CELL_WIDTH` I choose `CELL_SIZE` which is exactly similar to `IMAGE_SIZE` represent the height and width of an image as tuple. Anyway, any suitable name can be used for this and of course we can use a tuple of same value to represent the same height and width.

Now for the downsampling issue, yes it is not possible to use any values for `CELL_SIZE`. But I think having this option available would give user a flexibility to try with difference `CELL_SIZE` they can play with to better match the ground truth data. And of course if anyone doesn't use it it will have default value `32` anyway. We can remove it from `config` file in such case. But I think having it in `YOLOv2` network file is a good option for those who wants to use it and knows what he/she is doing.

tkng:
@ananno Did you confirm the size of the feature map? Please let me know if you get the result.

tkng:
@ananno ping

ananno:
My apologies. I replied on this few days back but somehow may be I didn't post it as I can not find it. Anyway... here it is:
## Feature map
I've tested feature map size by changing `CELL_SIZE` but the feature map size doesn't depends on `CELL_SIZE` as shown in the following images.
* Cell Size : 32
cell_size_32
* Cell Size : 16
cell_size_16
* Cell Size : 8
cell_size_8

On the other hand, by changing `CELL_SIZE` will affect the last layer of the prediction output from the network. In the last layer it reshape the output of the network in function `_split_predictions` where the output shape is calculated as `[self.batch_size, num_cell_y, num_cell_x, self.boxes_per_cell, self.num_classes + 5]`. The `self.num_cell` depends on `CELL_SIZE`. This results in dimension mismatch during reshaping if the `CELL_SIZE` set anything other than `32` as shown in image below:

cell_size_error
So, we can not change `CELL_SIZE` without changing the network architecture, _which is not an option._

## Solution
One way to achieve the ability to select different `CELL_SIZE` is to modify the `output_filters` calculation in line [978](https://github.com/blue-oil/blueoil/blob/849941081fe5c7a54b0d3a4f20ad21db083492eb/lmnet/lmnet/networks/object_detection/yolo_v2.py#L978) in case of `YoloV2` or in line [127](https://github.com/blue-oil/blueoil/blob/849941081fe5c7a54b0d3a4f20ad21db083492eb/lmnet/lmnet/networks/object_detection/lm_fyolo.py#L127) in case of `LMFYolo`.

If the `CELL_SIZE` is considered as a scaling factor for the anchor boxes in calculation of `output_filters` as:
```python
(self.num_classes + 5) * self.boxes_per_cell * int(self.anchor_scaling * self.anchor_scaling)
```
The dimension mismatch issue during reshaping can be solved. Here `self.anchor_scaling = 2.0` results in double the number of cell in line [100~111](https://github.com/blue-oil/blueoil/blob/849941081fe5c7a54b0d3a4f20ad21db083492eb/lmnet/lmnet/networks/object_detection/yolo_v2.py#L107) of `YoloV2` network `init` function:
```python
        self.anchor_scaling = float(anchor_scaling)        // 2.0
        if self.is_dynamic_image_size:
            self.image_size = tf.tuple([
                tf.constant(self.image_size[0], dtype=tf.int32), tf.constant(self.image_size[1], dtype=tf.int32)
            ])

            # TODO(wakisaka): Be enable to change `32`. it depends on pooling times.
            # Number of cell is the spatial dimension of the final convolutional features.
            image_size0 = tf.cast(self.anchor_scaling * (self.image_size[0] / 32), tf.int32)
            image_size1 = tf.cast(self.anchor_scaling * (self.image_size[1] / 32), tf.int32)
            self.num_cell = tf.tuple([image_size0, image_size1])
        else:
            self.num_cell = int(self.anchor_scaling * self.image_size[0]) // 32, \
                            int(self.anchor_scaling * self.image_size[1]) // 32
```

## Feature Maps & Outputs
### Anchor Scale : 1.0
* Feature Maps
feature_map_1 0
* Outputs
output_image_1 0
### Anchor Scale : 2.0
* Feature Maps
feature_map_2 0
* Outputs
output_image_2 0

Feature Maps are same dimension but the output increases 4 times.

tkng:
We already able to change anchor scale by hand, in the `config.py`. I don't understand why do wan to introduce a new concept `ANCHOR_SCALE`.

Moreover, there is a better solution, say, implementing a new better network like SSD, YOLOv3, CenterNet, etc. I don't recommend to continue editing YOLOv2.

ananno:
> We already able to change anchor scale by hand, in the `config.py`. I don't understand why you wan to introduce a new concept `ANCHOR_SCALE`.
> 
> Moreover, there is a better solution, say, implementing a new better network like SSD, YOLOv3, CenterNet, etc. I don't recommend to continue editing YOLOv2.

@tkng San:
I understand your point. I'll close this PR. Thank you for your support.

PR #707, opened 2019-12-25
Use depends_on instead of wait for pipeline

hadusam:
Use the new feature: `depends_on` of buildkite pipeline. [Ref](https://buildkite.com/docs/pipelines/wait-step).
We don't need to wait for complete all steps to run the dependency steps.

### DIFF
--- a/.buildkite/dlk-pipeline.yml
+++ b/.buildkite/dlk-pipeline.yml
@@ -41,6 +41,7 @@ steps:
       BUILDKITE_CLEAN_CHECKOUT: 'true'
   - command: "make test-dlk-aarch64"
     label: "dlk: code_generation for aarch64"
+    key: "build_aarch64"
     agents:
     - "agent-type=normal"
     - "env=production"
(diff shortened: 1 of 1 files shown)


hadusam:
dlk-test: 12min -> 9min

.buildkite/dlk-pipeline.yml
```diff
@@ -49,11 +50,11 @@ steps:
       BUILDKITE_CLEAN_CHECKOUT: 'true'
     artifact_paths:
     - "output/TestCodeGenerationAarch64/**/*"
-  - wait
   - command: |
       buildkite-agent artifact download "output/TestCodeGenerationAarch64/*" ./ --build ${BUILDKITE_BUILD_ID}
       python3 tests/device_tests/test_device.py
     label: "dlk: run binary on aarch64"
```
tfujiwar: Currently, we have a `run binary` step only for aarch64. Will we add this step to all other devices in the future?

hadusam: @tfujiwar 
Yes, of course!
Currently, we run tests with updating FPGA IP on DE10-Nano by each test. So we need to consider how to update this by buildkite steps. But we will do it 👍

tfujiwar: Thank you! I understood.

PR #713, opened 2020-01-07
Optimize func_AveragePool

### LINKED ISSUE #701: Optimize AveragePool implementation
According to the performance measurement of LmResnet for ImageNet on FPGA, there is some low hanging fruits, `AveragePool` is now consuming over 100ms.

primenumber:
close #701 

~5~6x faster on DE10-Nano, ~11~13x faster on Raspberry Pi 3B+, 20x faster on Core i7-8750H

## Description

- Optimize loop structure
- Parallelize with OpenMP

## Types of changes

## Checklist:

### DIFF
--- a/dlk/python/dlk/templates/src/func/average_pool.cpp
+++ b/dlk/python/dlk/templates/src/func/average_pool.cpp
@@ -32,29 +32,37 @@ void func_AveragePool(const TensorView<T_FLOAT, MemoryLayout::NHWC>& input,
   int idx_out = 0;
   const T_FLOAT num_k_elems = app.kernel_height * app.kernel_width * app.kernel_depth;
 
-  std::memset(output.data(), 0.0f, app.output_channels * app.output_height * app.output_width * sizeof(T_FLOAT));
-
-  for(T_UINT oc = 0; oc < app.output_channels; oc++) {
-    for(T_UINT wi = 0; wi < app.output_height; wi++) {
-      for(T_UINT wj = 0; wj < app.output_width; wj++) {
-        T_FLOAT out = 0;
-        for(T_UINT ki = 0; ki < app.kernel_height; ki++) {
-          for(T_UINT kj = 0; kj < app.kernel_width; kj++) {
-            T_INT row = (wi * app.stride) - app.padding + ki;
-            T_INT col = (wj * app.stride) - app.padding + kj;
-
-            T_INT inside = (row >= 0 && col >= 0 && row < (T_INT) app.input_height && col < (T_INT)app.input_width);
-            if (!inside) continue;
-            for(T_UINT kz = 0; kz < app.kernel_depth; kz++) {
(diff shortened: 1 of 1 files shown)


tsawada:
Please put me back after ownership review

dlk/python/dlk/templates/src/func/average_pool.cpp
```diff
@@ -32,30 +32,38 @@ void func_AveragePool(const TensorView<T_FLOAT, MemoryLayout::NHWC>& input,
   int idx_out = 0;
```
tkng: It looks we can remove this `idx_out`. (In the new code, `idx_out` is defined at line 58.)

dlk/python/dlk/templates/src/func/average_pool.cpp
```diff
@@ -32,30 +32,38 @@ void func_AveragePool(const TensorView<T_FLOAT, MemoryLayout::NHWC>& input,
...
         }
-        output(0, wi, wj, oc) += T_FLOAT(out) / num_k_elems;
       }
     }
+    for(size_t oc = 0; oc < app.output_channels; oc++) {
+      size_t idx_out = wi * (app.output_width * app.output_channels)
+        + wj * app.output_channels
+        + oc;
+      float out = 0.f;
+      for(size_t kz = 0; kz < app.kernel_depth; kz++) {
```
tkng: It's just a comment, in most case, `app.kernel_depth` is 1. (To be honest, I haven't seen any other settings yet in my six years experience of deep learning.)

In this case, further optimization by SIMD instructions would be easy. (I understand SIMD optimization is out of scope. Please try when you have a free time.)

dlk/python/dlk/templates/src/func/average_pool.cpp
```diff
@@ -32,30 +32,38 @@ void func_AveragePool(const TensorView<T_FLOAT, MemoryLayout::NHWC>& input,
...
+        out += tmp[oc * app.kernel_depth + kz];
+      }
+      output.data()[idx_out] = out / num_k_elems;
```
tkng: How about multiplying inverse of `num_k_elems`, instead of deviding by `num_k_elems`? In my experiment, most floating operating is now fast, on recent CPUs, but division is still slow on such CPUs. (I haven't tried yet, please try if you feel it's worth trying.)

dlk/python/dlk/templates/src/func/average_pool.cpp
primenumber: Oh, yes.
Fixed!

dlk/python/dlk/templates/src/func/average_pool.cpp
primenumber: Implemented.
As a result, this function runs 10〜20% faster.

tkng:
OA
RA

PR #714, opened 2020-01-08
Adding support for pytest with tox for single test file

ananno:
## Motivation and Context

Often when we develop one python module in blueoil, to run unit testing we've to use `pytest' on entire `tests` directory. This is time consuming. A simpler and efficient way to test only one `test_file` is needed in blueoil to run unit testing for one test file.

## Description

Blueoil use `tox` to run the `unit testing` with proper environment setup. For python files test the command `tox -e pytest` is used from inside the `blueoil/lmnet` directory. But often we need to run one test file which is under development without wasting time to run full `pytest` unit testing on entire `tests` directory. This modification to `tox.ini` file add a command `pytest-only` which takes a command line argument and pass it on the the `tox` command. By this way we can run unit testing on a single python test file using `tox`.

* **Usage**:
```
[blueoil/lmnet] $ tox -e pytest-only -- path/to/test_file.py
```
Above command will load the python test environment and configuration and run the provided python test file via command line argument.

> NOTE : The `--` symbol after the `pytest-only` command is mandatory. Which is a way to pass command line argument to tox command. Anything after `--` symbol will be considered as argument.

## How has this been tested?

`tox -e pytest-only {path/to/single/test_file.py}`

 Screenshot from 2020-01-09 17-23-17

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/tox.ini
+++ b/lmnet/tox.ini
@@ -21,7 +21,7 @@ commands = flake8 .
 [testenv:pytest]
 commands = pytest tests/
 
-[testenv:single-pytest]
+[testenv:pytest-only]
 commands = pytest {posargs}
 
 [testenv:py36-pytest]
--- a/lmnet/tox.ini
+++ b/lmnet/tox.ini
@@ -21,6 +21,9 @@ commands = flake8 .
 [testenv:pytest]
 commands = pytest tests/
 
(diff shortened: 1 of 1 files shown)


tsawada:
This PR doesn't require readability review. Please add me back if there's anything specific for review

Issue #725, opened 2020-01-14
training does not hangup in certain conditions


joelN123:
Even after training has finished, the program doesn't hangup, and continues to use up resource. This is a problem when submitting a job to a time-costed resource, because it will continue to eat up costly GPU resource for no benefit.

I found this only occurred when using both multigpu and prefetch, so I believe it is due to pool processes being not able to close down. Also, if running only for a small number of steps, the bug does not happen, possibly because it is not enough time for pool processes to be built up.

#### How to reproduce:
Run training on ImageNet dataset, enable prefetch, run 4000 steps is enough to see the bug. To be specific, after cloning Blueoil, running 'make build', and creating ImageNet config, I ran this command:

docker run --runtime=nvidia -e CUDA_VISIBLE_DEVICES="0,1,2,3" -e PYTHONPATH=.:lmnet:dlk/python/dlk --rm -v `pwd`:/home/blueoil -w /home/blueoil/lmnet -v <path_to_dataset>:/home/blueoil/lmnet/dataset:ro --user=$(id -u):$(id -g) <docker_image> horovodrun -np 4 python executor/train.py -c configs/core/classification/<imagenet_config_file>.py -i testpoolclosing --recreate

The training runs fine, reaches the end, but does not hangup.

joelN123:
closed with https://github.com/blue-oil/blueoil/pull/726

PR #726, opened 2020-01-15
Ensure pool processes have closed

joelN123:
In the main training file, make sure pool processes have closed down after training has finished by manually shutting them down.

## Motivation and Context
This is to solve the bug described in issue https://github.com/blue-oil/blueoil/issues/725 where the training loop has finished but the program doesn't end. It is especially a concern when using time-costed services, if the program still runs after training has finished, it is a waste of costly resource.

## Description
A simple function is added, which accepts a dataset object, and terminates, closes and joins pools from the prefetcher of that dataset. This is only relevant if prefetch is being used, so first it checks if enable_prefetch is true.

The function is called at the end of training on each of the datasets: train, validation, and train_validation_saving (if there is one).

Honestly, I'm not sure why this fixes the problem, the dataset prefetcher should shut down own its own without it. But it doesn't in certain important cases.

## How has this been tested?
How to reproduce the bug is explained in issue https://github.com/blue-oil/blueoil/issues/725

To test that this fix does work, I ran training under the same settings, except with the code change that is in this pull request, and it runs fine, and fixes the bug.

## Types of changes

## Checklist:

### DIFF
--- a/lmnet/executor/train.py
+++ b/lmnet/executor/train.py
@@ -58,13 +58,6 @@ def setup_dataset(config, subset, rank):
     return DatasetIterator(dataset, seed=rank, enable_prefetch=enable_prefetch)
 
 
-def close_dataset(dataset):
-    if dataset.enable_prefetch == True:
-        dataset.prefetcher.terminate = True
-        dataset.prefetcher.pool.close()
-        dataset.prefetcher.pool.join()
-
-
 def start_training(config):
     use_horovod = horovod_util.is_enabled()
     print("use_horovod:", use_horovod)
@@ -351,10 +344,10 @@ def start_training(config):
         if rank == 0:
             progbar.update(step + 1)
     # training loop end.
-    close_dataset(train_dataset)
-    close_dataset(validation_dataset)
+    train_dataset.close_dataset()
+    validation_dataset.close_dataset()
     if use_train_validation_saving:
-        close_dataset(train_validation_saving_dataset)
+        train_validation_saving_dataset.close_dataset()
     print("Done")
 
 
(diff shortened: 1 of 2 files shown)


lmnet/executor/train.py
```diff
@@ -58,6 +58,13 @@ def setup_dataset(config, subset, rank):
     return DatasetIterator(dataset, seed=rank, enable_prefetch=enable_prefetch)
 
 
+def close_dataset(dataset):
```
iizukak: @joelN123 
Hi, thank you for try to fix the bug!!
`train.py` is the main loop of training.
It's better to implement `close_dataset` feature in `DatasetItretaor` class.
How about that?

joelN123: ah, right, as a method of datasetiterator is better, yes.

joelN123: I will make a change now and test it still is working. Thank you for the advice

joelN123: oh, I should also remove the == True in the if statement

joelN123: I made that change, and it runs fine (hangs up properly) still.

lmnet/lmnet/datasets/dataset_iterator.py
```diff
@@ -306,6 +306,12 @@ def get_shuffle_index(self):
         self.seed += 1
         return random_indices
 
+    def close_dataset(self):
```
tkng: Since what is closed here is trivial, how about changing the method name to `close`?

joelN123: yep, that's better name, thank you

joelN123:
run test

joelN123:
the "looks good to me" is also needed as a separate review?

Joeper214:
> the "looks good to me" is also needed as a separate review?

I think It's not required. This PR can be merged of you have OA and RA.

Issue #735, opened 2020-01-23
Wrong naming of predicted class directory in prediction of classification task


lm-jira:
Because `prediction["probability"]` stores values of probability in `string` type, not `float`
when finding the highest probability with the code below, it will find that value like `3.4567e-05`is greater than `0.9999` and results in wrong name of the predicted class.

https://github.com/blue-oil/blueoil/blob/446f9eb362cc863ea05914c2228252c113117fa7/lmnet/lmnet/utils/predict_output/output.py#L222

To solve it, I think the code below should convert probability to float.

https://github.com/blue-oil/blueoil/blob/446f9eb362cc863ea05914c2228252c113117fa7/lmnet/lmnet/utils/predict_output/output.py#L221

lm-jira:
@iizukak I found a simple bug described above.
Is it okay for me to create PR by myself and ask for your review?

iizukak:
@lm-jira 
Of course.
You can create PR anytime!!

PR #745, opened 2020-01-23
Convert prob to float before finding output class in classification prediction

lm-jira:
In prediction of classification task, the output class is calculated from output `json` file, which store probability value as `string`.
The probability value needs to be converted to `float` before finding the class with maximum probability.

## Motivation and Context

In string comparison, `3.0e-05` is greater than `0.99`, which results in wrong output class. (The output class will be the class with prob `3.03-05` instead of the class with prob `0.99`)
This PR aims to solve bug described in https://github.com/blue-oil/blueoil/issues/735.

## Description

Convert the probability to float before finding the class with maximum probability.

## Types of changes

## Checklist:


tkng:
Readability Approval

PR #748, opened 2020-01-23
Remove TensorFlow's bug workaround

tfujiwar:
The bug in TensorFlow was fixed in v1.14.0 and we are using TensorFlow v1.15.0.


blueoil/cmd/train.py
```diff
@@ -55,17 +55,7 @@ def save_config_file(config_file, dest_dir):
...
-    # We can remove following 2 lines once it's been resolved in tensorflow
-    # issue link: https://github.com/tensorflow/tensorflow/issues/28508
-    if gfile.exists(config_file_dest):
-        gfile.remove(config_file_dest)
-
-    return gfile.copy(
-        config_file,
-        config_file_dest
-    )
+    return gfile.copy(config_file, config_file_dest)
```
hadusam: Nice PR!
I think `overwrite=True` is needed.
https://github.com/blue-oil/blueoil/pull/258/files#diff-d1b064ecd00d49a72497a198ddfcf734L277

tfujiwar: Thank you!! You are right.

Issue #750, opened 2020-01-23
Docker image for CPU


hadusam:
## Introduction
We don't really need GPU for `convert` part, but the current docker image provided by us needs GPU driver to run.
I think it is better to provide a docker image for CPU.

## What is preventing this?
If we run `convert` with tensorflow-cpu, following error occurred.
```
self = <tensorflow.python.client.session.Session object at 0x7f26eaf44d30>
fn = <function BaseSession._do_run.<locals>._run_fn at 0x7f26f408abf8>
args = ({<tensorflow.python.pywrap_tensorflow_internal.TF_Output; proxy of <Swig Object of type 'TF_Output *' at 0x7f26ef166c...ywrap_tensorflow_internal.TF_Output; proxy of <Swig Object of type 'TF_Output *' at 0x7f26ef00be10> >], [], None, None)
message = "Node 'conv1/BatchNorm/FusedBatchNormV3' (type: '_FusedConv2D', num of outputs: 1) does not have output 1"
m = None

    def _do_call(self, fn, *args):
      try:
        return fn(*args)
      except errors.OpError as e:
        message = compat.as_text(e.message)
        m = BaseSession._NODEDEF_NAME_RE.search(message)
        node_def = None
        op = None
        if m is not None:
          node_name = m.group(3)
          try:
            op = self._graph.get_operation_by_name(node_name)
            node_def = op.node_def
          except KeyError:
            pass
        message = error_interpolation.interpolate(message, self._graph)
        if 'only supports NHWC tensor format' in message:
          message += ('\nA possible workaround: Try disabling Grappler optimizer'
                      '\nby modifying the config for creating the session eg.'
                      '\nsession_config.graph_options.rewrite_options.'
                      'disable_meta_optimizer = True')
>       raise type(e)(node_def, op, message)
E       tensorflow.python.framework.errors_impl.OutOfRangeError: Node 'conv1/BatchNorm/FusedBatchNormV3' (type: '_FusedConv2D', num of outputs: 1) does not have output 1

/usr/local/pyenv/versions/3.6.3/envs/python3.6/lib/python3.6/site-packages/tensorflow_core/python/client/session.py:1384: OutOfRangeError
```
The details of this error can see #106 .
This error happened in `lmnet/export` and this feature is for debugging and testing. 
https://github.com/blue-oil/blueoil/blob/4ad07e2ea85b005188f1fa0db1cc1d0daefcc1d1/lmnet/executor/export.py#L112-L137

[...]

PR #752, opened 2020-01-23
Change the default value of save_npy_for_debug for convert command

hadusam:
Related #750 
`save_npy_for_debug` is for debug, so I would like to change the default value from `True` to `False`.


iizukak:
@hadusam 
Thank you for adding options.
I think adding option is not problem.

But, We want npy files often to run lm_*.so files.
The default `True` value is good for me.

hadusam:
@iizukak 
Oh, thank you!
I understand we often need npy files. 😱 
In this situation, what do you think about #750 ?
Is it better not to provide CPU docker image currently? Should we wait to fix the root cause of the error of saving npy files on CPU mode? 🤔

tkng:
> In this situation, what do you think about #750 ?

I think a better answer is to make `save_npy_for_debug` option work on CPU version of Docker Image. (If possible.)

hadusam:
After #789, we can save `npy` files without errors, so this change is not necessary now.

PR #751, opened 2020-01-23
Create Dockerfile for CPU and replace docker image for tests

hadusam:
Related #750 

Diff between `Dockerfile` and `Dockerfile_cpu`
```
$ diff docker/Dockerfile docker/Dockerfile_cpu 
1c1
< FROM nvidia/cuda:10.0-cudnn7-devel-ubuntu16.04 AS base
---
> FROM ubuntu:16.04 AS base
11,13c11
< ENV LD_LIBRARY_PATH=/usr/local/cuda/extras/CUPTI/lib64:${LD_LIBRARY_PATH} \
<     CUDA_HOME=/usr/local/cuda-10.0 \
<     PYENV_ROOT=/usr/local/pyenv
---
> ENV PYENV_ROOT=/usr/local/pyenv
54,56d51
< # Install OpenSSH for MPI to communicate between containers
< RUN mkdir -p /var/run/sshd
< 
79,96d73
< # Install OpenMPI
< RUN mkdir /tmp/openmpi && \
<     cd /tmp/openmpi && \
<     wget https://www.open-mpi.org/software/ompi/v4.0/downloads/openmpi-4.0.1.tar.gz && \
<     tar zxf openmpi-4.0.1.tar.gz && \
<     cd openmpi-4.0.1 && \
<     ./configure --enable-orterun-prefix-by-default && \
<     make -j $(nproc) all && \
<     make install && \
<     ldconfig && \
<     rm -rf /tmp/openmpi
< 
< # Configure OpenMPI to run good defaults:
< #   --bind-to none --map-by slot --mca btl_tcp_if_exclude lo,docker0 --mca btl_vader_single_copy_mechanism none
< RUN echo "hwloc_base_binding_policy = none" >> /usr/local/etc/openmpi-mca-params.conf && \
<     echo "rmaps_base_mapping_policy = slot" >> /usr/local/etc/openmpi-mca-params.conf && \
<     echo "btl_tcp_if_exclude = lo,docker0" >> /usr/local/etc/openmpi-mca-params.conf
< 
102c79
< RUN pip install -r /home/blueoil/gpu.requirements.txt
---
> RUN pip install -r /home/blueoil/cpu.requirements.txt
108,119d84
< 
< # Set env to install horovod with nccl and tensorflow option
< ENV HOROVOD_GPU_ALLREDUCE=NCCL \
<     HOROVOD_WITH_TENSORFLOW=1
< 
< # Install requirements for distributed training temporarily using CUDA stubs
[...]

PR #753, opened 2020-01-23
Add test for Dockerfile for GPU by running e2e test once

hadusam:
Related #750
After #751, we don't have tests run on the GPU environment.
We have Dockerfile for GPU, so we need to test that Docker image, it is enough to run e2e test once, I think. So I added it.


hadusam:
Our docker image is already supported for running without GPU, see #750.
So now we don't need another docker image specified for the CPU only, I think.

PR #755, opened 2020-01-24
Add target x86_avx to make_list of convert command

hadusam:
There are 5 types of targets in Makefile but convert command does not have a target of `x86_avx`.
So I added it 😄 

https://github.com/blue-oil/blueoil/blob/d389b8c6a3f8d8347323359bbbeb3acd8b36cae4/dlk/python/dlk/templates/Makefile#L70-L78

### DIFF
--- a/blueoil/cmd/convert.py
+++ b/blueoil/cmd/convert.py
@@ -80,9 +80,9 @@ def strip_binary(output):
 
     """
 
-    if output == "lm_x86.elf":
+    if output in {"lm_x86.elf", "lm_x86_avx.elf"}:
         subprocess.run(("strip", output))
-    elif output == "lib_x86.so":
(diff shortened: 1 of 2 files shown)


hadusam:
@tkng 
Could you review it for readability? 🙇 Thanks!

blueoil/cmd/convert.py
```diff
@@ -122,7 +125,7 @@ def make_all(project_dir, output_dir):
 
     # Make each target and move output files
     for target, output in make_list:
-        if target in ["lm_x86", "lm_arm", "lm_fpga", "lm_aarch64"]:
+        if target in ["lm_x86", "lm_x86_avx", "lm_arm", "lm_fpga", "lm_aarch64"]:
```
tsawada: make this a set as well?

Issue #743, opened 2020-01-23
Unit test for DLK C++ runtime


primenumber:
- [ ] refactor codes which enable to be built almost of all operator's code without jinja template generator
    - [ ] split `global.h` into `types.h` and `global.h`, some type definitions not depends on jinja.
    - [ ] modify codes to use `types.h` instead of `global.h` if it can
- [ ] select unit test library if we need
- [ ] write unit test code

PR #764, opened 2020-01-27
jinja free tensor_view.h

primenumber:
A part of #743 .
`tensor_view.h` is now free from jinja!

## Motivation and Context

For testability of C++ codes, operators should not depend on jinja.
TensorView class doesn't depend on `global.h`, so we can use this class without jinja in theory.
This PR enable it.

## Description

- Define `***_kernel_t` instead of `kernel_t`

## Types of changes

## Checklist:


iizukak:
great pr

PR #765, opened 2020-01-27
Use Set instead of List for membership testing

hadusam:
In the case of membership testing(`if a in b`), it is better to use Set as `b` instead of List because using Set(hashable objects) is faster than List.
> A set object is an unordered collection of distinct hashable objects. Common uses include **membership testing**, ...

https://docs.python.org/3.6/library/stdtypes.html#set-types-set-frozenset

> Sets, however, are significantly faster than lists if you want to check if an item is contained within it. They can only contain unique items though.

https://stackoverflow.com/questions/2831212/python-sets-vs-lists

I think this is not so critical, but it is better to follow this principle. So I refactored codes.
Related: https://github.com/blue-oil/blueoil/pull/755#discussion_r371079463

### DIFF
--- a/blueoil/cmd/convert.py
+++ b/blueoil/cmd/convert.py
@@ -125,7 +125,7 @@ def make_all(project_dir, output_dir):
 
     # Make each target and move output files
     for target, output in make_list:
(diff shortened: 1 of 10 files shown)


tkng:
Readability Approval

Issue #762, opened 2020-01-27
Move and rename dlk to blueoil/converter


iizukak:
We want to provide Blueoil as just one python package.
But, current Blueoil have `dlk` as independent python package.
Our main package directory is `blueoil`, So we need to move scripts in `dlk` to `blueoil`.

Here is the how move dlk.

- [x] `dlk/python/dlk/*` -> `blueoil/converter/*`
- [x] `dlk/python/tools` -> `blueoil/tools`
- [x] `dlk/utils` -> `blueoil/converter/utils`
- [x] `dlk/tests` -> `tests/unit/converter`
- [x] `dlk/.gitignore` -> `.gitignore` merge to root `.gitignore`
- [x] `dlk/.gitmodules` -> removed
- [x] `dlk/README.md` -> removed. Most of the text are deprecated
- [x] `dlk/setup.py` -> `setup.py` merge to root `setup.py`, include some requirements.
- [x] `dlk/setup.cfg` -> removed
- [x] `dlk/hw/intel/de10_nano` -> `hw/intel/de10_nano`
- [x] `dlk/docs` -> `docs/converter`
- [x] `dlk/examples` -> `blueoil/converter/examples`

We also need to change python import path.
This is big issue. We need to do in several PRs.

iizukak:
@primenumber 
Is it ok to move dlk to blueoil directory?
I think it will conflict your project.

primenumber:
@iizukak OK!

iizukak:
We create `tests/unit` directory for unit tests.
This issue's structure is also changed.

iizukak:
@kchygoe Thanks. We can close this issue.

kchygoe:
👍

PR #786, opened 2020-02-04
Merge dlk docs into blueoil docs

kchygoe:
## Motivation and Context

dlk/docs has been not used long time.
we can merge this into blueoil docs

## Description

- moved dlk/docs into docs/
- change some documents index
- remove deprecated links

## How has this been tested?

Locally uses sphinx

## Types of changes

## Checklist:


docs/conf.py
```diff
@@ -32,11 +32,15 @@
...
 import sys
+from typing import Dict
 
+import sphinx_rtd_theme
 from recommonmark.parser import CommonMarkParser
 from recommonmark.transform import AutoStructify
 
 sys.path.insert(0, os.path.abspath('../lmnet'))
+sys.path.insert(0, os.path.abspath('../dlk/python'))
+# sys.path.insert(0, os.path.abspath('../dlk/python/dlk'))
```
iizukak: Can you remove this comment line?

kchygoe: oh, sorry. thank you for the check!

kchygoe: fixed

docs/conf.py
```diff
@@ -32,11 +32,14 @@
 
 import os
 import sys
+from typing import Dict
 
+import sphinx_rtd_theme
```
hadusam: Are these necessary? Sorry, I cannot find lines using them. 🤔

kchygoe: let me check. 
Regarding the documents, it would be better to have another PR

kchygoe:
```
# The theme to use for HTML and HTML Help pages.  See the documentation for
# a list of builtin themes.
#
#html_theme = 'alabaster'
html_theme = 'sphinx_rtd_theme'
```
I supposed the theme is used in this but it would be dynamically loaded. I'll test it and let you know

kchygoe: I checked that this is not required. thank you for the review! Fixed.

iizukak:
`conf.py` are python scripts. So we need Readability Reviewer.

docs/conf.py
```diff
@@ -32,11 +32,13 @@
 
 import os
 import sys
+from typing import Dict
```
tsawada: why, is this used somewhere?

kchygoe: no, I have merged just the diffs from dlk/docs which is not correct since the first commit. fixed!
thank you for the review

kchygoe:
this is pat of https://github.com/blue-oil/blueoil/issues/762

PR #793, opened 2020-02-06
[WIP] Move models for dlk tests

hadusam:
Related #477 , move models for dlk tests from `dlk/example` to `dlk/tests/fixtures/models`.
After this PR, I will remove `dlk/example` directory with some unused models. #794


hadusam:
already fixed by #847

Issue #781, opened 2020-02-03
Remove convert_weight_from_darknet related scripts


iizukak:
In early development stage of Blueoil, someone use `convert_weight_from_darknet` but nobody maintein the script.

- [x] Remove config using  `convert_weight_from_darknet`
- [x] Remove `convert_weight_from_darknet` script

Issue #746, opened 2020-01-23
Move code in lmnet/executor directory to blueoil/cmd directory.


tfujiwar:
This is one of the tasks of #741 .

The following scripts exist both in `lmnet/executor` and `blueoil/cmd`. We need to merge them into one. Basically the scripts in `blueoil/cmd` will be replaced with the scripts in `lmnet/executor` but we need to continue supporting both "lmnet config" and "blueoil config" in the new scripts.
- [x] predict.py
- [x] train.py

The other scripts exist only in `lmnet/executor` directory. We can just move them and change import paths.
- [x] other scripts

tfujiwar:
Done

PR #795, opened 2020-02-06
Move executor scripts which exist only in “lmnet/executor”

tfujiwar:
Related to #746 .

This PR moves executor scripts which exist only in `lmnet/executor` to `blueoil/cmd`. This change has no conflicts with the existing scripts in `blueoil/cmd`. So the target scripts are simply moved. Paths in code, comments and docs which point the target scripts are updated.

Target scripts:
- build_tfds.py
- convert_weight_from_darknet.py (will be deleted: #781)
- evaluate.py
- export.py
- measure_latency.py
- output_event.py
- profile_model.py
- profile_template.md
- tune_ray.py


iizukak:
@tfujiwar Wow, this PR is what I want!!

tfujiwar:
@iizukak 
I agree with putting images (and `profile_template.md` also) in the other folder but I'm not sure it's good to put them inside `tests` folder because they are not for testing.

How about `blueoil/cmd/fixtures` ?

iizukak:
`blueoil/cmd/fixtures` is nice.

PR #800, opened 2020-02-07
Use blueoil as python package

iizukak:
For the future, We want to use `blueoil` as python package.
We need to just put `__init__.py` to `blueoil` directory.

In Docker, we already execute `setup.py install`.

We can use blueoil like

```
Python 3.6.8 (default, Dec 24 2018, 19:24:27)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from blueoil.cmd import train
>>> train.train
<function train at 0x2b4ddd374f28>
```

## Types of changes

## Checklist:


iizukak:
@tsawada Please review as readability reviewer.

Issue #796, opened 2020-02-06
Move lmnet/lmnet to blueoil/*


iizukak:
This issue is a part of #741 .

We decide to move `lmnet/lmnet` to `blueoil`.

- [x] Create `blueoil/nn` directory and put `__init__.py`.
- [x] Move `lmnet/lmnet/datasets` 
- [x] Move `lmnet/lmnet/layers`
- [x] Move `lmnet/lmnet/metrics`
- [x] Move `lmnet/lmnet/networks`
- [x] Move `lmnet/lmnet/utils`
- [x] Move `lmnet/lmnet/quantizations`
- [x] Move `lmnet/lmnet/*.py`

PR #810, opened 2020-02-08
Move lmnet/lmnet/datasets to blueoil/datasets

iizukak:
This PR is a part of #796
I moved `lmnet/lmnet/datasets` to `blueoil/datasets`

## Checklist:


odoku:
@iizukak 
I don't think Datasets are strictly related to Neural Network.
So it seems unnatural to have datasets under the nn directory. What do you think?

I think it is easy to discuss if there is a definition of what the nn module handles.

odoku:
You said you're not sure it's right to have the module directory at the same level as the cmd directory, but it doesn't seem to matter as far as you look at the famous libraries.

* [Scrapy](https://github.com/scrapy/scrapy/blob/master/setup.py#L43)
* [flake8](https://github.com/PyCQA/flake8/blob/master/setup.cfg#L62)
* [mypy](https://github.com/python/mypy/blob/master/setup.py#L183)

iizukak:
@odoku 
Year, I agreed with you.

iizukak:
@odoku `blueoil/nn/datasets` to `blueoil/datasets` moved. please review.

Issue #812, opened 2020-02-10
Move requirements.txt to setup.cfg


iizukak:
We want ot install blueoil like `python setup.py install`
So we need to move contents in `requirements.txt` to `setup.cfg`.

kchygoe:
🎉

iizukak:
fin~

Issue #505, opened 2019-10-04
Unify requirements.txt


kchygoe:
There are many requirements.txt in the project. These should be cleaned up.
At the same time, unused packages should be removed if there are.

```
./docs/requirements.txt
./lmnet/dist.requirements.txt
./lmnet/requirements.txt
./lmnet/docs.requirements.txt
./lmnet/cpu.requirements.txt
./lmnet/dev.requirements.txt
./lmnet/gpu.requirements.txt
./lmnet/test.requirements.txt
```

kchygoe:
After https://github.com/blue-oil/blueoil/issues/534

kchygoe:
This issue is no longer required! thx @iizukak !!

PR #825, opened 2020-02-10
Remove *.requirements.txt and Add requires to setup.cfg

iizukak:
Solve #812 #505 

## Types of changes

## Checklist:


setup.cfg
```diff
@@ -13,10 +13,39 @@ zip_safe = False
...
+  tensorflow-datasets==1.0.2
+  whaaaaat==0.5.2
+  cython==0.27.3
+
+[options.extras_require]
+cpu =
+  tensorflow==1.15.2
+gpu =
+  tensorflow-gpu==1.15.2
+  gast==0.2.2
```
kchygoe: nits: `gast` looks just the dependency of tensorflow

iizukak: @kchygoe Wow, then we can remove this?

docker/Dockerfile
```diff
@@ -94,12 +94,13 @@ RUN echo "hwloc_base_binding_policy = none" >> /usr/local/etc/openmpi-mca-params
     echo "rmaps_base_mapping_policy = slot" >> /usr/local/etc/openmpi-mca-params.conf && \
     echo "btl_tcp_if_exclude = lo,docker0" >> /usr/local/etc/openmpi-mca-params.conf
 
+WORKDIR /home/blueoil
```
iizukak: @kchygoe I changed `WORKDIR` directory here for `pip install`. It's OK/

kchygoe: ah, I think it's okey but we have to check if it is okey to install setup.py without copying all other files.
If setup.py/cfg doesn't require `blueoil/*` files, we still need to install `blueoil` at the end of dockerfile

kchygoe: Regarding `gast`,  let's clean up all requirements later ( I want to check all versions as well)

iizukak: @kchygoe I do not want to `pip install -e` several times.
How about to execute `COPY` before this line?

kchygoe: I checked CI and `setup.py` works with the args. so can you just add `pip install -e .` after `COPY blueoil /home/blueoil/blueoil`? it would install blueoil itself

iizukak:
@kchygoe I see. I'll try after your #847 merged.

iizukak:
@kchygoe Some change applied. Please check again.

Issue #741, opened 2020-01-23
Move code in lmnet directory to blueoil directory.


iizukak:
`lmnet` directory name confuse user often.
It's means both network name and whole training system name.
We want to move things in `lmnet` to `blueoil`

- [x] `lmnet/configs` - > `blueoil/configs`
- [x] `lmnet/executor` - > `blueoil/cmd`
- [x] `lmnet/tests` -> `tests/`
- [x] `lmnet/lmnet` -> `blueoil/`
- [x] `lmnet/third_party` -> removed
- [x] `lmnet/tox.ini` -> removed
- [x] `lmnet/README.md` -> `docs/something.md` README shoud be edited. because it's contains deprecated sentence.

This is big issue!!
All of import code should be replace with new path.
We should done in several PRs.

In this issue, we will ignore sort order. After that we sort order again in #817

iizukak:
@tfujiwar Issue created.

iizukak:
I updated the directory structure.

Now I consider to move `lmnet/lmnet` to `lmnet/nn`, Because in `blueoil` directory, we have `cmd` and `converter` in the future. In this case, I think scripts in `lmnet/lmnet` is not same level as `cmd` and `converter`, it's good to add layer like `lmnet/lmnet`. Disucussed with @odoku and @suttang , I decide the new layer name as `nn`. The name comes from TensorFlow and PyTorch.

tfujiwar:
Done.

PR #843, opened 2020-02-14
Move lmnet/tests to tests/unit

iizukak:

- Move `lmnet/tests` to `tests/unit`
- Change Makefile target name
- Change hardcoded path in `blueoil/environment.py`.

I know this PR is too big, but most of the files are just moved.

This PR is a part of #741


tests/e2e/conftest.py
```diff
@@ -99,7 +99,7 @@ def run_all_steps(dirs, config_file):
     assert os.path.exists(os.path.join(lib_dir, 'lm_x86.elf'))
 
     # Predict
-    predict_input_dir = os.path.join(dirs["blueoil_dir"], "lmnet/tests/fixtures/sample_images")
+    predict_input_dir = os.path.join(dirs["blueoil_dir"], "tests/unit/fixtures/sample_images")
```
tsawada: this should be split but you can fix this later

PR #844, opened 2020-02-14
Change flake8 target lmnet to blueoil

iizukak:

- This PR change `flake8` target lmnet to blueoil
- Several pep8 error occur. Fixed.
- Change make target name `test-lmnet-pep8` to `test-blueoil-pep8`

This PR is a part of #741


blueoil/cmd/init.py
```diff
@@ -131,7 +129,9 @@
     ("cosine", "'cosine' -> cosine learning rate."),
     ("2-step-decay", "'2-step-decay' -> learning rate decrease by 1/10 on {epochs}/2 and {epochs}-1."),
     ("3-step-decay", "'3-step-decay' -> learning rate decrease by 1/10 on {epochs}/3 and {epochs}*2/3 and {epochs}-1"),
-    ("3-step-decay-with-warmup", "'3-step-decay-with-warmup' -> warmup learning rate 1/1000 in first epoch, then train the same way as '3-step-decay'"),
+    ("3-step-decay-with-warmup",
```
odoku: It is recommended to write long sentences like this!

```python
s = (
    "spam "
    "egg "
    "spam"
)

print(s == "spam egg spam")
```

https://www.python.org/dev/peps/pep-0008/#maximum-line-length

iizukak: @odoku changed. this is what you want?

blueoil/cmd/init.py
```diff
@@ -131,7 +129,11 @@
     ("cosine", "'cosine' -> cosine learning rate."),
     ("2-step-decay", "'2-step-decay' -> learning rate decrease by 1/10 on {epochs}/2 and {epochs}-1."),
     ("3-step-decay", "'3-step-decay' -> learning rate decrease by 1/10 on {epochs}/3 and {epochs}*2/3 and {epochs}-1"),
-    ("3-step-decay-with-warmup", "'3-step-decay-with-warmup' -> warmup learning rate 1/1000 in first epoch, then train the same way as '3-step-decay'"),
+    (
+     "3-step-decay-with-warmup",
+     "'3-step-decay-with-warmup' -> \
+      warmup learning rate 1/1000 in first epoch, then train the same way as '3-step-decay'"
+    ),
```
tfujiwar: This backslash doesn't work because it's included in a string. And as @odoku -san mentioned, using backslashes for wrapping long lines is not recommended.

In this case, code like this would be fine.
```python
(
    "3-step-decay-with-warmup",
    "'3-step-decay-with-warmup' -> "
    "warmup learning rate 1/1000 in first epoch, then train the same way as '3-step-decay'"
),
```

iizukak: Wow I see.

iizukak: @tfujiwar Changed with 1577c22

PR #845, opened 2020-02-14
Add divide 255 pre processor test

Joeper214:
Increase code coverage by adding pre_processor unit test code for Divide255 Class.

### DIFF
--- a/tests/unit/test_pre_processor.py
+++ b/tests/unit/test_pre_processor.py
@@ -47,7 +47,7 @@ def test_resize():
 
 
 def test_divide_255():
-    orig_image = np.full((1, 1, 3), 1)
+    orig_image = np.ones((1, 1, 3))
     expect = np.array([[[0.00392157, 0.00392157, 0.00392157]]])
 
     pre_processor = DivideBy255()
--- a/tests/unit/test_pre_processor.py
+++ b/tests/unit/test_pre_processor.py
@@ -18,6 +18,7 @@ import pytest
 
 from blueoil.pre_processor import (
     Resize,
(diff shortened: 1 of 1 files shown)


tests/unit/test_pre_processor.py
```diff
@@ -45,6 +46,20 @@ def test_resize():
...
+def test_divide_255():
+    orig_image = np.full((1, 1, 3), 1)
+    expect = np.array([[[0.00392157, 0.00392157, 0.00392157]]])
+
+    pre_processor = DivideBy255()
+    processed = pre_processor(image=orig_image)
+    processed_imaged = processed["image"]
+
+    assert isinstance(processed_imaged, np.ndarray)
+    assert isinstance(processed_imaged, np.ndarray)
```
tfujiwar: Remove one of them.

Joeper214: Fixed! thanks!

tests/unit/test_pre_processor.py
```diff
@@ -45,6 +46,19 @@ def test_resize():
     assert resized_mask.shape[2] == 3
 
 
+def test_divide_255():
+    orig_image = np.full((1, 1, 3), 1)
```
iizukak: To initialize array with 1, I prefer to use `np.ones`

iizukak:
```suggestion
    orig_image = np.ones((1, 1, 3))
```

Joeper214: fixed. Thanks!

Issue #603, opened 2019-11-08
Merge dlk/docs to docs of top directory.


suttang:
dlk/docs -> docs

kchygoe:
DONE!

PR #847, opened 2020-02-16
Move DLK to blueoil/converter 

kchygoe:

Related to https://github.com/blue-oil/blueoil/issues/762 Mainly 
- dlk/python/dlk => blueoil/converter 
- dlk/tests => tests/converter 
- others along with above changes

https://github.com/blue-oil/blueoil/issues/762


tests/converter/test_binary.py
```diff
@@ -18,7 +18,7 @@
 import unittest
 from os.path import join
 
-from scripts import generate_project as gp
+from blueoil.converter import generate_project as gp
```
iizukak: If tests in `dlk/tests` if unit test, it's preffer to put these test files in `tests/unit/converter`.

iizukak:
https://github.com/blue-oil/blueoil/blob/master/output_template/python/lmnet/nnlib.py
This file is a symlink to dlk. We should replace.

iizukak:
`setup.py` in dlk looks just removed. It's ok?
How about `install_requires` ?

tests/converter/test_binary.py
kchygoe: oh, okey

kchygoe:
> `PYTHONPATH`
I've forgotten it! Let me remove!
>

tests/converter/test_binary.py
kchygoe: I suppose these are not unittest... it would be better to re-define unittests for converter. can we discuss those in another issue?

kchygoe:
Notes
- `blueoil/converter` flake8 adoption should be in another PR.
- `tests/converter` should be refactored by new unittests
- `tests/fixtures` should be removed and placed on other server as artifacts

tests/converter/test_binary.py
iizukak: @kchygoe I see. OK to put tests in `tests/converter`

.buildkite/dlk-pipeline.yml
```diff
@@ -62,4 +54,4 @@ steps:
     env:
       BUILDKITE_CLEAN_CHECKOUT: 'true'
       DEVICE_TEST_INPUT_PATH: './output/TestCodeGenerationAarch64'
-      PYTHONPATH: 'dlk/python/dlk/scripts/pylib'
+      PYTHONPATH: '.'
```
tsawada: Is this necessary? I think Python might look into current directory with empty PYTHONPATH

blueoil/converter/frontend/base.py
```diff
@@ -28,7 +28,7 @@ def read(self, path: str) -> Graph:
 
         Args:
             path (str): Path to the file to be read
-        
+
```
tsawada: Are these whitespace diffs related to the move? If not I recommend not to include them.

.buildkite/dlk-pipeline.yml
kchygoe: No, It didn't work in the raspberry-pi when I checked. let me work on that in another issue. thx.

blueoil/converter/frontend/base.py
kchygoe: Thx, but please just ignore white spaces in the diff by the review URL `&w=1`. It's too annoying to avoid that manually.

.buildkite/dlk-pipeline.yml
kchygoe: In the dir,
```
$ DEVICE_TEST_INPUT_PATH=./output/TestCodeGenerationAarch64 python3 ./tests/device_tests/test_device.py
  File "./tests/device_tests/test_device.py", line 21, in <module>
    from blueoil.converter.nnlib import NNLib
ModuleNotFoundError: No module named 'blueoil'
```

kchygoe: (however, in python3 interpreter on the same dir, it works...)

docker/Dockerfile
```diff
@@ -125,17 +125,10 @@ FROM blueoil-base AS blueoil-dev
...
-# Install dlk
-WORKDIR /home/blueoil/dlk
-RUN PYTHONPATH=python/dlk python setup.py install
-RUN chmod 777 /home/blueoil/dlk
-
 # Copy blueoil test code to docker image
 COPY tests /home/blueoil/tests
 
-ENV PYTHONPATH $PYTHONPATH:/home/blueoil:/home/blueoil/dlk/python/dlk
+ENV PYTHONPATH $PYTHONPATH:/home/blueoil
```
primenumber: Is it needed?

kchygoe: we don't use anymore! let me remove

kchygoe: removed

kchygoe:
@tsawada please give me a review!

iizukak:
@tsawada We need readability reviewer.

kchygoe:
https://github.com/blue-oil/blueoil/issues/603

PR #851, opened 2020-02-18
Update contents of Object Detection tutorial.

lm-lily:
Add missing lines of configuration setting and change libdlk_arm.so, libdlk_fpga.so, libdlk_x86.so to lib_arm.so, lib_fpga.so and lib_x86.so in the part of "Run inference script on x86 Linux".


tfujiwar:
@iizukak OA is required.

iizukak:
Ownership Approval

PR #853, opened 2020-02-18
Fix some files are missing in output_template

iizukak:

Fix 852
I copied `config.py` from [here](https://github.com/blue-oil/blueoil/blob/v0.19.0/output_template/python/lmnet/utils/config.py)

### DIFF
--- a/output_template/python/config.py
+++ b/output_template/python/config.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 The Blueoil Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


output_template/python/config.py
```diff
@@ -0,0 +1,82 @@
...
+from blueoil.data_processor import Sequence
+
+
+def load_yaml(config_file):
+    with open(config_file) as config_file_stream:
+        config = yaml.load(config_file_stream, Loader=yaml.Loader)
+
+    # use only upper key.
+    keys = [key for key in config.keys() if key.isupper()]
+    config_dict = {key: config[key] for key in keys}
```
tsawada: `return EasyDict({k: v for c.items() if k.isupper()})`

output_template/python/config.py
```diff
@@ -0,0 +1,82 @@
...
+    processors = []
+    if pre_processor_config is None:
+        pre_processor_config = {}
+    for p in pre_processor_config:
+        for class_name in p:
+            class_args = p[class_name]
+            if class_args is None:
+                class_args = {}
+            cls = getattr(module, class_name)
+            # Create none initialized processor `cls` instance.
```
tsawada: this comment doesn't make sense to me. s/none/a new/?

iizukak:
@tsawada In e9794e44d21cc2104c283d6385092f1544c08798, I removed this file accidentally, In this PR, I recovered `config.py`  manually. And it's looks as new file.

iizukak:
Commented issues are fixed.

PR #859, opened 2020-02-19
Use integer division in pre_processor

Joeper214:
Use integer division in pre_processor for consistency
n/a


blueoil/pre_processor.py
```diff
@@ -279,11 +279,11 @@ def joints_to_gaussian_heatmap(joints, image_size,
     for i in range(num_joints):
 
         if joints[i, 2] > 0:
-            center_x = int(joints[i, 0] / stride)
-            center_y = int(joints[i, 1] / stride)
+            center_x = joints[i, 0] // stride
```
Joeper214: @tsawada You're right. lines have float numbers. Like this one. It's okay to cast them to int before division right?

PR #871, opened 2020-02-20
Fix python files of output_template

hadusam:
## What this patch does to fix the issue.
Fix #852 

* Add `output_template/python/blueoil/post_processor.py` as a symbolic link
* Add `blueoil/utils/predict_output/__init__.py` to support python2 (DE10-Nano)
* Replace symlink of `data_processor.py` to the previous file (support python2) to revert a part of #832
  * https://github.com/blue-oil/blueoil/pull/871/commits/f88852d6f9fff54d8b5fc47e012f5f8d6401833c Just copied `data_processor.py` from [Before #832](https://github.com/blue-oil/blueoil/blob/bbe6c540d3ee4f80bed55c2c4150d4845c010d67/output_template/python/lmnet/data_processor.py)
  * https://github.com/blue-oil/blueoil/pull/871/commits/45091d287bc9c8021e24c1d8301698311e94ca00 Updated import path

PR #872, opened 2020-02-20
Add device inference test to e2e tests

### LINKED ISSUE #854: Add tests for scripts in output_template
Because there are no tests for the scripts in `output_template`, they sometimes get broken but we cannot notice that (i.e. https://github.com/blue-oil/blueoil/issues/852).

Adding an inference step at the end of the E2E test is one solution.
https://github.com/blue-oil/blueoil/blob/2d134b72bea9dd3ba418f6bc460c0b8c42c14664/tests/e2e/conftest.py#L60-L67

hadusam:
A part of #854 
Added test on `ubuntu(x86)`, `ubuntu(x86_avx)`, `raspbery-pi(aarch64)`, `de10-nano(arm)`
I also tried to add `de10-nano(fpga)`, but the inference was hanged up. I think it is caused by our test configs so I decided to fix and add test for inference on fpga after this PR.

After #870 and #871


.buildkite/blueoil-pipeline.yml
```diff
@@ -12,33 +12,123 @@ steps:
...
+    env:
+      BUILDKITE_CLEAN_CHECKOUT: 'true'
+  - command: |
+      buildkite-agent artifact download "tmp/*" ./ --build \${BUILDKITE_BUILD_ID}
+      export CWD=\$(pwd)
+      for TEST_CASE in \$(ls tmp/)
+      do
+        echo \${TEST_CASE}
+        cd tmp/\${TEST_CASE}/export/*/*/output/python
+        sudo python run.py -i ../../inference_test_data/raw_image.png -c ../models/meta.yaml -m ../models/lib/libdlk_arm.so
```
tfujiwar: Do we need `sudo` to run with `libdlk_arm.so`? I thought it's needed only for `libdlk_fpga.so`.

hadusam:
@tfujiwar 
I changed my mind to use python for testing instead of shell.
Please review this again! Thanks!

tests/device_tests/test_device_e2e.py
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+    def _get_test_cases(self, input_path, lib_name):
+        return [[case, self._get_param(case, input_path, lib_name)] for case in os.listdir(input_path)]
+
+    def _run(self, python_path, image, model, config):
+        sys.path.append(python_path)
+        from run import run_prediction
+        run_prediction(image, model, config)
+        assert os.path.exists(os.path.join('output', "output.json"))
```
tfujiwar: Please use `self.assertTrue` in unittest.

tests/device_tests/test_device_e2e.py
```diff
@@ -0,0 +1,66 @@
...
+            'image': os.path.join(test_data_dir, "raw_image.png"),
+            'model': os.path.join(lib_dir, lib_name),
+            'config': os.path.join(model_dir, "meta.yaml"),
+            }
+
+    def _get_test_cases(self, input_path, lib_name):
+        return [[case, self._get_param(case, input_path, lib_name)] for case in os.listdir(input_path)]
+
+    def _run(self, python_path, image, model, config):
+        sys.path.append(python_path)
```
tfujiwar: Maybe we need to initialize `sys.path` for every test case to avoid the conflict of the module name `run` among all test cases.
```py
class DeviceE2eTest(unittest.TestCase):
    sys_path_default = copy.deepcopy(sys.path)

    def _run():
        sys.path = copy.deepcopy(sys_path_default) + [python_path]
```

tests/device_tests/test_device_e2e.py
```diff
@@ -0,0 +1,66 @@
...
+class DeviceE2eTest(unittest.TestCase):
+    """Test Case of Device E2E Test."""
+
+    def _get_param(self, test_case, input_path, lib_name):
+        test_case_dir = os.path.join(input_path, test_case)
+        output_dir = glob.glob(os.path.join(test_case_dir, "export/*/*/output"))
+        if output_dir:
+            output_dir = output_dir[0]
+        else:
+            raise Exception("No such directory")
```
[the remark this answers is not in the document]
hadusam: @tfujiwar 
Thanks!
I tried to use it at first but there is not FileNotFoundError of python2. So I use a custom exception. But it is better to use IOError for python2, so I changed it with `if` statement.
Please check it again!

tests/device_tests/test_device_e2e.py
hadusam: Fixed!

tests/device_tests/test_device_e2e.py
hadusam: @tfujiwar 
Nice review! Thanks a lot! It would cause a bug, I didn' t noticed.
I fixed it 👍

PR #886, opened 2020-03-02
Fix LMBiSeNet inference (.elf) test error on x86 and DE10-Nano

oatawa1:
## What this patch does to fix the issue.
- Edit the model by set all channel sizes within hardware restriction at 1024.

* It can pass the test but the result images are not good. (This will solve in other PR.)

## Link to any relevant issues or pull requests.
#830, #249 , #567 

### Learning Rate / Loss
Screen Shot 2020-03-03 at 11 04 57

### Metrics / IOU
Screen Shot 2020-03-03 at 10 59 41

### Images / Overlap Output Input
Screen Shot 2020-03-03 at 11 00 21

###  Run inference test on x86 Linux (Ubuntu 16.04)
```
$ ./models/lib/lm_x86.elf ../inference_test_data/000_images_placeholder\:0.npy ../inference_test_data/708_output\:0.npy

-------------------------------------------------------------
Comparison: Default network test  succeeded!!!
-------------------------------------------------------------
TotalInitTime 35231,  sum:35.231ms
TotalRunTime 1.07207e+06,  sum:1072.07ms
..Convolution 18247,1377,  sum:19.624ms
....kn2row-1x1 18244,1377,  sum:19.621ms
......matrix_multiplication 18242,1375,  sum:19.617ms
..BatchNorm 8966,277,  sum:9.243ms
..QTZ_linear_mid_tread_half 31401,  sum:31.401ms
....pack_input 9634,  sum:9.634ms
..ExtractImagePatches 132,23,17,2,161,25,19,  sum:0.379ms
..QuantizedConv2D 119548,30057,4538,9685,5468,9673,6464,9679,25243,4175,4246,4458,4256,4955,4260,5488,2044,3763,1007,4254,1443,4257,1859,4259,9242,2344,2099,270997,200955,242116,8149,  sum:1010.98ms
....Convert Tensor 131,30,13,5,13,5,14,5,10,5,2,4,2,4,2,5,3,1,1,0,1,0,1,0,2,2,3,200,45,23,21,  sum:0.553ms
....quantized-kn2row 119415,30025,4524,9678,5454,9666,6449,9673,25232,4170,4243,4453,4253,4950,4257,5482,2039,3761,1005,4253,1441,4256,1857,4258,9239,2341,2095,270797,200909,242092,8126,  sum:1010.39ms
[...]

Issue #830, opened 2020-02-12
Verify LMBiSeNet runs on DE10-Nano


iizukak:
We are not sure that `LMBiSeNetQuantize` can run on DE10-Nano or not.
https://github.com/blue-oil/blueoil/blob/master/blueoil/networks/segmentation/lm_bisenet.py

iizukak:
Maybe we have channel size problem.
https://github.com/blue-oil/blueoil/issues/249

PR #889, opened 2020-03-03
Fix output image results in Semantic Segmentation task from run.py

oatawa1:
Resolve the broken output image in Semantic Segmentation task from `run.py`

* Before 
0006R0_f02040_x86
* After
0006R0_f02040

#886 #830

### DIFF
--- a/blueoil/utils/predict_output/output.py
+++ b/blueoil/utils/predict_output/output.py
@@ -118,18 +118,17 @@ class JsonOutput():
             }
 
             prediction = []
+            label = np.argmax(output, axis=2)
             for i, class_name in enumerate(self.classes):
-                img = output[:, :, i] * 255
-                img = PIL.Image.fromarray(img, mode="F")
-                img = img.convert("L")
+                img = np.zeros(label.shape[:2], dtype=np.uint8)
+                img[label == i] = 255
+                img = PIL.Image.fromarray(img, mode="L").convert('1')
                 img = img.resize((raw_image.shape[1], raw_image.shape[0]))
-
                 # base64 encode
                 buffered = BytesIO()
                 img.save(buffered, format="PNG")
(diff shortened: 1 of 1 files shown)


blueoil/utils/predict_output/output.py
```diff
@@ -118,18 +118,17 @@ def _semantic_segmentation(self, outputs, raw_images, image_files):
...
 
             prediction = []
+            label = np.argmax(output, axis=2)
             for i, class_name in enumerate(self.classes):
-                img = output[:, :, i] * 255
-                img = PIL.Image.fromarray(img, mode="F")
-                img = img.convert("L")
+                img = np.zeros(label.shape[:2], dtype=np.uint8)
+                img[label == i] = 255
+                img = PIL.Image.fromarray(img, mode="L").convert('1')
```
iizukak: @oatawa1 
Sorry, I don't understand why this diff solve the problem
Can you teach me?

blueoil/utils/predict_output/output.py
```diff
@@ -118,18 +118,17 @@ def _semantic_segmentation(self, outputs, raw_images, image_files):
...
                 img = img.resize((raw_image.shape[1], raw_image.shape[0]))
-
```
iizukak: Please not include these format refactoring in one PR.

blueoil/utils/predict_output/output.py
oatawa1: I committed again to make it similar to the previous commit.

The new commit assigns each pixel color by the predicted class (the highest probability from softmax), Thus, it can be seen as assign 255 to the pixel that belongs to that class.
$ \sum_{\for i=class_no}{max(p_i) * 255} $

While each pixel in the old commit is a combination of each class based on softmax probability.
$ \sum_{\for i=class_no}{p_i * 255} and sum(p_i) = 1.0 $

PR #890, opened 2020-03-03
Fix type annotation

primenumber:

Fix type annotation.


tkng:
Ownership Approval
Readability Approval

Issue #901, opened 2020-03-06
CI is unstable: it sometimes fails during docker build


tfujiwar:
https://buildkite.com/blueoil/dlk-test/builds/2062#f7781e88-35c3-4c5f-bdca-2d01cf7dbc8e
```
[2020-02-25T03:35:01Z] Step 34/36 : COPY docker/docker-entrypoint.sh /
[2020-02-25T03:35:05Z] failed to export image: failed to set parent sha256:c79a9178e77e00a16596779f4417a424d8162eb764953a26752be9026d31d7c7: unknown parent image ID sha256:c79a9178e77e00a16596779f4417a424d8162eb764953a26752be9026d31d7c7
[2020-02-25T03:35:05Z] Makefile:15: recipe for target 'build' failed
[2020-02-25T03:35:05Z] make: *** [build] Error 1
[2020-02-25T03:35:05Z] 🚨 Error: The command exited with status 2
```

Building the same docker image at the same time might be a cause of this error.

tfujiwar:
A different but similar error occurred during docker build.

https://buildkite.com/blueoil/dlk-test/builds/2258#5f89a5a4-4c4a-47d4-b4d7-dd148a6fa52e
```
[2020-03-16T00:44:38Z] Successfully installed blueoil cffi-1.14.0 cloudpickle-1.3.0 horovod-0.18.1 mpi4py-3.0.2 pycparser-2.20
[2020-03-16T00:45:03Z] unknown parent image ID sha256:ffc2bd785e6a71b19963d5166e698eea90db02f440bfbd5dcba5fc4d23134162
[2020-03-16T00:45:03Z] Makefile:15: recipe for target 'build' failed
[2020-03-16T00:45:03Z] make: *** [build] Error 1
[2020-03-16T00:45:03Z] 🚨 Error: The command exited with status 2
```

tfujiwar:
Updated a script to clean up docker images in CI servers.

PR #905, opened 2020-03-06
Add docker build step at the beginning of all CI pipeline

tfujiwar:
In CI, Building the same docker image at the same time sometimes causes an error. So I added a common "docker build" step at the beginning of all CI pipelines. It avoids CI pipelines to build the same docker image many times.

Issue #901 .


iizukak:
@tfujiwar 
OK Thanks. We want stable CI strongly.

Please teach me Fujiwara-sense,
Why this diff can avoid simultaneous build of Docker image?
Before this diff, 

```
command: "make test-classification"
command: "make test-dlk-main"
command: "make test-unit-main"
```

build docker image as their dependency, In you PR, call `make build` explicitly, I'm not sure why this can avoid simultaneous build.

tfujiwar:
@iizukak Sorry for lacking the information.

In `blueoil-pipeline.yml` for example, without this PR, 4 main steps run in parallel. It may cause an error.
```
┬-> make test-classification
├-> make test-object-detection
├-> make test-semantic-segmentation
└-> make test-keypoint-detection
```

With this PR, I added a `docker build` step and `wait` step at the beginning. So the following main 4 steps cannot start before `docker build` step finishes.
https://buildkite.com/docs/pipelines/wait-step
```
docker build ┬-> make test-classification
             ├-> make test-object-detection
             ├-> make test-semantic-segmentation
             └-> make test-keypoint-detection
```

Adding `docker build` step and `wait` step is enough for avoiding parallel builds in one pipeline but we need to think about avoiding parallel builds among all pipelines. So I specified `concurrency` and `concurrency_group` to the docker build step. With these settings, we can limit the parallel number of jobs that belong to the same `concurrency_group` on the organization level.
https://buildkite.com/docs/pipelines/controlling-concurrency
```
concurrency: 1
concurrency_group: "docker/gpu/${BUILDKITE_COMMIT}"
```

iizukak:
Ownership Approval

PR #907, opened 2020-03-10
Remove PYTHONPATH

primenumber:
`PYTHONPATH` for docker-compose is no longer needed.


iizukak:
OA

PR #912, opened 2020-03-12
Fix comparison order in generationg thresholds

primenumber:

threshold_table[c, -1] should be 1 if threshold_table[c, :-1] is __monotonically non-decreasing__, not __strictly increasing__.
This PR fix it.

### DIFF
--- a/blueoil/converter/core/optimizer.py
+++ b/blueoil/converter/core/optimizer.py
@@ -308,7 +308,7 @@ def pass_compute_thresholds(graph: Graph) -> None:
 
         for c in range(ch):
             threshold_table[c, -1] = 1 \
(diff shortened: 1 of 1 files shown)


primenumber:
Umm, we should check `(scaling_factor[ch_id] < 0) ^ (ch_id in bn_nega_idx)`, since threshold_table was croped into [0, max_th_value] already, and ceiled/floored, so we cannot detect thresholds are increasing order or decreasing order.

primenumber:
And generally, we should check increasing or decreasing by more generic strategy...

primenumber:
Fixed algorithm, please re-review @iizukak

Issue #924, opened 2020-03-13
The flag of FUNC_TIME_MEASUREMENT should be set in compiling all ELF files


tk26eng:
Now `-DFUNC_TIME_MEASUREMENT` is set only for compiling `lm_fpga.elf` in `Makefile`.
But I think it's better for us to set the flag to all ELF files in generating them.
Actually time measurement results are very useful in development.
Most developer wants the information in debug.
We get little longer logs from those ELF files if the flag is enabled, but it's no problem because those ELF files are only for debug.

https://github.com/blue-oil/blueoil/blob/65e60638fd87c2e2da864eda0d3ec14e06f7142e/blueoil/converter/templates/Makefile#L118-L136

What do you think about this ?

PR #928, opened 2020-03-17
Add time measurement flag to all elf binaries in Makefile

tk26eng:
This patch makes code simpler.

#924

### DIFF
--- a/blueoil/converter/templates/Makefile
+++ b/blueoil/converter/templates/Makefile
@@ -116,19 +116,19 @@ clean:
 	-$(RM) $(OBJ)
 
 lm_x86:           CXX = g++
-lm_x86:           FLAGS += $(INCLUDES) -O3 -std=c++14 -DUSE_PNG -pthread -g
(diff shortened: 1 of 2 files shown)


blueoil/converter/templates/Makefile
```diff
@@ -116,19 +116,19 @@ clean:
 	-$(RM) $(OBJ)
 
 lm_x86:           CXX = g++
-lm_x86:           FLAGS += $(INCLUDES) -O3 -std=c++14 -DUSE_PNG -pthread -g
+lm_x86:           FLAGS += $(INCLUDES) -O3 -std=c++14 -DUSE_PNG -pthread -g -DFUNC_TIME_MEASUREMENT
```
tsawada: maybe we should place all `-D` flags next to each other.
`-DUSE_PNG -DFUNC_TIME_MEASUREMENT -pthread -g`

tk26eng: Maybe it's better to place `-D` like you pointed out.
That will be modified in another PR.

tk26eng:
Fix #924

PR #930, opened 2020-03-17
Fix run.py time measurement log output

kchygoe:
which has been devided twice.

log output about time measurement is wrong


kchygoe:
ya, it would be better to change function name and variables later

PR #936, opened 2020-03-19
Delete cmake dependency

### LINKED ISSUE #436: Unify dlk build system
Currently, generated dlk project has CMakeLists.txt and normal Makefile, we are maintaining both files.
We should unify these (maybe remove Makefile.tpl).

### LINKED ISSUE #542: Choose make or cmake to maintaine for DLK
Currently, [blueoil convert](https://github.com/blue-oil/blueoil/blob/master/blueoil/cmd/convert.py#L73) run `make` command with `Makefile` created by generate_project.py.
DLK test([test_code_generation.py](https://github.com/blue-oil/blueoil/blob/master/dlk/tests/test_code_generation.py#L364-L380)) run `make` command with `Makefile` overwritten by `cmake` command.
To get library files, there are two methods by using `cmake` or not. And one(directly make) is not tested(maintained) and another(cmake -> make) is tested but not used by `blueoil convert` interface.

I think we should choose one method to maintain. Which one is used well by developers?🤔
If `cmake` is selected, we should change `blueoil convert` to use `cmake` and it is better to remove [Makefile.tpl](https://github.com/blue-oil/blueoil/blob/master/dlk/python/dlk/templates/Makefile.tpl) to avoid confusing.
If make is selected, change dlk test to use make directly and try to remove cmake settings to avoid confusing.

tk26eng:
This PR removes dependency for `cmake`.
Test should be modified to use `make` instead of `cmake`.

Fix #436 
Fix #542


tests/converter/test_code_generation.py
```diff
@@ -336,9 +336,18 @@ def codegen_cpu(self,
...
-        lib_name = 'libdlk_' + cpu_name
+        cpu_to_lib = {
+            'aarch64_False':'aarch64', 
+            'arm_False':'arm',
+            'arm_fpga_False':'fpga', 
+            'x86_64_False':'x86',
+            'x86_64_True':'x86_avx'
+        }
+
+        lib_base_name = cpu_to_lib[cpu_name + '_' + str(use_avx)]
```
iizukak: AVX is for x86 only.
Added True and False flag to all CPU architecture looks redundant.

iizukak:
Please set readability reviewer after ownership approval.

tests/converter/test_code_generation.py
tk26eng: Yes, I know AVX flag is redundant for other CPU than x86.
But I used AVX flag for all CPUs to apply `cpu_to_lib`  table to make the code simple.

Is `if statement` better than translation table ?

iizukak: > But I used AVX flag for all CPUs to apply cpu_to_lib table to make the code simple.

Sorry, If my understand of your code is correct, I do not agree with you...
Create new table to just add `_avx` looks not suitable way.

How about to write like

```
lib_base_name = "x86_avx" if use_avx else cpu_name
```

tk26eng: I wanted to remove `if statement` but I gave it up .
I changed to use `if statement` instead of table.
Is it OK for you ?

tests/converter/test_code_generation.py
```diff
@@ -355,40 +364,21 @@ def codegen_cpu(self,
...
-        run_and_check(['make', 'VERBOSE=1', 'lib', '-j8'],
+        run_and_check(['make', 'lib_' + lib_base_name, '-j8'],
                       project_dir,
                       join(output_path, "make.out"),
                       join(output_path, "make.err"),
                       self,
-                      check_stdout_include=['Building'],
                       check_stderr_block=['error: ']
                       )
+
```
iizukak: @tk26eng 
Sorry for little thing.
We do not change code style fix with other implementation.

tests/converter/test_code_generation.py
```diff
@@ -336,9 +336,18 @@ def codegen_cpu(self,
                cache_dma=cache_dma,
                )
 
-        lib_name = 'libdlk_' + cpu_name
+        if cpu_name == 'arm_fpga':
```
iizukak: Thanks. Looks good for me.

tk26eng:
ping @iizukak

tests/converter/test_code_generation.py
```diff
@@ -336,9 +336,18 @@ def codegen_cpu(self,
...
+            lib_base_name = 'fpga'
+        elif cpu_name == 'x86_64':
+            if use_avx == False:
+                lib_base_name = 'x86'
+            else:
+                lib_base_name = 'x86_avx'
```
tfujiwar: This would be more natural.
```py
if use_avx:
    lib_base_name = 'x86_avx'
else:
    lib_base_name = 'x86'
```

tk26eng: I fixed it.

PR #938, opened 2020-03-19
Fix checkpoint loader to use GFile

kchygoe:
When `OUTPUT_DIR=gs://xxxx/xxx`,
Training cannot be kept running.
This is critical for gcs/s3 users

### DIFF
--- a/blueoil/cmd/train.py
+++ b/blueoil/cmd/train.py
@@ -395,8 +395,8 @@ def train(config_file, experiment_id=None, recreate=False):
     if not tf.io.gfile.exists(checkpoint):
         raise Exception('Checkpoints are not created in {}'.format(experiment_dir))
 
-    with open(checkpoint) as stream:
-        data = yaml.load(stream)
+    with tf.io.gfile.GFile(checkpoint) as stream:
+        data = yaml.load(stream, Loader=yaml.FullLoader)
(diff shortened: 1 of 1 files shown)


blueoil/cmd/train.py
```diff
@@ -395,8 +395,8 @@ def train(config_file, experiment_id=None, recreate=False):
     if not tf.io.gfile.exists(checkpoint):
         raise Exception('Checkpoints are not created in {}'.format(experiment_dir))
 
-    with open(checkpoint) as stream:
-        data = yaml.load(stream)
+    with tf.io.gfile.GFile(checkpoint) as stream:
+        data = yaml.load(stream, Loader=yaml.FullLoader)
```
tsawada: is this loader change related to GFile change?

It seems there are a few other use of `yaml.load` within blueoil thus maybe we should fix all in another diff.

kchygoe: no, this is not directly related.

I just noticed that `load` method has been changed to specify `Loader` . 
Let me dedicate them as you commented.
https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

kchygoe: fixed

Issue #481, opened 2019-10-02
refactor train.py


tkng:
The function `start_training` in `train.py` is too long, it should be decomposed into some functions.

It might be better to add a new class `Trainer`?

iizukak:
@tkng 
Can you check #693 ?
The PR is that this issue want?

tkng:
@iizukak thanks, I'll take a look.

tkng:
I think #693 is a good starting point.

iizukak:
OK Thanks.

PR #948, opened 2020-03-25
Remove a dataset subset type "train_validation_saving"

### LINKED ISSUE #946: Can we remove a dataset subset type "train_validation_saving"?
A dataset subset type `train_validation_saving` makes `train.py` complicated. There are 6 if-statements with `use_train_validation_saving` in `train.py` and one of them has more than 40 lines.
https://github.com/blue-oil/blueoil/blob/969633c20aeea7ee964b2742efd5dfae331e6e80/blueoil/cmd/train.py#L256-L301

If we can remove this subset type, we can make `train.py` more readable (#481). Currently, this subset type is used only in cifar10 and cifar100.

tfujiwar:
This PR aims to refactor train.py (#481). The detail is written in #946 .
If this feature is still used and needs to be maintained, it's OK to discard this PR.

### DIFF
--- a/blueoil/cmd/train.py
+++ b/blueoil/cmd/train.py
@@ -72,20 +72,9 @@ def start_training(config):
 
     ModelClass = config.NETWORK_CLASS
     network_kwargs = {key.lower(): val for key, val in config.NETWORK.items()}
-    if "train_validation_saving_size".upper() in config.DATASET.keys():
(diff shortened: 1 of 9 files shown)


iizukak:
Thanks.
In my opinion, `use_train_validation_saving` is not useful anymore.

Issue #692, opened 2019-12-19
Add unit tests for `profile_model.py`


iizukak:
Currently, we have test file for `profile_model.py` but there are no unit tests.
https://github.com/blue-oil/blueoil/blob/master/lmnet/tests/executor_tests/test_profile.py

suttang:
File moved: https://github.com/blue-oil/blueoil/blob/master/tests/unit/executor_tests/test_profile.py

PR #952, opened 2020-03-27
Add _save_json unit test for profile model

Joeper214:
This PR creates a unit test for `_save_json` method in profile_model.py to increase test code coverage.
related to #692

### DIFF
--- a/tests/unit/executor_tests/test_profile.py
+++ b/tests/unit/executor_tests/test_profile.py
@@ -64,7 +64,9 @@ def test_save_json():
         node_flops_dict=test_dict["flops"]
     )
     output_file = os.path.join(environment.EXPERIMENT_DIR, "{}_profile.json".format(test_dict["model_name"]))
-    file_data = json.loads(open(output_file, 'r').read())
+    with open(output_file, 'r') as fp:
+        file_data = json.load(fp)
+
     assert os.path.isfile(output_file)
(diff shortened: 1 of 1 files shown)


tfujiwar:
@Joeper214 Could you request a readability review after OA?

tests/unit/executor_tests/test_profile.py
```diff
@@ -35,6 +41,34 @@ def test_profile():
...
+
+    _save_json(
+        name=test_dict["model_name"],
+        image_size=(test_dict["image_size_height"], test_dict["image_size_width"]),
+        num_classes=test_dict["num_classes"],
+        node_param_dict=test_dict["parameters"],
+        node_flops_dict=test_dict["flops"]
+    )
+    output_file = os.path.join(environment.EXPERIMENT_DIR, "{}_profile.json".format(test_dict["model_name"]))
+    file_data = json.loads(open(output_file, 'r').read())
```
tfujiwar: You need to close the file or use `with` block (Sorry, I forgot to point this out... 🙇).

And you can avoid `read()` by using `json.load()` instead of `json.loads()`. It will consume less memory (in this case, it's not a big problem because JSON file is not so big but it's better in general).
```py
with open(output_file, 'r') as fp:
  file_data = json.load(fp)
```

PR #962, opened 2020-04-01
Fix incorrect channel_data_format on Conv2D from `channels_first/channels_last` to `NCHW/NHWC`

oatawa1:
Fix incorrect channel_data_format on Conv2D 

From TF api docs (https://www.tensorflow.org/api_docs/python/tf/nn/conv2d)
conv2d use `NCHW/NHWC` not `channels_first/channels_last`.

### DIFF
--- a/blueoil/networks/classification/darknet.py
+++ b/blueoil/networks/classification/darknet.py
@@ -243,7 +243,7 @@ class Darknet(Base):
         self.conv_19 = conv2d(
             "conv_19", self.block_18, filters=self.num_classes, kernel_size=1,
             activation=None, use_bias=True, is_debug=self.is_debug,
-            kernel_initializer=kernel_initializer, data_format=channel_data_format,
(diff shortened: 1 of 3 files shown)


iizukak:
TensorFlow can run with invalid argument?

oatawa1:
@iizukak 
I'm sorry, I checked the incorrect version API document. I will close this issue.

For TF 1.15 
`data_format`: A string, one of `channels_last` (default) or `channels_first`. The ordering of the dimensions in the inputs. `channels_last` corresponds to inputs with shape `(batch, height, width, channels)` while `channels_first` corresponds to inputs with shape `(batch, channels, height, width)`.
(https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/layers/Conv2D)

About this
> TensorFlow can run with invalid argument?
No it cannot run with invalid argument.

PR #976, opened 2020-04-13
Fix Blueoil init in the case of no validation dataset

### LINKED ISSUE #964: Object Detection tutorial command failed when validation dataset path is empty.
When validation dataset is empty.
Under the error occured.

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/blueoil/dataset/annotations-bbox.csv'
```

oatawa1:
Previous code set the validation dataset directory to `''`, thus when the configuration file generate  `validation_extend_dir` = `''`. So the blueoil train recognize it has validation dataset in the default directory. (if dir is the **relative path** it refers to `default path` **not** `current path`.)

 This commit changes the validation dataset directory to `None`. So it will not generate `validation_extend_dir` in the configuration file.

Closes #964


oatawa1:
@iizukak Thank you.

oatawa1:
@tfujiwar Thank you.

PR #982, opened 2020-04-13
Change condition to show an warning from generate_project.py

### LINKED ISSUE #971: UserWarning for LmSegnetV1Quantize
I got UserWarning when convert `LmSegnetV1Quantize`.
Please update the network to fix this issue. 

```
finish
import pb file
optimize graph step: start
optimize graph step: done!
generate code step: start
generate code step: done!
Output files are generated in /home/blueoil/saved/segmentation_20200408012343/export/save.ckpt-917/360x480/output
Please see /home/blueoil/saved/segmentation_20200408012343/export/save.ckpt-917/360x480/output/README.md to run prediction
/home/blueoil/blueoil/converter/core/operators.py:60: UserWarning: WRN Output channels need to be multiple of 32 for space2depth1 of SpaceToDepth, but got output channel size of 12
  self._check_consistency()
```

tk26eng:
The number of output channel is checked if SpaceToDepth operator exists when `generate_project.py` runs.
The number of output channel for SpaceToDepth should be multiple of 32 in case of quantized situation.
But for floating point (not quantized) operation, the number of output channel doesn't have to be multiple of 32.

This PR removes the wrong checking of output channel for floating point operation.

Fix #971


blueoil/converter/core/operators.py
```diff
@@ -814,7 +814,8 @@ def _check_consistency(self) -> None:
             2. (kernel_size^2 * {8, 16}).
         """
         super()._check_consistency()
-        if self.channel % 32 != 0:
+        if self.input_ops['input'].op_type == 'LinearMidTreadHalfQuantizer' and \
```
iizukak, line 817: @tk26eng 
Thanks!

If characters less than 120, it's better to avoid `\`.
And please fix the documentation about `space_to_depth`.
https://docs.blueoil.org/converter/supported_ops.html#

tk26eng: @iizukak 
Those have been fixed.
Please check them again.

tk26eng:
@tfujiwar 
We should apply this patch also to it.
Thank you for your suggestion.

Issue #953, opened 2020-03-27
Document generation process contains error and warning


oatawa1:
When generate documents using `$ make html` 
Error and many warnings about a module not found occurred.

Example of warning and error.
`autodoc: failed to import many modules`

```
WARNING: autodoc: failed to import class 'Optimizer' from module 'blueoil.converter.core.optimizer'; the following exception was raised:
Traceback (most recent call last):
  File "/home/wasin/.venv/lib/python3.6/site-packages/sphinx/util/inspect.py", line 251, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'blueoil.converter.core.optimizer' has no attribute 'Optimizer'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/wasin/.venv/lib/python3.6/site-packages/sphinx/ext/autodoc/importer.py", line 71, in import_object
    obj = attrgetter(obj, attrname)
  File "/home/wasin/.venv/lib/python3.6/site-packages/sphinx/ext/autodoc/__init__.py", line 226, in get_attr
    return autodoc_attrgetter(self.env.app, obj, name, *defargs)
  File "/home/wasin/.venv/lib/python3.6/site-packages/sphinx/ext/autodoc/__init__.py", line 1575, in autodoc_attrgetter
    return safe_getattr(obj, name, *defargs)
  File "/home/wasin/.venv/lib/python3.6/site-packages/sphinx/util/inspect.py", line 267, in safe_getattr
    raise AttributeError(name)
AttributeError: Optimizer
```

The full log is here.
https://gist.github.com/oatawa1/8574d2ceb1e15d95c813eb7ff9034fdc

- [x] Fix Error #989 
- [x] Refactor API doc directory #995

PR #989, opened 2020-04-14
Remove error from make html doc and update converter doc using Sphinx-apidoc

oatawa1:
Remove error when generate doc using `make html` which mainly from reference error.
- For converter update using autogenerate from `Sphinx-apidoc`.
- For `optimization.rst`, I fix it but I think this file is obsolete. **Shall we remove it?**: removed

#953 (Error part) 
The warning part will be solved in another PR to reduce the amount of commit code.


iizukak:
@oatawa1 Thanks. We can remove `optimization.rst,`

oatawa1:
@iizukak Thanks. I removed it in the latest commit

oatawa1:
The deployment version is different from the local test version. I will fix it

oatawa1:
@iizukak Thanks.

Can I merge this PR?

The current problem is the current deployment version show only the class name it should show both class name along with module and arg name.
https://deploy-preview-989--blueoil.netlify.com/converter/converter.core.html

In my opinion, I want to merge this one to fix the error problem.
Then, this Python API will be removed in next PR, because I will migrate the Python API doc to autogenerate.

oatawa1:
The problem of deployment is fixed in #995 . So I merged this PR.

PR #994, opened 2020-04-15
Add __init__.py to the missing locations in blueoil/.

oatawa1:
## What this patch does to fix the issue.
The __init__.py was missing in some folder, so it cannot locate package hierarchy.

Add `__init__.py` to five directories in `blueoil/` as list:
- `blueoil/networks/keypoint_detection/`
- `blueoil/networks/object_detection/`
- `blueoil/metrics/`
- `blueoil/converter/utils/`
- `blueoil/utils/tfds_builder/`

## Link to any relevant issues or pull requests.
Closes #993

PR #995, opened 2020-04-15
Create auto generate Python API document for `blueoil/` directory.

### LINKED ISSUE #990: Need Python API document for whole blueoil project.
The current Python API document covers only the converter part.

This tool can be used to generate python API. https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html

oatawa1:
Can merge after #989 and #994 

Create Python API document for whole `blueoil/` directory using **Sphinx AutoAPI**
https://sphinx-autoapi.readthedocs.io/en/latest/index.html (`sphinx-autoapi` dependency)

Excluded location are `blueoil/templates` and `blueoil/config`.

Closes #990 

## Todo
- [x] #989 Merged
- [x] #994 Merged
- [x] Remove Converter Python API from docs/converter directory because it's already generated now. (Do this after #989 merged and rebased to avoid conflict.)


oatawa1:
@iizukak Thanks.

docs/conf.py
```diff
@@ -195,6 +196,16 @@
...
 
 
+# -- Options for Sphinx AutoAPI output ----------------------------------------------
+
+# Paths (relative or absolute) to the source code that you wish
+# to generate your API documentation from.
+autoapi_dirs = ['../blueoil/']
+
+# A list of patterns to ignore when finding files.
+autoapi_ignore = ['*migrations*', '*/configs/*', '*/templates/*']
```
tfujiwar, line 206: Can we specify more concrete paths for `configs` and `templates`? `blueoil/configs/*` or something.

oatawa1: I think the Sphinx-autoapi ignore format is an absolute path. Thus I cannot ignore the `*` prefix.
I add `/blueoil/` to them to `*/blueoil/configs/*`, `*/blueoil/templates/*`.

https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html#confval-autoapi_ignore

tfujiwar: Thanks for checking. I guess we need to include the root dir like `../blueoil/configs/*`. If this doesn't work, it's fine as it is 👍 

https://github.com/readthedocs/sphinx-autoapi/blob/9397c0709f086f9a656bdc5bc535134b4084349b/autoapi/mappers/base.py#L225

oatawa1: I tried `../blueoil/configs/*` before, it does not work, unfortunately.

tfujiwar: OK! Thanks.

PR #1014, opened 2020-05-01
Add specification to blueoil document index.

oatawa1:
The current specification document is not included in document index. Thus, I make them accessible from index page.


oatawa1:
@iizukak thanks.

PR #1016, opened 2020-05-01
Move support ops to the end of converter document part

oatawa1:
The support ops document appears in between the converter document.
This commit moves support ops to the end of the converter document part.


oatawa1:
@iizukak thanks.

PR #1025, opened 2020-05-07
Remove deprecated properties from dataset class

yd8534976:
Remove some deprecated properties from dataset class. These feature will be done in https://github.com/blue-oil/blueoil/blob/e57843009854c3b74737a841b3d6b850dab23dae/blueoil/datasets/dataset_iterator.py#L287-L363 .

### DIFF
--- a/blueoil/datasets/delta_mark.py
+++ b/blueoil/datasets/delta_mark.py
@@ -23,7 +23,7 @@ import pandas as pd
 from blueoil import data_processor
 from blueoil.datasets.base import Base, ObjectDetectionBase, StoragePathCustomizable
 from blueoil.utils.image import load_image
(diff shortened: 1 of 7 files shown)


blueoil/datasets/delta_mark.py
```diff
@@ -82,32 +81,11 @@ def __init__(
...
-                self._indices = shuffle(range(self.num_per_epoch), seed=self.seed)
-            else:
-                self._indices = list(range(self.num_per_epoch))
-
-        return self._indices
-
-    def _get_index(self, counter):
-        return self.indices[counter]
-
-    def _shuffle(self):
```
tfujiwar, line 105: This method is called in `dataset_iterator.py` but I think it's not needed. Could you refactor it also?

Issue #702, opened 2019-12-24
Optimize Add implementation


tkng:
According to the performance measurement of LmResnet for ImageNet on FPGA, there is some low hanging fruits, `Add` operator is now consuming over 20ms.

primenumber:
@tkng Try #1027 !!

PR #1027, opened 2020-05-11
Optimize binary operation for converter runtime

primenumber:

This PR improve performance for binary operator, if the shape of inner dimension is same between left hand side and right hand side.

#702


iizukak (approved):
OA

PR #1034, opened 2020-05-15
Change colormap scope

kchygoe:
Change COLOR_MAP scope


blueoil/common.py
```diff
@@ -26,34 +26,33 @@ class Tasks(Enum):
...
-    (0,   0,   192),  # COLOR18
-    (192, 128, 128)   # COLOR19
-]
-
-
 def get_color_map(length):
+    # Color Palette for General Purpose
+    # Sample image is here
+    # https://github.com/blue-oil/blueoil/tree/master/docs/_static/color_map.png
+    COLOR_MAP = [
```
tfujiwar, line 33: Could you use snake_case for local variables?

kchygoe: got it

PR #1045, opened 2020-05-22
Missing outputs of input placeholder

nlpng:
This PR is to deal with cases that the input placeholder has two or more shared branches, as the branches maybe removed during the optimization pass.


blueoil/converter/core/optimizer.py
```diff
@@ -634,8 +634,14 @@ def pass_lookup(graph: Graph) -> None:
                     {'input': placeholder[0], 'lsb': pe_lsb, 'msb': pe_msb}, dimension_format='ChHWBCl')
 
         get_nodes_in_branch(quantizer, placeholder[0], to_be_removed)
+
+        reserved_placeholder_ops = []
+        for out_op in placeholder[0].output_op_list:
+            if out_op not in to_be_removed:
+                reserved_placeholder_ops.append(out_op)
```
tfujiwar, line 641: Usign list comprehension is better for readability and efficiency.
```py
reserved_placeholder_ops = [
    out_op
    for out_op in placeholder[0].output_op_list
    if out_op not in to_be_removed
]
```
or
```py
reserved_placeholder_ops = [
    out_op for out_op in placeholder[0].output_op_list
    if out_op not in to_be_removed
]
```

blueoil/converter/core/optimizer.py
```diff
@@ -634,8 +634,14 @@ def pass_lookup(graph: Graph) -> None:
...
         get_nodes_in_branch(quantizer, placeholder[0], to_be_removed)
+
+        reserved_placeholder_ops = [
+            out_op for out_op in placeholder[0].output_op_list
+            if out_op not in to_be_removed
+        ]
         placeholder[0].remove_output('output')
-        placeholder[0].add_output('output', pe)
+        for add_to_placeholder in reserved_placeholder_ops:
+            placeholder[0].add_output('output', add_to_placeholder)
```
tfujiwar, line 644: Oh sorry, my previous review was bad...
Those two loops are not needed to be separated. Could you make them into one?

nlpng: Thanks, I fixed in commit db82531, not sure it is the fix you are looking for, but now it is only one for loop.

tfujiwar: Thank you.

Issue #1058, opened 2020-05-27
fix python code according to the "Python reviewers guide"


lm-noshiro:
ref: https://leapmind.kibe.la/notes/1958

lm-noshiro:
I'll do this!

PR #1059, opened 2020-05-27
refactor if structure

lm-noshiro:

* Refactored `if` statement structures of several Python scripts according to [Python Reviewers Guide](https://leapmind.kibe.la/notes/1958).
    * Replaced unnecessary use of `elif` with `if` if it's immediately after `return` / `raise` statement.
    * Changed to use *early return pattern*

#1058

### DIFF
--- a/blueoil/cmd/init.py
+++ b/blueoil/cmd/init.py
@@ -139,22 +139,22 @@ learning_rate_schedule_map = OrderedDict([
 def network_name_choices(task_type):
     if task_type == 'classification':
         return [definition['name'] for definition in classification_network_definitions]
-    elif task_type == 'object_detection':
+    if task_type == 'object_detection':
         return [definition['name'] for definition in object_detection_network_definitions]
-    elif task_type == 'semantic_segmentation':
+    if task_type == 'semantic_segmentation':
         return [definition['name'] for definition in semantic_segmentation_network_definitions]
-    elif task_type == 'keypoint_detection':
+    if task_type == 'keypoint_detection':
         return [definition['name'] for definition in keypoint_detection_network_definitions]
 
 
 def dataset_format_choices(task_type):
     if task_type == 'classification':
         return [definition['name'] for definition in classification_dataset_formats]
-    elif task_type == 'object_detection':
+    if task_type == 'object_detection':
         return [definition['name'] for definition in object_detection_dataset_formats]
-    elif task_type == 'semantic_segmentation':
+    if task_type == 'semantic_segmentation':
         return [definition['name'] for definition in semantic_segmentation_dataset_formats]
-    elif task_type == 'keypoint_detection':
+    if task_type == 'keypoint_detection':
         return [definition['name'] for definition in keypoint_detection_dataset_formats]
 
 
--- a/blueoil/converter/plugins/tf.py
+++ b/blueoil/converter/plugins/tf.py
@@ -485,40 +485,39 @@ class Importer(object):
             node_rank = len(input_node.get_shape())
             return out_format if node_rank == len(out_format) else rank_to_format[node_rank]
 
-        if isinstance(node, Node):
-            if node.get_format() is None:
-                out_format = output_format or guess_node_format(node)
-            else:
-                out_format = node.get_format()
-
-            op_type = self.convert_operator(node.op_type)
-            if op_type == 'Conv':
-                return out_format, [out_format, _default_w_format, 'C']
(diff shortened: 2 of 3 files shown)


blueoil/cmd/init.py
```diff
@@ -139,22 +139,22 @@
 def network_name_choices(task_type):
     if task_type == 'classification':
         return [definition['name'] for definition in classification_network_definitions]
-    elif task_type == 'object_detection':
+    if task_type == 'object_detection':
```
iizukak, line 142: I'm not sure this change is what @tsawada san suggested.
`elif` looks no problem in my opinion...

tsawada:
```python
defs = {
  'classification': classification_network_definitions,
  'object_detection': object_detection_network_definitions,
  ...
}
if task_type not in defs:
  raise ValueError
return [... for ... in defs[task_type]
```

tools/debug.py
```diff
@@ -128,57 +128,56 @@ def derive_full_bit_qwords(self, n):
...
+        for hi in range(0, self.row):
+            in_col_offset = 0
+            out_col_offset = 0
+
+            for c in range(0, full_bit_col):
+                out = pack_bits(self.orig, self.nbits_per_word, hi, in_col_offset)
+                for x in range(self.nbits):
+                    print(f"{c*self.nbits+x}: {format(out[x], '032b')}")
+
+                import ipdb
```
tsawada, line 170: can we move import at the top of the file? And I'm not sure if we should set_trace every loop.

blueoil/cmd/init.py
lm-noshiro: @iizukak The `el(se)` here is semantically redundant and I prefer to use `if`, but it's not necessary, so I'll undo it if you don't like it.

@tsawada That's a good idea. Is it appropriate to apply it in this PR?

tools/debug.py
lm-noshiro: I think it's better to do that too. Can I send it to another PR?

blueoil/cmd/init.py
tsawada: I think it's better to use tables here.

tools/debug.py
lm-noshiro: fixed.

blueoil/cmd/init.py
lm-noshiro: fixed

iizukak:
@lm-noshiro 
Thank you for fix some points.
Can you check conflicted files?

blueoil/converter/plugins/tf.py
```diff
@@ -485,41 +485,40 @@ def guess_node_format(input_node: Any) -> str:
...
-                    input_format_list.append(node_object_format)
-                return out_format, input_format_list
-        else:
+        if not isinstance(node, Node):
             return output_format, [output_format]
 
+        if node.get_format() is None:
+            out_format = output_format or guess_node_format(node)
+        else:
+            out_format = node.get_format()
```
tsawada, line 494: I'm not 100% sure but is this equivalent to:
```python
output_format = node.get_format() or output_format or guess_node_format(node)
```

tools/debug.py
```diff
@@ -128,57 +129,55 @@ def derive_full_bit_qwords(self, n):
...
             return ((n + (self.nbits_per_word - 1)) // self.nbits_per_word) * self.nbits
-        else:
-            raise NotImplemented
+        raise NotImplemented
 
     def make_packed_matrix(self, qmat):
         if self.arrangement == 'Sequential':
             raise NotImplemented
         if self.arrangement == 'BitInterleaving':
             raise NotImplemented
```
tsawada, line 141: it looks like these are not necessary. third `if` catches these as well.

tools/debug.py
```diff
@@ -128,57 +129,55 @@ def derive_full_bit_qwords(self, n):
...
+        qmat = np.zeros([self.row, self.col], dtype=np.uint32)
+
+        orig_col = self.orig.shape[1]
+        full_bit_col = orig_col // self.nbits_per_word
+        residual_bit_col = orig_col % self.nbits_per_word
+
+        def pack_bits(x, num_data, row, col_offset):
+            out = np.zeros([self.nbits], dtype=np.uint32)
+
+            for bi in reversed(range(0, num_data)):
```
tsawada, line 154: `range(0, a) == range(a)`
occurs a few times in this file.

PR #1063, opened 2020-05-28
Support FPGA with aarch64

### LINKED ISSUE #1042: Support FPGA with 64bit ARM processor
This is a new feature to enable FPGA with 64 bit ARM processor.
Now FPGA available only with 32bit ARM processor and we want to add 64 bit as well.

tk26eng:
This patch will support fpga with aarch64.
 
Fix #1042

### DIFF
--- a/blueoil/cmd/convert.py
+++ b/blueoil/cmd/convert.py
@@ -151,7 +151,7 @@ def make_all(project_dir, output_dir):
             subprocess.run(("make", "clean", "--quiet"))
             subprocess.run(("make", "build", "-j4", "--quiet") + _build_options(**arch, target=target))
             strip_binary(**arch, target=target)
-            output_file_path = os.path.join(output_dir, _output_binary_name)
+            output_file_path = os.path.join(output_dir, _output_binary_name(**arch, target=target))
             os.rename(output, output_file_path)
     # Return running directory
     os.chdir(running_dir)
--- a/blueoil/cmd/convert.py
+++ b/blueoil/cmd/convert.py
@@ -146,7 +146,7 @@ def make_all(project_dir, output_dir):
     os.chdir(project_dir)
 
     # Make each target and move output files
-    for arch in archtectures:
+    for arch in architectures:
         for target in targets:
             subprocess.run(("make", "clean", "--quiet"))
             subprocess.run(("make", "build", target_arch, target_use_fpga, target_type, "-j4", "--quiet"))
--- a/blueoil/cmd/convert.py
+++ b/blueoil/cmd/convert.py
@@ -72,22 +72,52 @@ def get_output_directories(output_root_dir):
     return output_directories
 
 
-def strip_binary(output):
+def _build_options(arch, use_fpga, target):
+    return "ARCH=" + arch + " USE_FPGA=" + use_fpga + " TYPE=" + target
+
+
+def _output_binary_name(arch, use_fpga, target):
+    if target is "executable"
+        output = "lm_"
+    elif target in {"dynamic", "static"}
+        output = "libdlk_"
+
+    output += arch
+
+    if use_fpga is "enable":
+        output += "_fpga"
+
+    if target is "executable"
+        output += ".elf"
+    elif target is "dynamic"
+        output += ".so"
+    elif target is "static"
+        output += ".a"
+
+    return output
+
+
+def strip_binary(arch, use_fpga, target):
     """Strip binary file.
 
     Args:
         output:
 
     """
+    # TODO: These operations should be performed in Makefile instead of here
 
-    if output in {"lm_x86.elf", "lm_x86_avx.elf"}:
-        subprocess.run(("strip", output))
-    elif output in {"libdlk_x86.so", "libdlk_x86_avx.so"}:
-        subprocess.run(("strip", "-x", "--strip-unneeded", output))
-    elif output in {"lm_arm.elf", "lm_fpga.elf"}:
-        subprocess.run(("arm-linux-gnueabihf-strip", output))
-    elif output in {"libdlk_arm.so", "libdlk_fpga.so"}:
-        subprocess.run(("arm-linux-gnueabihf-strip", "-x", "--strip-unneeded", output))
+    output = _output_binary_name(arch, use_fpga, target)
+
+    if arch in {"x86", "x86_avx"}
+        if target is "executable":
+            subprocess.run(("strip", output))
+        elif target is "dynamic":
+            subprocess.run(("strip", "-x", "--strip-unneeded", output))
+    elif arch is "arm"
+        if target is "executable":
+            subprocess.run(("arm-linux-gnueabihf-strip", output))
+        elif target is "dynamic":
+            subprocess.run(("arm-linux-gnueabihf-strip", "-x", "--strip-unneeded", output))
 
 
 def make_all(project_dir, output_dir):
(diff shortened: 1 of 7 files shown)


blueoil/converter/templates/Makefile
```diff
@@ -5,7 +5,7 @@ INPUTS_SRC_DIR := ./src/inputs
 .PHONY: help
 help:
 	@echo make clean
-	@echo make build ARCH={x86, x86_avx, aarch64, arm, fpga} TYPE={executable, dynamic, static}
+	@echo make build ARCH={x86, x86_avx, aarch64, arm, fpga, fpga_aarch64} TYPE={executable, dynamic, static}
```
primenumber, line 8: How about to introduce a flag like `USE_FPGA` ?

tk26eng:
run dlk test

blueoil/converter/templates/Makefile
tk26eng: @primenumber 
USE_FPGA looks good because it's easy to remember.
I'll try to use it instead.

tk26eng:
@iizukak 
Could you review this PR ?

Sorry for long PR.
This PR includes modification of test code because changing argument of Makefile needs fixing test.

iizukak:
@tk26eng 
I'm reviewing PR. It's takes time.
Is is ok to submmit review on next Monday?

tk26eng:
@iizukak 

> @tk26eng
> I'm reviewing PR. It's takes time.
> Is is ok to submmit review on next Monday?

Sure. Of couse it's OK on Monday.

blueoil/converter/templates/include/global.tpl.h
```diff
@@ -31,12 +31,23 @@ limitations under the License.
...
+#define IP_CSR_ADDR 0xB0000000
+#define HW_BASE_ADDR 0x1F000000
+#endif
+
+#define INPUT_OFFSET 0x00000000
+#define OUTPUT_OFFSET 0x08000000
+#define KERNEL_OFFSET 0x18000000
+#define THRESHOLD_OFFSET 0x1F000000
 
+#define INPUT_ADDR (HW_BASE_ADDR+INPUT_OFFSET)
```
iizukak, line 47: `INPUT_OFFSET ` is zero. Can we use

```
#define INPUT_ADDR HW_BASE_ADDR
```

and remove `INPUT_OFFSET` can be removed?

blueoil/converter/templates/include/global.tpl.h
```diff
@@ -31,12 +31,23 @@ limitations under the License.
...
+#ifdef AARCH32
 #define IP_CSR_ADDR 0xFF200000
-#define INPUT_ADDR 0x20000000
-#define OUTPUT_ADDR 0x28000000
-#define KERNEL_ADDR 0x38000000
-#define THRESHOLD_ADDR 0x3F000000
+#define HW_BASE_ADDR 0x20000000
+#else
+#define IP_CSR_ADDR 0xB0000000
+#define HW_BASE_ADDR 0x1F000000
```
iizukak, line 39: Where `0xB0000000 ` and `0x1F000000` values comes from?
General one?
It's better to add link to the documentation.

blueoil/converter/templates/include/global.tpl.h
tk26eng: It's better to keep `INPUT_OFFSET` because the value of `INPUT_OFFSET` might be changed.
And I would like to use same manner as others like `OUTPUT_OFFSET`.

blueoil/converter/templates/include/global.tpl.h
tk26eng: Those address values are temporal value.
There is no default value about them because different board has different value.
Actually it's better to make them variable to fit various hardware but this time current implementation just can use two case (arm and aarch64).
Maybe we can implement more flexible structure, but it should be another PR.

tk26eng:
@iizukak 

Could you check my comment ?

blueoil/converter/templates/include/global.tpl.h
iizukak: @tk26eng OK. I understood. Improve in another PR is reasonable

blueoil/converter/templates/include/global.tpl.h
iizukak: OK. Then no problem to keep it.

tk26eng:
@tfujiwar 
@primenumber 

Could you review this ?

tests/converter/test_code_generation.py
```diff
@@ -86,45 +86,65 @@ def dict_codegen_segmentation(cpu_name) -> dict:
...
+    additional_test_configuration = updated_dict(dict_codegen_classification(cpu_name, use_fpga),
                                                  {'use_run_test_script': True})
     additional_test_cases = [
-        {'use_avx': True, 'input_name': 'raw_image.png', 'test_level': TEST_LEVEL_FUTURE_TARGET},
-        {'use_avx': True, 'input_name': 'preprocessed_image.npy', 'from_npy': True},
-        {'use_avx': False, 'input_name': 'raw_image.png', 'test_level': TEST_LEVEL_FUTURE_TARGET},
-        {'use_avx': False, 'input_name': 'preprocessed_image.npy', 'from_npy': True},
+        {'input_name': 'raw_image.png', 'test_level': TEST_LEVEL_FUTURE_TARGET},
+        {'input_name': 'preprocessed_image.npy', 'from_npy': True},
+    ]
```
tfujiwar, line 127: Why are `test_cases` and `additional_test_cases` separated?

tests/converter/test_code_generation.py
```diff
@@ -298,7 +335,7 @@ def codegen_cpu(self,
                     from_npy=False,
                     need_arm_compiler=False,
                     cache_dma=False,
-                    use_avx=False,
+                    use_fpga=False,
```
tfujiwar, line 338: Using boolean for this flag is good. But in test cases, this flag is defined as a string.

iizukak:
This PR also need Python readability reviewer

blueoil/cmd/convert.py
```diff
@@ -100,31 +100,34 @@ def make_all(project_dir, output_dir):
...
+        ["ARCH=arm", "USE_FPGA=disable", "TYPE=dynamic", "libdlk_arm.so"],
+        ["ARCH=arm", "USE_FPGA=enable", "TYPE=dynamic", "libdlk_arm_fpga.so"],
+        ["ARCH=aarch64", "USE_FPGA=disable", "TYPE=dynamic", "libdlk_aarch64.so"],
+        ["ARCH=aarch64", "USE_FPGA=enable", "TYPE=dynamic", "libdlk_aarch64_fpga.so"],
+        ["ARCH=x86", "USE_FPGA=disable", "TYPE=static", "libdlk_x86.a"],
+        ["ARCH=x86_avx", "USE_FPGA=disable", "TYPE=static", "libdlk_x86_avx.a"],
+        ["ARCH=arm", "USE_FPGA=disable", "TYPE=static", "libdlk_arm.a"],
+        ["ARCH=arm", "USE_FPGA=enable", "TYPE=static", "libdlk_arm_fpga.a"],
+        ["ARCH=aarch64", "USE_FPGA=disable", "TYPE=static", "libdlk_aarch64.a"],
+        ["ARCH=aarch64", "USE_FPGA=enable", "TYPE=static", "libdlk_aarch64_fpga.a"],
```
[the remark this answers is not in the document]
tk26eng: I understood that.
I'll do that.

tests/converter/test_code_generation.py
```diff
@@ -86,45 +86,65 @@ def dict_codegen_segmentation(cpu_name) -> dict:
...
-    configurations.extend(get_configurations_by_test_cases(test_cases, dict_codegen_classification_resnet(cpu_name)))
-    configurations.extend(get_configurations_by_test_cases(test_cases, dict_codegen_object_detection(cpu_name)))
-    configurations.extend(get_configurations_by_test_cases(test_cases, dict_codegen_segmentation(cpu_name)))
+    configurations.extend(get_configurations_by_test_cases(test_cases,
+                                                           dict_codegen_classification(cpu_name, use_fpga)))
+    configurations.extend(get_configurations_by_test_cases(test_cases,
+                                                           dict_codegen_classification_resnet(cpu_name, use_fpga)))
+    configurations.extend(get_configurations_by_test_cases(test_cases,
+                                                           dict_codegen_object_detection(cpu_name, use_fpga)))
+    configurations.extend(get_configurations_by_test_cases(test_cases, dict_codegen_segmentation(cpu_name, use_fpga)))
```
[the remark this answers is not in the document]
tk26eng: That's simpler way.
I'll change the part.

tests/converter/test_code_generation.py
```diff
@@ -374,9 +406,9 @@ def codegen_cpu(self,
         self.assertTrue(os.path.exists(generated_lib))
 
         if not use_run_test_script:
-            if cpu_name == 'x86_64':
+            if cpu_name == 'x86' or cpu_name == 'x86_avx':
```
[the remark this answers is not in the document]
tk26eng: It makes sense.
Nice.

tests/converter/test_code_generation.py
```diff
@@ -133,39 +153,56 @@ def get_configurations_x86_64():
...
+    use_fpga = "disable"
+    test_cases = [
+        {'hard_quantize': True, 'threshold_skipping': True},
+        {'hard_quantize': True, 'threshold_skipping': False},
+        {'hard_quantize': False, 'threshold_skipping': True},
+        {'hard_quantize': False, 'threshold_skipping': False},
+    ]
+    configurations = get_configurations_by_architecture(test_cases, cpu_name, use_fpga)
+
+    return [(i, configuration) for i, configuration in enumerate(configurations)]
```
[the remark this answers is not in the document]
tk26eng: There's no reason why it should be list.
I'll use just enumerate to fix that.

tests/converter/test_code_generation.py
tk26eng: I'm sorry but I'm not sure why they are separated because it's code of someone else.
I checked the code again and maybe the difference is which way is used for the test.
First one calls dynamic library from test code directory.
Second one use run.py (not directly) to call dynamic library for the test.

tests/converter/test_code_generation.py
tk26eng: Maybe boolean might better as you say.
But this time, can I use string type instead?
Many all other part uses `use_fpga` as string because `use_fpga` will be argument of Makefile like `make build ARCH=aarch64 USE_FPGA=enable TYPE=exuecutable`.
If we use boolean, we need to convert it from boolean to string in the middle.

tk26eng:
Sorry for late response.
I've fixed part which you pointed out.
Could you review this again when you have time ?

tests/converter/test_code_generation.py
tfujiwar: Thank you! I've understood.

tests/converter/test_code_generation.py
tfujiwar: OK! No problem.

tk26eng:
> Thank you for fixing! It got much better than before. Sorry, I forgot to point this. Could you add new test functions (`TestCodeGenerationX8664Avx`, `TestCodeGenerationAarch64Fpga`) in Makefile and CI settings? Currently, they are not executed in CI.
> 
> * https://github.com/blue-oil/blueoil/blob/master/Makefile
> * https://github.com/blue-oil/blueoil/blob/master/.buildkite/dlk-pipeline.yml
> 
> You can check what is executed in CI at the link of `buildkite/dlk-test` below.

I totally missed those parts.
Thank you for pointing out them.
I've fixed them.

Issue #1026, opened 2020-05-07
Make `output_template/python/run.py` work with custom data_processors


tfujiwar:
We can use custom data_processors for training but they are not available from `output_template/python/run.py`.

tfujiwar:
Fixed by https://github.com/blue-oil/blueoil/pull/1064 .

PR #1064, opened 2020-05-28
Rewrote config module for inference.

odoku:

#1026

## Modified part

* Deleted because the imp module is deprecated.
* The duplicated process is made into a function and the implementation is simplified.
* Changed to can load custom processor.

## How to load custom processor

It is now possible to specify an import path for Pre / Post Processor in meta.yaml.

```yaml
POST_PROCESSOR:
  - blueoil.post_processor.FormatCenterNet:
      image_size: [160, 256]
      num_classes: 1
      down_ratio: 2
      top_k_results_output: 50
      use_offset: false
      data_format: NHWC

PRE_PROCESSOR:
  - blueoil.pre_processor.ResizeWithGtBoxes:
      size: [160, 256]
  - DivideBy255: null
```

### DIFF
--- a/output_template/python/config.py
+++ b/output_template/python/config.py
@@ -25,6 +25,17 @@ from blueoil import post_processor, pre_processor
 
 
 def load_yaml(config_file):
+    """Load meta.yaml that is output when the model is converted.
+
+    Args:
+        config_file (str): Path of the configuration file.
+
+    Returns:
+        EasyDict: Dictionary object of loaded configuration file.
+
+    Examples:
+        >>> config = load_yaml("/path/of/meta.yaml")
+    """
     with open(config_file) as config_file_stream:
         config = yaml.load(config_file_stream, Loader=yaml.Loader)
     # use only upper key.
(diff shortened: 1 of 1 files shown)


output_template/python/config.py
```diff
@@ -15,64 +15,137 @@
...
-        for class_name in p:
-            class_args = p[class_name]
-            if class_args is None:
-                class_args = {}
-            cls = getattr(module, class_name)
-            # Create none initialized processor `cls` instance.
+    processor_config = processor_config or []
+    for p in processor_config:
+        for class_name, kwargs in p.items():
+            try:
```
iizukak, line 108: I think it's better to use `if` `else` than `try` `catch`.
Can you rewrite like

```python
if hasattr(module, class_name):
    cls = getattr(module, class_name)
else:
    cls = import_from_string(class_name)
```

Sorry for not documented coding style...

output_template/python/config.py
```diff
@@ -15,64 +15,137 @@
...
+
+    Examples:
+        >>> join = import_from_string("os.path.join")
+    """
+    try:
+        return import_module(import_string)
+    except ImportError:
+        pass
+
+    try:
```
iizukak, line 140: How about
```python
if "." in import_string:
    module_name, attr_name = import_string.rsplit(".", 1)
```

output_template/python/config.py
```diff
@@ -15,64 +15,137 @@
...
+        module_name, attr_name = import_string.rsplit(".", 1)
+    except ValueError:
+        raise ImportError("Invalid import string '{}'".format(import_string))
+
+    module = import_module(module_name)
+    try:
```
iizukak, line 146: How about
```python
if hasattr(module, attr_name):
    return getattr(module, attr_name)
```

output_template/python/config.py
```diff
@@ -15,64 +15,137 @@
...
+
+    Examples:
+        >>> join = import_from_string("os.path.join")
+    """
+    try:
+        return import_module(import_string)
+    except ImportError:
+        pass
+
+    if "." in import_string:
```
tsawada, line 140: Use the guard pattern:

```python
if not ...:
    raise ...
_, _ = import_string(...)
```

output_template/python/config.py
```diff
@@ -15,64 +15,137 @@
...
+        module_name, attr_name = import_string.rsplit(".", 1)
+    else:
+        raise ImportError("Invalid import string '{}'".format(import_string))
+
+    module = import_module(module_name)
+    if hasattr(module, attr_name):
```
tsawada, line 146: same

output_template/python/config.py
```diff
@@ -15,64 +15,137 @@
...
+                    "max_output_size": 100,
+                    "per_class": True,
+                }},
+            ]
+        >>> post_processors = build_post_process(config)
+    """
+    return _build_process(post_processor, post_processor_config)
+
+
+def _build_process(module, processor_config=None):
```
tsawada, line 103: it seems we don't write to `processor_config`. We can use `[]` as the default value.

odoku: @tsawada 

The original code explicitly checks for None.
This code is very strange, but I suspect that None may be given as an argument.
I can't judge all the code, so I implemented it as a safety measure.
What do you think?

https://github.com/blue-oil/blueoil/blob/d18a9b4dde11a7675a67ef561da5a9a2cbecbf8d/output_template/python/config.py#L33-L40

tsawada: @iizukak 

Hmm. I grepped the code and it seems `pre_processor_config` comes from `config.PRE_PROCESSOR`.
Can this be None?
https://github.com/blue-oil/blueoil/blob/d18a9b4dde11a7675a67ef561da5a9a2cbecbf8d/output_template/python/run.py#L135

iizukak: `POST_PROCESSOR` can be None.
https://github.com/blue-oil/blueoil/blob/master/blueoil/configs/core/classification/lmnet_cifar100.py#L69

I don't know `PRE_PROCESSOR` None case.

tsawada: got it. As we handle POST_PROCESSOR as well, I think this is fine.

Issue #479, opened 2019-10-02
Consider upgrading to TensorFlow 2.0


tkng:
TensorFlow 2.0 was released. We should consider upgrading our TF dependency to 2.0.

However, since this is large major version up, we still don't know how much source code needs to be rewritten for the upgrade.

Fortunately, there's a script [tf_upgrade_v2.py](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/compatibility) to upgrade to TF 2.0. First, we should try this script and estimate how the upgrade is difficult.

yd8534976:
There is an official guide just for reference: https://www.tensorflow.org/guide/upgrade

ytfksw:
## Identification of targets and estimation of difficulty (for Future PRs) 
By using `tf_upgrade_v2.py` and changing it to tf.compat.v1, you can use the 1.x features in tf2 version. I want to know the targets to change checked the difficulty of the change.

Files and directories under "blueoil" directory are the targes to change. I directly applied `tf_upgrade_v2.py` each of the directories and files under `blueoil` to know the targets to change and to know the difficulty of the change.

### Convert targets
* [ ] blocks.py (*1)
* [ ] cmd
* [ ] configs
* [ ] converter: (*2)
* [ ] data_processor.py: (*2)
* [ ] datasets: (*2)
* [ ] layers (*1)
* [ ] metrics
* [ ] networks (*1)
* [ ] quantizations
* [ ] templates (*3)
* [ ] utils

### Not Convert targets
The following are files that have not been modified by running the script and are not subject to change:
* \_\_init\_\_.py
* common.py
* data_augmentor.py
* environment.py
* generate_lmnet_config.py
* post_processor.py
* pre_processor.py
* visualize.py

### Requires manual adjustment
Files can be converted automatically, but some manual adjustment is required. For `contrib`, the operator needs to be replaced carefully. Others may not be too difficult.

(*1)`contrib` is deprecated in tf2. `tf_upgrade_v2` does not make any changes. We need to change to another layer manually. For example, the following contrib layer are included:
* tf.contrib.layers.batch_norm
* tf.contrib.layers.fully_connected
* tf.contrib.layers.dropout
* tf.contrib.layers.separable_conv2d

`dlk` may not be able to be converted by replacing `contrib` with a similar layer.
We need to choose a layer that dlk supports. The accuracy also may decrease due to the change. We need to re-train the model check the accuracy after the replacements. This part can be difficult.

[...]

PR #1067, opened 2020-05-29
tf_upgrade_v2 for network/base.py

ytfksw:

I converted only the blueoil/network/base.py with the command line tool tf_upgrade_v2 to run on tensorflow2 because there are a lot of targets in the network folder.

Here's what I did:

install newest tenosrflow (tensorflow-gpu==1.5.2)
Run the upgrade script tf_upgrade_v2
Check the upgrade report for warnings and errors
Run test (make test)

* #479 
* https://www.tensorflow.org/guide/upgrade

### DIFF
--- a/blueoil/networks/base.py
+++ b/blueoil/networks/base.py
@@ -140,7 +140,7 @@ class BaseNetwork(object):
             learning_rate = self.optimizer_kwargs["learning_rate"]
 
         else:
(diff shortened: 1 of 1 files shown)


ytfksw:
As you can see in the log, there are no changes that need to be noted.
Because tf.name_scope in version 2 behaves differently from tf.name_scope in version 1,
it has been converted to tf.compat.v1.name_scope.

Issue #484, opened 2019-10-02
remove pandas dependency


tkng:
We're depending on `pandas`, but actually it's used for just reading CSV or json files.

To reduce disk image size, we should remove this dependency.

suttang:
Currently, we are using pandas not only for reading CSV in the following codes.
we need to rewrite these codes.

https://github.com/blue-oil/blueoil/blob/502e844f76286ad10f4c21d3d9bfa2c36cac15ca/lmnet/lmnet/datasets/delta_mark.py#L272-L279

https://github.com/blue-oil/blueoil/blob/502e844f76286ad10f4c21d3d9bfa2c36cac15ca/lmnet/executor/output_event.py#L65-L89

Other pandas code is used for just reading CSV.

lm-lily:
@iizukak 
Iizuka san, I created issue #1052 which is related to this issue. 
Do we still need delta mark in blueoil?

lm-lily:
delta_mark.py with pandas dependency is ignored as filed in #1066 .

Issue #1066, opened 2020-05-28
Remove delta-mark


lm-lily:
This is a minor issue to remove the outdated `.../datasets/delta-mark.py` from Blueoil.

lm-lily:
Solved.

PR #1072, opened 2020-05-29
As for issue #484: remove pandas dependency

lm-lily:
Remove pandas dependencies.

Fix for issue #484.


lm-lily:
delta_mark.py with pandas dependency is ignored as filed in #1066 .

iizukak:
@lm-lily Can you remove pandas dependency from `setup.cfg`?
https://github.com/blue-oil/blueoil/blob/master/setup.cfg#L26

blueoil/datasets/pascalvoc_2007.py
```diff
@@ -54,12 +54,12 @@ def num_max_boxes(self):
             return 42
 
     def _annotation_file_from_image_id(self, image_id):
-        annotation_file = os.path.join(self.annotations_dir, "{:06d}.xml".format(image_id))
+        annotation_file = os.path.join(self.annotations_dir, "{}.xml".format(image_id))
```
iizukak, line 57: Why `:06d` removed?
Related to pandas?

blueoil/datasets/pascalvoc_2007.py
```diff
@@ -54,12 +54,12 @@ def num_max_boxes(self):
...
         return annotation_file
 
     def _image_file_from_image_id(self, image_id):
         """Return image file name of a image."""
-        return os.path.join(self.jpegimages_dir, "{:06d}.jpg".format(image_id))
+        return os.path.join(self.jpegimages_dir, "{}.jpg".format(image_id))
```
iizukak, line 62: Same here.

blueoil/datasets/pascalvoc_base.py
```diff
@@ -184,16 +182,16 @@ def _all_image_ids(self, data_type=None, is_debug=False):
...
 
         filename = os.path.join(self.imagesets_dir, data_type + ".txt")
-        df = pd.read_csv(
-            filename,
-            delim_whitespace=True,
-            header=None,
-            names=['image_id'])
+        image_id = list()
+
+        with open(filename) as f:
```
iizukak, line 187: Why don't you use csv module?

blueoil/datasets/camvid.py
```diff
@@ -80,15 +79,13 @@ def files_and_annotations(self):
...
-        df = pd.read_csv(
-            filename,
-            delim_whitespace=True,
-            header=None,
-            names=['image_files', 'label_files'],
-        )
 
-        image_files = df.image_files.tolist()
-        label_files = df.label_files.tolist()
+        image_files, label_files = list(), list()
```
iizukak, line 83: Why Don't you use csv module?

blueoil/datasets/pascalvoc_2007.py
lm-lily: > @lm-lily Can you remove pandas dependency from `setup.cfg`?
> https://github.com/blue-oil/blueoil/blob/master/setup.cfg#L26

@iizukak 
There is still delta-mark remained in Blueoil (issue #1066).
Is it ok to remove pandas dependency from `setup.cfg` before issue #1066 is solved?

lm-lily:
> @lm-lily Can you remove pandas dependency from `setup.cfg`?
> https://github.com/blue-oil/blueoil/blob/master/setup.cfg#L26

@iizukak 
There is still delta-mark remained in Blueoil (issue #1066).
Is it ok to remove pandas dependency from `setup.cfg` before issue #1066 is solved?

blueoil/datasets/pascalvoc_2007.py
lm-lily: The file name extracted using my submitted code (after removing pandas dataframe method) failed with `:06d` format.
I have to adjust this part to make it right.

blueoil/datasets/pascalvoc_2007.py
lm-lily: The reason is the same as reply above.

blueoil/datasets/pascalvoc_base.py
lm-lily: The lines here are about reading a txt file, not csv file.

blueoil/datasets/camvid.py
lm-lily: This is for reading a text file instead of csv file.

iizukak:
@lm-lily OK. Then, It's good to remove DeLTA Mark dataset loader first.
Thank you for trying solve a lot of issues!

lm-lily:
This PR is put to pending until PR #1086 is merged into Blueoil.

iizukak:
@lm-lily Thank you for the review request. I'll review again soon.

blueoil/datasets/ilsvrc_2012.py
```diff
@@ -57,8 +56,8 @@ def __init__(
     @functools.lru_cache(maxsize=None)
     def classes(self):
         # wget https://raw.githubusercontent.com/Lasagne/Recipes/master/examples/resnet50/imagenet_classes.txt
-        df = pd.read_csv(os.path.join(self.data_dir, 'imagenet_classes.txt'), sep="\n", header=None)
-        return df[0].tolist()
+        linelist = [line.rstrip('\n') for line in open(os.path.join(self.data_dir, 'imagenet_classes.txt'))]
```
iizukak, line 59: Can we use `readlines` instead of `line.rstrip('\n')` to reduce code?

blueoil/datasets/ilsvrc_2012.py
```diff
@@ -72,17 +71,14 @@ def _init_files_and_annotations(self):
...
-            txt_file,
-            delim_whitespace=True,
-            header=None,
-            names=['filename', 'class_id'])
+        files, labels = list(), list()
+        with open(txt_file) as f:
+            for line in f:
+                items = line.split()
+                files.append(items[0])
+                labels.append(int(items[1]))
```
iizukak, line 79: Why we should cast to `int` ?

blueoil/datasets/ilsvrc_2012.py
```diff
@@ -72,17 +71,14 @@ def _init_files_and_annotations(self):
     def _files_and_annotations(self):
         txt_file = self.texts[self.subset]
 
-        df = pd.read_csv(
-            txt_file,
-            delim_whitespace=True,
-            header=None,
-            names=['filename', 'class_id'])
+        files, labels = list(), list()
+        with open(txt_file) as f:
```
iizukak, line 75: Python `csv` module can load space separated file.
https://docs.python.org/ja/3/library/csv.html#csv.Dialect.delimiter

It's good to use standard library than implement by ourself.

blueoil/datasets/camvid.py
```diff
@@ -259,15 +254,13 @@ def files_and_annotations(self):
...
+        with open(filename) as f:
```
iizukak, line 259: Python `csv` module can load space separated file.
https://docs.python.org/ja/3/library/csv.html#csv.Dialect.delimiter

It's good to use standard library than implement by ourself.

blueoil/datasets/camvid.py
```diff
@@ -182,15 +179,13 @@ def files_and_annotations(self):
...
```
iizukak, line 184: Python `csv` module can load space separated file.
https://docs.python.org/ja/3/library/csv.html#csv.Dialect.delimiter

It's good to use standard library than implement by ourself.

blueoil/datasets/camvid.py
iizukak: Python `csv` module can load space separated file.
https://docs.python.org/ja/3/library/csv.html#csv.Dialect.delimiter

It's good to use standard library than implement by ourself.

blueoil/cmd/output_event.py
```diff
@@ -63,32 +74,42 @@ def output(tensorboard_dir, output_dir, metrics_keys, steps, output_file_base="m
...
+    data_by_row = list(map(list, zip(*value_matrix)))
+    columns.insert(0, "step")
 
     output_csv = os.path.join(output_dir, "{}.csv".format(output_file_base))
-    df.to_csv(output_csv, index=False)
+
+    with open(output_csv, "w") as fp:
+        wr = csv.writer(fp)
+        wr.writerow(columns)
+        for row in data_by_row:
```
iizukak, line 105: Can we use `writerows` ?
https://docs.python.org/ja/3/library/csv.html#csv.csvwriter.writerows

blueoil/cmd/output_event.py
```diff
@@ -63,32 +74,42 @@ def output(tensorboard_dir, output_dir, metrics_keys, steps, output_file_base="m
 
     columns = [_column_name(event_accumulator, metrics_key)
                for event_accumulator, metrics_key in itertools.product(event_accumulators, metrics_keys)]
-    columns.sort()
-    df = pd.DataFrame([], columns=columns)
+
+    values_step_dict = {}
```
iizukak, line 78: Do we need this initialize code?

blueoil/cmd/output_event.py
```diff
@@ -63,32 +74,42 @@ def output(tensorboard_dir, output_dir, metrics_keys, steps, output_file_base="m
...
-        df = df[steps, :]
-
-    df = df.sort_index(ascending=False)
-
-    # index to column. and re-order column.
-    df["step"] = df.index
-    df = df[["step"] + columns]
+            for step in step_list:
+                if step not in values_step_dict:
+                    values_step_dict[step] = ''
```
iizukak, line 93: Key of `values_step_dict` looks `value`. Not `step`. Correct?
I'm not sure why this line's key is `step`.

blueoil/datasets/pascalvoc_base.py
```diff
@@ -184,16 +182,16 @@ def _all_image_ids(self, data_type=None, is_debug=False):
             raise ValueError("Must provide data_type = train or val or trainval or test")
 
         filename = os.path.join(self.imagesets_dir, data_type + ".txt")
-        df = pd.read_csv(
-            filename,
-            delim_whitespace=True,
-            header=None,
-            names=['image_id'])
+        image_id = list()
```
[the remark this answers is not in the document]
lm-lily: In reply to https://github.com/blue-oil/blueoil/pull/1072#discussion_r449313275

```read().splitlines()``` is possible. 
I will modify my code to use ```read()```.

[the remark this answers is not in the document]
Comment (lm-lily) [reply blueoil/datasets/pascalvoc_base.py:185]: > `image_id` looks like single value.
> How about `image_ids` ?

Noted.

File: blueoil/datasets/pascalvoc_2007.py (hunk 1 above)
Comment (lm-lily) [reply blueoil/datasets/pascalvoc_2007.py:57]: > > The file name extracted using my submitted code (after removing pandas dataframe method) failed with :06d format.
> 
> Sorry, Why this happen?

The image_id has the format of sample data: ```2007_000032```. If we apply format ```{:06}``` to it, the filename will not match the format we need. It will be ```2007000032``` instead of ```2007_000032```, and hence cause the error of ```file not found```.

File: blueoil/datasets/ilsvrc_2012.py (hunk 5 above)
Comment (lm-lily) [reply blueoil/datasets/ilsvrc_2012.py:59]: If we want to use readlines, we need to write the codes as follow, which will have more codes than the one written above.
```
with open(os.path.join(self.data_dir, 'imagenet_classes.txt'), "r") as f:
    linelist = f.readlines()
```
And this will include the unwanted '/n' in the list.

File: blueoil/datasets/ilsvrc_2012.py (hunk 6 above)
Comment (lm-lily) [reply blueoil/datasets/ilsvrc_2012.py:79]: The labels is taken as int in the implementation of pandas dataframe.
In my codes, the labels data is read in as string initially, I need to cast it to int to match with the consequent data handling.

File: blueoil/datasets/ilsvrc_2012.py (hunk 7 above)
Comment (lm-lily) [reply blueoil/datasets/ilsvrc_2012.py:75]: @iizukak 
If I implement this block using csv, the code will be like the followed:
```
import csv

files, labels = list(), list()
with open(txt_file) as f:
    reader = csv.reader(f, delimiter=' ')
    for lines in reader:
        files.append(lines[0])
        labels.append(lines[1])
```

In term of amount of code, both are about the same, except csv requires extra `import csv` line, and this also means an extra library is required in this code.
Since there is no advantage in simplicity, is it still better to use csv library in this case?
I don't have much knowledge in good programming style, so I hope to get my doubt cleared in this discussion.
Thank you!

File: blueoil/datasets/camvid.py (hunk 8 above)
Comment (lm-lily) [reply blueoil/datasets/camvid.py:259]: Reply same as https://github.com/blue-oil/blueoil/pull/1072#discussion_r453459110.

File: blueoil/datasets/camvid.py (hunk 9 above)
Comment (lm-lily) [reply blueoil/datasets/camvid.py:184]: Reply same as https://github.com/blue-oil/blueoil/pull/1072#discussion_r453459110.

File: blueoil/cmd/output_event.py (hunk 10 above)
Comment (lm-lily) [reply blueoil/cmd/output_event.py:105]: Thank you! It's a good suggestion, I have changed it.

File: blueoil/cmd/output_event.py (hunk 12 above)
Comment (lm-lily) [reply blueoil/cmd/output_event.py:93]: @iizukak 
I am sorry, it's my mistake, I forget to change line#37 to `(event.value, event.step)`. The sequence is reversed.
Thank you for capturing this significant mistake!

File: blueoil/cmd/output_event.py (hunk 11 above)
Comment (lm-lily) [reply blueoil/cmd/output_event.py]: It is not necessary. I will remove this.
Thank you for pointing it out.

File: blueoil/datasets/ilsvrc_2012.py (hunk 5 above)
Comment (iizukak) [reply blueoil/datasets/ilsvrc_2012.py:59]: I see.

**Comment by iizukak** [comment]
@lm-lily 
Thanks. Please wait for the Python readability review.

File: blueoil/datasets/pascalvoc_2007.py (hunk 1 above)
Comment (tfujiwar) [reply blueoil/datasets/pascalvoc_2007.py:57]: @lm-lily I'm still not sure why this is related to removing pandas. If this is an existing bug, please make another PR to fix that.

Comment (lm-lily) [reply blueoil/datasets/pascalvoc_2007.py:57]: @tfujiwar 
Thank you Fujiwara san.
Then should I revert this change?

Comment (lm-lily) [reply blueoil/datasets/pascalvoc_2007.py]: @tfujiwar 
`lmnet-test` failed when I revert the above code.
```
[2020-07-20T05:41:28Z] self = <blueoil.datasets.pascalvoc_2007.Pascalvoc2007 object at 0x7ffb8908dc50>
--
  | [2020-07-20T05:41:28Z] image_id = '000085'
  | [2020-07-20T05:41:28Z]
  | [2020-07-20T05:41:28Z]     def _annotation_file_from_image_id(self, image_id):
  | [2020-07-20T05:41:28Z] >       annotation_file = os.path.join(self.annotations_dir, "{:06d}.xml".format(image_id))
  | [2020-07-20T05:41:28Z] E       ValueError: Unknown format code 'd' for object of type 'str'
  | [2020-07-20T05:41:28Z]
  | [2020-07-20T05:41:28Z] ../blueoil/datasets/pascalvoc_2007.py:57: ValueError

```
My previous explanation was a fault.
In the original code, the data is handled by pandas dataframe, which will handle the data type conversion internally.
My code takes in the file name as string type and that caused the lmnet-test failure.
Since the file name format is safe to be read as whole, I replace `{:06d}` to `{}` as the quick fix.
I will fix this code again and please advice if you have any better suggestion.

**Comment by lm-lily** [comment]
@tfujiwar 
May I know about the review status? Thank you.

File: blueoil/cmd/output_event.py [hunk 14]
```diff
@@ -63,32 +74,40 @@ def output(tensorboard_dir, output_dir, metrics_keys, steps, output_file_base="m
...
                for event_accumulator, metrics_key in itertools.product(event_accumulators, metrics_keys)]
-    columns.sort()
-    df = pd.DataFrame([], columns=columns)
+
+    value_matrix = []
+
+    for metrics_key in metrics_keys:
+        if not value_matrix:
+            step_list = sorted(_step_list(event_accumulator, metrics_key), reverse=True)
+            value_matrix.append(step_list)
```
tfujiwar, line 83: With these lines, the list of steps comes from the first metrics_key. Is that the same with the original code? Did you check the result is the same?

lm-lily: @tfujiwar 
Yes, I have verified the result.
The list of steps is the same and is repeated for each metrics_key, it is enough to just read it once from any metrics_key.

tfujiwar: OK. Then what about this code? I think we don't need to use a for-loop.
```py
value_matrix = [sorted(_step_list(event_accumulator, metrics_keys[0]), reverse=True)]
```

lm-lily: @tfujiwar 
This can't work, there will be `TypeError: 'set' object is not subscriptable` because `metrics_keys` is a set.
Can I make it this way? (use `break` to break the for loop).
```   
for metrics_key in metrics_keys:
        step_list = sorted(_step_list(event_accumulator, metrics_key), reverse=True)
        value_matrix.append(step_list)
        break
```

tfujiwar: Hmm, I've understood the situation. It might be better to refactor the existing code also but this PR is OK 👍

PR #1079, opened 2020-06-04
use byte arrays to remove floating point number literals

### LINKED ISSUE #1078: Avoid error comes from floating point literal decimal representation in code generation
Conversion between floating point value ↔ decimal string representation cause some small error.
Several solution exists:
- Using hexadecimal representation (But not allowed in C++14 standard)
- Embedding as array of binary value and using `memcpy`
I think that second solution may be more preferred.

tkng:

Conversion between floating point value :left_right_arrow: decimal string representation cause some small error.

In this pull request, I removed floating point number literals. Instead, parameters are represented as an array of unsigned char, then casted to floating point number by `reinterpret_cast`.

fix #1078

### DIFF
--- a/blueoil/converter/templates/manual/consts/input.tpl.cpp
+++ b/blueoil/converter/templates/manual/consts/input.tpl.cpp
@@ -20,9 +20,9 @@ limitations under the License.
 {% if node.transposed_data %}
 
 #ifdef RUN_ON_FPGA
-static Base<{{ node.dtype.cpptype() }}>::type {{ node.name }}_raw[] = {
+alignas(4) static unsigned char {{ node.name }}_raw[] = {
   {% for d in node.transposed_data -%}
-  {{- d -}},
+  {{- pack_to_bytes(d) -}},
   {%- endfor %}
 };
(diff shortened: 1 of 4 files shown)


blueoil/converter/templates/manual/consts/input.tpl.cpp
```diff
@@ -33,10 +34,11 @@ static constexpr decltype({{ node.name }}_output)::tensor_info_t<std::size_t> {{
 const TensorView<{{ node.dtype.cpptype() }}, MemoryLayout::{{ node.transposed_dimension_format }}> {{ node.name }}_output(
     reinterpret_cast<{{ node.dtype.cpptype() }}*>({{ node.name }}_raw),
     {{ node.name }}_shape);
+
 #elif defined USE_NEON || defined USE_AVX
-static Base<{{ node.dtype.cpptype() }}>::type {{ node.name }}_raw[] = {
+alignas(16) static unsigned char {{ node.name }}_raw[] = {
```
primenumber, line 39: Where from 16-byte align?

PR #1082, opened 2020-06-05
Remove YouTube Faces format from keypoint detection tutorial.

oatawa1:
Remove YouTube Faces format from keypoint detection tutorial. 
Because keypoint detection can only choose one data format.
```
[?] choose dataset format: Mscoco for Single-Person Pose Estimation
 > Mscoco for Single-Person Pose Estimation
```


oatawa1:
@iizukak Thank you.

Issue #804, opened 2020-02-07
The time measurement results are different between run.py and lm_*.elf


hadusam:
After #803, we can reduce the difference, but there are still gaps between `run.py` and `lm_*elf`.
We need to discuss and investigate this issue. 🤔 

## Test case
* Network: `LmnetV1Quantize`

## Run with FPGA on DE10-Nano
* run.py's inference: 16ms
* elf results: 9ms
```
root@blueoil-de10nano-04:~/masu_test/argo/export/save.ckpt-156250/32x32/output/python# python run.py --trial 1 -i ../../inference_test_data/raw_image.png -c ../models/meta.yaml -m ../models/lib/lib_fpga.so
...
INFO:__main__:Benchmark avg result(sec) for 1 trials: pre_process: 0.007635  inference: 0.016171 post_process: 0.016254  Total: 0.04006
root@blueoil-de10nano-04:~/masu_test/argo/export/save.ckpt-156250/32x32/output/python# head output/output.json {
    "benchmark": {
        "inference": 0.016170999999999935, 
        "post": 0.01625399999999999, 
        "pre": 0.007635000000000058, 
        "total": 0.040059999999999985
    }, 
    "classes": [
        {
            "id": 0, 
root@blueoil-de10nano-04:~/masu_test/argo/export/save.ckpt-156250/32x32/output/python# ../models/lib/lm_fpga.elf ../../inference_test_data/000_images_placeholder\:0.npy ../../inference_test_data/133_output\:0.npy 
-------------------------------------------------------------
Comparison: Default network test  succeeded!!!
-------------------------------------------------------------
TotalInitTime 3068,  sum:3.068ms
TotalRunTime 9425,  sum:9.425ms
...
```

## Run without FPGA on DE10-Nano
* run.py's inference: 36ms
* elf results: 19ms
```
root@blueoil-de10nano-04:~/masu_test/argo/export/save.ckpt-156250/32x32/output/python# python run.py --trial 1 -i ../../inference_test_data/raw_image.png -c ../models/meta.yaml -m ../models/lib/lib_arm.so
...
INFO:__main__:Benchmark avg result(sec) for 1 trials: pre_process: 0.007937  inference: 0.036363 post_process: 0.017636  Total: 0.061936
root@blueoil-de10nano-04:~/masu_test/argo/export/save.ckpt-156250/32x32/output/python# head output/output.json 
{
    "benchmark": {
        "inference": 0.036363000000000034, 
[...]

PR #1083, opened 2020-06-08
Use time.time/time.perf_counter instead of time.clock for demo scripts

hadusam:
`time.clock` does not return real-time, so this is not suitable for calculating elapsed time.

This is same as the previous case of `run.py`:  https://github.com/blue-oil/blueoil/issues/804#issuecomment-583297531


output_template/python/lmnet/utils/demo.py
```diff
@@ -147,16 +147,21 @@ def release(self):
...
+    start = get_time()
 
     data = pre_process(image=image)["image"]
     data = np.expand_dims(data, axis=0)
 
-    network_only_start = time.clock()
+    network_only_start = get_time()
     result = nn.run(data)
-    fps_only_network = 1.0/(time.clock() - network_only_start)
+    fps_only_network = 1.0/(get_time() - network_only_start)
```
tfujiwar, line 162:
```suggestion
    fps_only_network = 1.0 / (get_time() - network_only_start)
```

output_template/python/lmnet/utils/demo.py
```diff
@@ -147,16 +147,21 @@ def release(self):
...
 
     output = post_process(outputs=result)['outputs']
 
-    fps = 1.0/(time.clock() - start)
+    fps = 1.0/(get_time() - start)
```
tfujiwar, line 166:
```suggestion
    fps = 1.0 / (get_time() - start)
```

PR #1085, opened 2020-06-09
Fix checking pad

primenumber:
correct order: `[top, bottom, left, right]`
(it is same as tf.pad)

### DIFF
--- a/tests/converter/test_consistency_check.py
+++ b/tests/converter/test_consistency_check.py
@@ -114,7 +114,7 @@ class TestConsistencyCheck(unittest.TestCase):
             [1, 3, 3, 3],
             Float32(),
             input_ops,
-            pads=[1, 1, 2, 2],
+            pads=[1, 2, 1, 2],
             strides=[2, 2]
         )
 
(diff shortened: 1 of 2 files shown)


tests/converter/test_consistency_check.py
```diff
@@ -114,7 +114,7 @@ def test_conv_consistency(self) -> None:
             [1, 3, 3, 3],
             Float32(),
             input_ops,
-            pads=[1, 1, 2, 2],
+            pads=[1, 2, 1, 2],
```
iizukak, line 117: @primenumber 
Sorry, I can not understand this change clear.
pad format is `[top, bottom, left, right]` and this line means using different padding value for top and buttom. right?

iizukak, line 117: Sorry, I can not understand this change clearly.  
pad format is [top, bottom, left, right] and this line means using different padding value for top and buttom. right?

primenumber: Yes.

Issue #1092, opened 2020-06-15
Remove easydict dependency


iizukak:
We use easydict to access dictionary.
https://github.com/blue-oil/blueoil/blob/master/setup.cfg#L23

But the library longer maintenance(~2years).
https://pypi.org/project/easydict/

I want to remove library dependence.

patrick-su:
@iizukak seems like we can close this?

iizukak:
@patrick-su 
NO. @odoku san implement SmartDict class, but not remove easydict dependency.

PR #1094, opened 2020-06-16
Implemented easydict.

odoku:

#1092

### DIFF
--- a/blueoil/utils/smartdict.py
+++ b/blueoil/utils/smartdict.py
@@ -1,6 +1,6 @@
-class EasyDict(dict):
+class SmartDict(dict):
     def __init__(self, d=None, **kwargs):
-        super(EasyDict, self).__init__()
+        super(SmartDict, self).__init__()
         self.update(d or {}, **kwargs)
 
     def update(self, d=None, **kwargs):
@@ -17,14 +17,14 @@ class EasyDict(dict):
             ]
         elif isinstance(value, dict) and not isinstance(value, self.__class__):
             value = self.__class__(value)
-        super(EasyDict, self).__setitem__(name, value)
+        super(SmartDict, self).__setitem__(name, value)
 
     def __getattr__(self, name):
         if name in self:
             return self[name]
-        return super(EasyDict, self).__getattr__(name)
+        return super(SmartDict, self).__getattr__(name)
 
     def __setattr__(self, name, value):
         if hasattr(self, name):
-            return super(EasyDict, self).__setattr__(name, value)
+            return super(SmartDict, self).__setattr__(name, value)
         self[name] = value
--- a/blueoil/utils/smartdict.py
+++ b/blueoil/utils/smartdict.py
@@ -18,7 +18,7 @@
 class SmartDict(dict):
     def __init__(self, d=None, **kwargs):
         super(SmartDict, self).__init__()
-        self.update(d or {}, **kwargs)
+        self.update(d, **kwargs)
 
     def update(self, d=None, **kwargs):
         d = d or {}
--- a/blueoil/utils/smartdict.py
+++ b/blueoil/utils/smartdict.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # =============================================================================
+
+
 class SmartDict(dict):
     def __init__(self, d=None, **kwargs):
         super(SmartDict, self).__init__()
--- a/blueoil/utils/smartdict.py
+++ b/blueoil/utils/smartdict.py
@@ -1,3 +1,18 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 The Blueoil Authors. All Rights Reserved.
+#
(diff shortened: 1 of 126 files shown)


blueoil/utils/smartdict.py
```diff
@@ -0,0 +1,30 @@
+class SmartDict(dict):
```
iizukak, line 1: Please add license header.

tests/unit/util_tests/test_smartdict.py
```diff
@@ -0,0 +1,63 @@
+import pytest
```
iizukak, line 1: Same here.

iizukak:
@tfujiwar Please review 🙇

a-hanamoto:
I'll check tomorrow so please wait for a while. Thanks!

blueoil/utils/smartdict.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+# =============================================================================
+
+
+class SmartDict(dict):
+    def __init__(self, d=None, **kwargs):
+        super(SmartDict, self).__init__()
+        self.update(d or {}, **kwargs)
+
+    def update(self, d=None, **kwargs):
```
a-hanamoto, line 23: Can we make this method non-public? `_update`

blueoil/utils/smartdict.py
```diff
@@ -0,0 +1,47 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
+
+
+class SmartDict(dict):
+    def __init__(self, d=None, **kwargs):
+        super(SmartDict, self).__init__()
+        self.update(d or {}, **kwargs)
```
a-hanamoto, line 21: `d` will be replaced `d or {}` soon after calling `update`, so `d` is OK, I think.

blueoil/utils/smartdict.py
```diff
@@ -0,0 +1,47 @@
...
+    def update(self, d=None, **kwargs):
+        d = d or {}
+        d.update(kwargs)
+        for key, value in d.items():
+            self[key] = value
+
+    def __setitem__(self, name, value):
+        if isinstance(value, (list, tuple)):
+            value = [
+                self.__class__(x) if isinstance(x, dict) else x
```
a-hanamoto, line 32: `SmartDict`? I wonder why there are both of `self.__class__` and `SmartDict`

blueoil/utils/smartdict.py
odoku: @a-hanamoto 
Since the update function is in dict, I think it is necessary for SmartDict which inherits dict.
https://docs.python.org/ja/3/library/stdtypes.html?highlight=dict#dict.update

blueoil/utils/smartdict.py
odoku: @a-hanamoto 
Your question is, why do you write "self.__class__(x)" instead of "SmartDict(x)"?
This is because if you inherit "SmartDict", "self.__class__" will be a subclass.

```python
class HyperSmartDict(SmartDict):
    pass

d = HyperSmartDict()
d.__class__ == HyperSmartDict
```

a-hanamoto: Got your point. Thanks!

blueoil/utils/smartdict.py
odoku: I added test case for update function. Please review it.

blueoil/utils/smartdict.py
```diff
@@ -0,0 +1,47 @@
...
+                for x in value
+            ]
+        elif isinstance(value, dict) and not isinstance(value, self.__class__):
+            value = self.__class__(value)
+        super(SmartDict, self).__setitem__(name, value)
+
+    def __getattr__(self, name):
+        if name in self:
+            return self[name]
+        return super(SmartDict, self).__getattr__(name)
```
tfujiwar, line 42: I think `__getattr__` should raise an AttributeError instead of delegating to the super class. Because Dict doesn't have a method `__getattr__`, the current error message is like:
```
AttributeError: 'super' object has no attribute '__getattr__'
```
It should be like:
```
AttributeError: 'SmartDict' object has no attribute 'x'
```

blueoil/utils/smartdict.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __setattr__(self, name, value):
+        if hasattr(self, name):
+            return super(SmartDict, self).__setattr__(name, value)
+        self[name] = value
```
tfujiwar, line 47: I think `self[name] = value` is enough. Currently, the results are different for an existing and non-existing key.
```py
d = SmartDict(a=None)
d.a = [{"k":"v"}]
d.b = [{"k":"v"}]

d.a[0].__class__  # dict
d.b[0].__class__  # SmartDict
```

iizukak:
OA

PR #1097, opened 2020-06-16
Remove unused utils

joelN123:
function `restore_saved_image_size` and function `init_config` are not used anymore. By doing grep -r "restore_saved_image_size" blueoil/ these keywords only appear in their definitions.
In this case, it's better to remove them. Unless developers make use of them, or there are future plans that require them.


joelN123:
ah, I branched from another branch, not master.

PR #1099, opened 2020-06-17
Update tutorials to use CPU in converting step

### LINKED ISSUE #750: Docker image for CPU

tk26eng:
This patch updates tutorials to use `CUDA_VISIBLE_DEVICES=-1`.
The environment variable means use CPU instead of GPU in converting step.

Fix #750

### DIFF
--- a/docs/tutorial/image_cls.md
+++ b/docs/tutorial/image_cls.md
@@ -140,7 +140,7 @@ Convert trained model to executable binary files for x86, ARM, and FPGA.
 Currently, conversion for FPGA only supports Intel Cyclone® V SoC FPGA.
 
     $ docker run --rm \
-        -e CUDA_VISIBLE_DEVICES=-1\
+        -e CUDA_VISIBLE_DEVICES=-1 \
         -e OUTPUT_DIR=/home/blueoil/saved \
         -v $(pwd)/saved:/home/blueoil/saved \
         blueoil_$(id -un):{TAG} \
--- a/docs/tutorial/image_cls.md
+++ b/docs/tutorial/image_cls.md
@@ -140,7 +140,7 @@ Convert trained model to executable binary files for x86, ARM, and FPGA.
 Currently, conversion for FPGA only supports Intel Cyclone® V SoC FPGA.
 
     $ docker run --rm \
-        -e CUDA_VISIBLE_DEVICES=0 \
+        -e CUDA_VISIBLE_DEVICES=-1\
         -e OUTPUT_DIR=/home/blueoil/saved \
         -v $(pwd)/saved:/home/blueoil/saved \
         blueoil_$(id -un):{TAG} \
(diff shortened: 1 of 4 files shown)


tk26eng:
@iizukak 

Could you review this ?

PR #1104, opened 2020-06-19
Add update pillow version in DE10-nano installation

oatawa1:
Reported from @lm-tajimi about Outpute_template contains newer Pillow library dependency. That DE10 OS image's Pillow is not supported.

I updated the document and add  update pillow dependency in the DE10-nano installation

-

### DIFF
--- a/docs/install/install.md
+++ b/docs/install/install.md
@@ -213,12 +213,15 @@ Third, update kernel modules with:
 
 ---
 
-## Install required packages
+## Install required packages and dependencies
 
-Login to the FPGA board and update required packages.
+Login to the FPGA board and update required packages and dependencies.
 
     $ apt-get update
     $ apt-get install python-dev python-setuptools python-pip unzip
+    $ export LC_ALL=en_US.UTF-8
+    $ pip install --upgrade pip
+    $ pip install -U Pillow --no-cache-dir
 
 ---
 


docs/install/install.md
```diff
@@ -213,12 +213,15 @@ Third, update kernel modules with:
...
 
-## Install required packages
+## Install required packages and dependencies
 
-Login to the FPGA board and update required packages.
+Login to the FPGA board and update required packages and dependencies.
 
     $ apt-get update
     $ apt-get install python-dev python-setuptools python-pip unzip
+    $ export LC_ALL=en_US.UTF-8
```
iizukak, line 222: @oatawa1 
Maybe It's better to update `requirements.txt` .
https://github.com/blue-oil/blueoil/blob/master/output_template/python/requirements.txt

oatawa1: Did you mean I should update the Pillow version in requirement.txt? 
But it is already updated (6.2.2).

I wonder I should fix the document to install requirement.txt instead.
But if it is the case, maybe we can refer to this document (https://docs.blue-oil.org/tutorial/run_fpga.html#run-the-demonstration). Because requirement.txt does not come with the DE10-nano OS image.

iizukak: @oatawa1 
Sorry for confusion.
`requirements.txt` is already update? good.
Yes, Please write documentation to use `pip install -r requirements.txt` before run inference on DE10-Nano.

Here is the documentation but it's not good location...
https://github.com/blue-oil/blueoil/blob/master/output_template/python/README.md#getting-started-fpga-x86-and-arm

oatawa1: Sorry, I am still not fully understand.

I agree with you that it's not good location to put `pip install -r requirements.txt` in https://github.com/blue-oil/blueoil/blob/master/output_template/python/README.md#getting-started-fpga-x86-and-arm

The `pip install -r requirements.txt` is not only located in the location you point out, it is also written in **2.5 Run a trained neural network on the FPGA board**.
https://docs.blue-oil.org/tutorial/run_fpga.html#run-the-demonstration  
So I think we already have the document in the step before run inference on DE10-Nano.

iizukak: @oatawa1 Umm, yes. Then the issue is not a problem ...?

oatawa1:
This document already exists in https://docs.blue-oil.org/tutorial/run_fpga.html#run-the-demonstration.

Thus I close this PR.

PR #1110, opened 2020-06-22
fix loading of dataset class from yaml config

joelN123:
In some cases, the dataset class saved to the `config.yaml` is `'abc.DATASET_CLASS'`. In these cases, trying to load `config.yaml` fails, causing error. However, in these cases, the base class is a sensible class, which it can load. So this fix is to load the base class in these cases.

The default situation is to load the `config.py`, not the `config.yaml`. For this reason, this bug has not been previously caught by the tests. However, I think the point of the `config.yaml` is to provide a separate config file that could also be used. So, I think it's good to fix it.

p.s. the error seems to occur when init was used before training. It also occurs during the auto-tests. Maybe the error doesn't happen for dataset classes which are hand-crafted (not generated).

This PR also fixes the error that is blocking PR https://github.com/blue-oil/blueoil/pull/1096 due to loading from `config.yaml`.
#### case 1:
```
$ print(config_dict['DATASET_CLASS'])
<class 'abc.DATASET_CLASS'>
$ print(config_dict['DATASET_CLASS'].__bases__[0])
<class 'blueoil.datasets.delta_mark.ObjectDetectionBase'>
```

(in this case, load `<class 'blueoil.datasets.delta_mark.ObjectDetectionBase'>`)

#### case 2:
```
$ print(config_dict['DATASET_CLASS'])
<class 'blueoil.datasets.cifar100.Cifar100'>
$ print(config_dict['DATASET_CLASS'].__bases__[0])
<class 'blueoil.datasets.base.Base'>
```

(in this case, load as usual `<class 'blueoil.datasets.cifar100.Cifar100'>`)

## How to reproduce the bug
```
$ docker run -it --runtime=nvidia -e CUDA_VISIBLE_DEVICES="0" -e PYTHONPATH=.:lmnet:dlk/python/dlk --rm -v `pwd`:/home/blueoil -w /home/blueoil --user=$(id -u):$(id -g) <docker-container> bash
$ python blueoil/cmd/main.py init
```
to create a config `my_model.py`. then train by:
```
$ python blueoil/cmd/main.py train -c my_model.py -e testing123 --recreate
```
Converting at this point will work fine because it will load from the `config.py` in the `saved` directory. But instead, if you do
```
$ rm saved/testing123/config.py
```
Then now, running convert
```
$ python blueoil/cmd/main.py convert -e testing123
```
will load from the `config.yaml` instead, and will cause an error:

[error_message.txt](https://github.com/blue-oil/blueoil/files/4811902/error_message.txt)

The main point is: `yaml.constructor.ConstructorError: while constructing a Python object cannot find 'DATASET_CLASS' in the module 'abc'`.

After making the change in this PR, the error doesn't happen any more. So, it is able to load the config from `config.yaml`.

## Other possible ways to fix it
In terms of readability, the fix in this PR is may be not ideal, but I'm not sure how else to fix the bug in a simple way. Some other ideas for ways to fix it:
- don't use this yaml file anymore (maybe delete it? I'm not sure if it is used for anything other than loading the config).
- a similar solution as in this PR, but written in a more readable way (I don't know it, but maybe there is a way).
- make a larger change to the behaviour of loading and saving `DATASET_CLASS`. In particular, [line 41 of classification.tpl.py](https://github.com/blue-oil/blueoil/blob/0d22ed42f78a8ea8682d389b6daf356fec783fab/blueoil/templates/lmnet/classification.tpl.py#L41) and other template files, when this is loaded, it becomes a class, then for the `config.yaml`, it is saved as a class. The further loading of this class is what causes the difficulty, so changing how this is done (to some completely different way) could be a solution.

### DIFF
--- a/blueoil/utils/config.py
+++ b/blueoil/utils/config.py
@@ -216,6 +216,9 @@ def _save_config_yaml(output_dir, config):
     config_dict = _easy_dict_to_dict(config)
     file_path = os.path.join(output_dir, file_name)
 
+    if str(config_dict['DATASET_CLASS']) == "<class 'abc.DATASET_CLASS'>":
+        config_dict['DATASET_CLASS'] = config_dict['DATASET_CLASS'].__bases__[0]
+
     class Dumper(yaml.Dumper):
         def ignore_aliases(self, data):
             return True


joelN123:
ah right, I should first request for owner approval, then after that, for readability approval. I'll remember for next time!

iizukak:
@joelN123 
Please check my understand, In your reproduce step,
Why do you remove `config.py` file?

joelN123:
Yes, so I'm trying to show there is an error when loading the `config.yaml`. But if the `config.py` exists, then the `config.yaml` is not loaded, so the error doesn't happen. So, it's necessary to delete the `config.py`, to show that there is an error when loading `config.yaml`. (in [export](https://github.com/blue-oil/blueoil/blob/0d22ed42f78a8ea8682d389b6daf356fec783fab/blueoil/cmd/export.py#L174), it calls [utils/config.py](https://github.com/blue-oil/blueoil/blob/0d22ed42f78a8ea8682d389b6daf356fec783fab/blueoil/utils/config.py#L72), it loads only one file, not both).

joelN123:
Maybe "fix **saving** of dataset class **for** yaml config" would be a better name for this PR. Well, the loading was causing the error, but the way of fixing is to change the way it is saved.

joelN123:
I put a comment into the code because I think that someone reading the code later will not know why these two lines of code are here. (unless they look at the blame of the line and find this PR, which would use up their time). So, better to put a comment to explain.

joelN123:
@a-hanamoto, please could you do a readability review when you have time

blueoil/utils/config.py
```diff
@@ -216,6 +216,10 @@ def _save_config_yaml(output_dir, config):
     config_dict = _easy_dict_to_dict(config)
     file_path = os.path.join(output_dir, file_name)
 
+    # For this particular class name, the parent class is more appropriate for the config.
+    if str(config_dict['DATASET_CLASS']) == "<class 'abc.DATASET_CLASS'>":
```
a-hanamoto, line 220: I'm not sure the type of `config_dict['DATASET_CLASS']`, but maybe we can write like
`if config_dict['DATASET_CLASS'] == abc.DATASET_CLASS`

joelN123: yes, it's not ideal to compare using the `str` function, but I'm not sure how to avoid it.
The type of `config_dict['DATASET_CLASS']` is `<class 'abc.ABCMeta'>` In both cases. However, it is necessary to distinguish between the two cases. So, an example for each of the two cases:
1.
using the dataset module [cifar10.py](https://github.com/blue-oil/blueoil/blob/master/blueoil/datasets/cifar10.py)
`type(config_dict['DATASET_CLASS'])` is `<class 'abc.ABCMeta'>`
`str(config_dict['DATASET_CLASS'])` is `<class 'blueoil.datasets.cifar100.Cifar100'>`
`config_dict['DATASET_CLASS'].__bases__[0]` is `<class 'blueoil.datasets.base.Base'>`
2.
using `python blueoil/cmd/main.py init` to generate the dataset class
`type(config_dict['DATASET_CLASS'])` is `<class 'abc.ABCMeta'>`
`str(config_dict['DATASET_CLASS'])` is `<class 'abc.DATASET_CLASS'>`
`config_dict['DATASET_CLASS'].__bases__[0]` is `<class 'blueoil.datasets.image_folder.ImageFolderBase'>`

In the case of 2, the useful thing to save is the base class, but in case 1 we can save in the usual way. If we only see the type, then we can't know which thing to do. But if seeing the string of the object, it becomes clear what to do.

a-hanamoto: Got it. How about using `isinstance` ?

joelN123: ah, right, `isinstance` would be a pythonic way to do it. Thank you, I will try it out

joelN123: I tried it out, but not much luck. I write the main details down here. Since it is a class, I think `issubclass` function makes sense. However, it's difficult to access the class, or a parent of it.

An example line of the config file created by `blueoil/cmd/main.py init` is:
`DATASET_CLASS = type('DATASET_CLASS', (ImageFolderBase,), {'extend_dir': 'cifar/train', 'validation_extend_dir': 'cifar/test'})`
then in line 133 of [utils/config.py](https://github.com/blue-oil/blueoil/blob/383aebfd4a1a4150ef4a09b7b1d1221871a4ead7/blueoil/utils/config.py#L133) (loading the config for starting training), the exec command generates the class `<class 'abc.DATASET_CLASS'>`.

I thought maybe it's possible to do something like `issubclass(config_dict['DATASET_CLASS'], type('DATASET_CLASS', (), {}))`, but I think this will always evaluate False, because python will see this as a definition of a new class.
In fact, even using `type('DATASET_CLASS', (ImageFolderBase,), {'extend_dir': 'cifar/train', 'validation_extend_dir': 'cifar/test'})` and comparing with `issubclass` evaluates as False in the previous example.

Anyway, I think doing this is no better than using `str(config_dict['DATASET_CLASS'])`. The benefit of using `issubclass` would be if we had a module which created the class in the first place. But what we actually have is `blueoil/cmd/main.py init` generates a config containing the definition of the class, and the class is generated from that.

I feel like to avoid using the check `str(config_dict['DATASET_CLASS'])` (or similar), then it is necessary to change the behaviour of how `DATASET_CLASS` is used for generated configs. I think `<class 'abc.DATASET_CLASS'>` is the class of generated configs (including the configs used in the tests).

a-hanamoto: @joelN123 I heard iizukak-san will contact you so could you please re-assign me as a reviewer after the discussion?

joelN123: yes, that's right. ok, I will remember to do that. Thank you for your time on this PR.

PR #1119, opened 2020-07-09
Insert Cast operator automatically

### LINKED ISSUE #474: Tensor which outputs quantized data must output both packed and unpacked tensor in some case
Example case with pseudo code:
```
a = quantize(in)
k1 = quantize(w1)
o1 = qconv(a, k1) // conv1
o2 = conv(a, w2) // conv2
```
in `conv1` , `a` must be float tensor, but in `conv2` , `a` must be packed tensor.
But current implementaion, we cannot output both packed and unpacked tensor.

primenumber:
Fix #474 .
Quantized network is using `QUANTIZED_PACKED` typed tensor, but some operators require `float` tensor.
It will cause compile error.
This PR enable automatic-casting routine.
Old workaround routines `pass_propagate_output_type_backward` and `restore_shape` are removed.


blueoil/converter/templates/include/func/cast.h
```diff
@@ -0,0 +1,26 @@
+/* Copyright 2018 The Blueoil Authors. All Rights Reserved.
```
tfujiwar, line 1: The year of first publication.
```suggestion
/* Copyright 2020 The Blueoil Authors. All Rights Reserved.
```

blueoil/converter/templates/src/func/cast.cpp
```diff
@@ -0,0 +1,170 @@
+/* Copyright 2018 The Blueoil Authors. All Rights Reserved.
```
tfujiwar, line 1:
```suggestion
/* Copyright 2020 The Blueoil Authors. All Rights Reserved.
```

primenumber:
:thinking: 

> This PR is not approved yet.

PR #1122, opened 2020-07-17
FIx to set correct gs:// path

### LINKED ISSUE #1121: OUTPUT_DIR with the gcs path does not work
`os.path.abspath()` removes one of the path slashes.

```
$ python
Python 3.6.9 (default, Jul  3 2019, 15:36:16)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> path = 'gs://hoge/fuga/piyo/'
>>> OUTPUT_DIR = os.path.abspath(path)
>>> OUTPUT_DIR
'/home/User/workingdir/gs:/hoge/fuga/piyo'
```

https://github.com/blue-oil/blueoil/blob/r0.25/blueoil/environment.py#L42

iizukak:

Fix #1121

If `path` is started with `gs://`, then we do *not* add convert to local absolute path.

### DIFF
--- a/blueoil/environment.py
+++ b/blueoil/environment.py
@@ -33,13 +33,13 @@ _CHECKPOINTS_DIR = os.path.join(OUTPUT_DIR, "{experiment_id}", "checkpoints")
 
 def set_data_dir(path):
     global DATA_DIR
-    DATA_DIR = os.path.abspath(path)
+    DATA_DIR = path if is_gcs_path(path) else os.path.abspath(path)
     os.environ["DATA_DIR"] = DATA_DIR
 
 
 def set_output_dir(path):
     global OUTPUT_DIR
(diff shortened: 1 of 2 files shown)


iizukak:
I tested both `DATA_DIR` and `OUTPUT_DIR` are GCS path case manually.

blueoil/environment.py
```diff
@@ -105,3 +105,13 @@ def teardown_test_environment():
...
+
+
+def is_gcs_path(path):
+    """Check argument string is GCS path or not
+    Args:
+        path: str
+    Returns:
+        True when string is GCS path
+    """
+    return path[0:5] == "gs://"
```
a-hanamoto, line 117:
```suggestion
    return path.startswith("gs://")
```

iizukak:
OA

PR #1135, opened 2020-07-29
Fix comments typo in base.py

oatawa1:
Fix comments typo in base.py


oatawa1:
@iizukak @tsawada Thank you for your review.

PR #1141, opened 2020-07-30
CI: Add mypy test

tfujiwar:
with `soft_fail = true` because the current code has many errors about typing.


tsawada:
🎉  Thank you!

PR #1148, opened 2020-07-31
Make lint/type tests available without docker build

kchygoe:
This change makes lint/type tests executed on container.
Docker build is no longer required for these tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,18 @@ deps:
 
 .PHONY: install
 install: deps
-	pip install -e .[cpu,tests,docs]
-	pip install pycocotools==2.0.0
+	pip3 install -U pip
+	pip3 install -e .[cpu]
+	pip3 install pycocotools==2.0.0
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,13 +12,18 @@ deps:
...
 
 .PHONY: install-gpu
 install-gpu: install
-	pip install -e .[gpu]
-	pip install -e .[dist]
+	pip3 install -e .[gpu]
+	pip3 install -e .[dist]
 
 .PHONY: lint
 lint:
```
tfujiwar, line 29: Maybe we can remove the target `lint` now.

kchygoe: it would be better to remain target `lint` and remove `test-blueoil-pep8`

kchygoe: Fixed target names from `test-blueoil-pep8` to `test-lint` and from `test-blueoil-mypy` to `test-mypy`

Issue #1152, opened 2020-08-03
Improve code quality in output_template


hadusam:
### No tests
There is no tests for `output_template/python` except run.py of e2e test.
It is better to add tests as well as other python files (e.g. under `blueoil/`).

### Redundant codes
usb_camera_demo.py and motion_jpeg_server_from_camera.py are demo scripts and internal logic is similar but there are redundant implementation about visualization and measure inference time(fps). And run.py has also function to measure inference time.
I think there is some better way to improve readability of these codes.

PR #1155, opened 2020-08-07
Add unittest CI environment for inference scripts (output_template)

hadusam:
A part of #1152 

* add Dockerfile for inference
* add test-unit-inference to Makefile
* add test-unit-inference to buildkite pipeline
* add a small test for run.py to check test environment works well (I am planning to add more tests after this PR)

### DIFF
--- a/output_template/python/dev-requirements.txt
+++ b/output_template/python/dev-requirements.txt
@@ -1,0 +1,3 @@
+-r requirements.txt
+pytest==5.3.5
+pytest-xdist==1.31.0
--- a/output_template/python/dev-requirements.txt
+++ b/output_template/python/dev-requirements.txt
@@ -1,3 +1,0 @@
--r requirements.txt
-pytest==5.3.5
-pytest-xdist==1.31.0
--- a/tests/output_template/test_run.py
+++ b/tests/output_template/test_run.py
@@ -1,0 +1,13 @@
+import time
+from unittest import TestCase
+
+from run import _timerfunc
+
+
+class TestRun(TestCase):
+
+    def test_timerfunc(self):
+        def sleep(sec):
+            time.sleep(sec)
+        _, got = _timerfunc(sleep, [1])
+        self.assertEqual(int(got), 1)
--- a/.buildkite/lmnet-pipeline.yml
+++ b/.buildkite/lmnet-pipeline.yml
@@ -28,6 +28,14 @@ steps:
     timeout_in_minutes: "30"
     env:
       BUILDKITE_CLEAN_CHECKOUT: 'true'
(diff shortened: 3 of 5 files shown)


kchygoe:
Good PR!

hadusam:
@iizukak Could you review this PR? 🙇

iizukak (approved):
OA
Sorry for the slow review

output_template/python/dev-requirements.txt
```diff
@@ -0,0 +1,3 @@
+-r requirements.txt
```
tfujiwar, line 1:
```suggestion
```

tfujiwar (commented):
pytest is installed but the test is using unittest. Is that intentional?

tests/output_template/test_run.py
```diff
@@ -0,0 +1,13 @@
...
+from run import _timerfunc
+
+
+class TestRun(TestCase):
+
+    def test_timerfunc(self):
+        def sleep(sec):
+            time.sleep(sec)
+        _, got = _timerfunc(sleep, [1])
+        self.assertEqual(int(got), 1)
```
tfujiwar, line 13: It's safer to use assertAlmostEqual.
```suggestion
        self.assertAlmostEqual(got, 1)
```

hadusam: @tfujiwar Thanks!!
In my understanding, `sleep(1)` will sleep at least 1 sec, and maybe longer than 1 sec, so `int(got)` is enough safe in (1<=got<2).

https://docs.python.org/3/library/time.html#time.sleep
> Also, the suspension time may be longer than requested by an arbitrary amount because of the scheduling of other activity in the system.
> Changed in version 3.5: The function now sleeps at least secs even if the sleep is interrupted by a signal, except if the signal handler raises an exception

In my environment, sleep took 1.001488673966378 seconds to complete. So if we use `assertAlmostEqual`, we need to set `places=<2`, but I am not sure about the good value of `places` in other situations. I am afraid that this test is fragile with a strict assertion.
Ummm, how about using `self.assertAlmostEqual(got, 1, places=1)`?

hadusam:
@tfujiwar 
> pytest is installed but the test is using unittest. Is that intentional?

Nice review! 👍 We don't need `pytest` for these unit tests.
I removed dev-requirements.txt (including pytest) and change to use unittest runner of python3 😄

tests/output_template/test_run.py
tfujiwar: Ah, you are right. `assertAlmostEqual` was too strict. Thank you.

tfujiwar (approved):
RA
Nice PR for quality improvement 👍

PR #1156, opened 2020-08-12
Python3 support of motion_jpeg_server_from_camera.py

hadusam:
After #1155 

* python3 support of motion_jpeg_server_from_camera.py
* add small test of motion_jpeg_server_from_camera.py


iizukak (approved):
OA

tfujiwar (approved):
RA

PR #1163, opened 2020-08-14
Fix an error of runtime with no bounding box of object detection

### LINKED ISSUE #1162: An error happens with no detected bounding box on object detection
The `run` program occur an problem when it runs like below.

```
$ ./run -i ../person.png -c ../meta.yaml 
classes: 
Person
task: IMAGE.OBJECT_DETECTION
expected input shape: 
1
224
384
3
Run
Floating point exception (core dumped)
```

This problem happens with no bounding box in object detection task.
It should be fixed to check the condition like that.

tk26eng:
This PR fix the error of runtime.
The problem occurs when no bounding box is detected with object detection task.
Then some values divided by zero, so the exception happens.

Fix #1162


tk26eng:
@primenumber 

Could you please review this PR ?

primenumber (approved):
RA for C++

Issue #1167, opened 2020-08-18
Incompatible version `absl-py`


Suikaba:
There exists a dependency error of `absl-py` in the current blueoil (964c64b).

Following tutorials, I have run a docker command like this:
```
docker run --rm -it \
     -v $(pwd)/cifar:/home/blueoil/cifar \
     -v $(pwd)/config:/home/blueoil/config \
     blueoil_$(id -un):v0.27.0-1-g964c64b \
     blueoil init -o config/cifar10_test.py
```

Then this error occured:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/pkg_resources/__init__.py", line 582, in _build_master
    ws.require(__requires__)
  File "/usr/local/lib/python3.6/dist-packages/pkg_resources/__init__.py", line 899, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/local/lib/python3.6/dist-packages/pkg_resources/__init__.py", line 790, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (absl-py 0.9.0 (/usr/local/lib/python3.6/dist-packages), Requirement.parse('absl-py<0.9,>=0.7'), {'tensorflow-metadata'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/blueoil", line 33, in <module>
    sys.exit(load_entry_point('blueoil', 'console_scripts', 'blueoil')())
  File "/usr/local/bin/blueoil", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/local/lib/python3.6/dist-packages/importlib_metadata/__init__.py", line 105, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/blueoil/blueoil/cmd/main.py", line 21, in <module>
[...]

PR #1168, opened 2020-08-18
Fix absl-py version to 0.8.1

oatawa1:
This PR fix absl-py version to 0.8.1 (the latest version before 0.9)
Because absl-py require <0.9, >=0.7

Note that Blueoil code didn't require ` absl-py` but it required by TensorFlow.

#1167

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,8 +33,10 @@ install_requires =
 
 [options.extras_require]
 cpu =
+  absl-py==0.8.1
   tensorflow==1.15.3
 gpu =
+  absl-py==0.8.1
   tensorflow-gpu==1.15.3
 dist =
   horovod==0.19.5


iizukak:
@oatawa1 
I'm wondering why this issue happen.
There is no problem if we use 0.8.1?
I think some library required 0.9 in dependency and this issue happen.

oatawa1:
@iizukak I thought the 0.8.1 could work so.

I checked it using `pip check`
First check which dependency need `absl-py`
```sh
# pip check
tensorflow-metadata 0.23.0 requires absl-py, which is not installed.
tensorflow-gpu 1.15.3 requires absl-py, which is not installed.
tensorflow-datasets 1.3.0 requires absl-py, which is not installed.
tensorboard 1.15.0 requires absl-py, which is not installed.
```

After I installed the absl-py.

absl-py==0.9.0
```sh
# pip check
tensorflow-metadata 0.23.0 has requirement absl-py<0.9,>=0.7, but you have absl-py 0.9.0.
```

absl-py==0.8.1
```sh
# pip check
No broken requirements found.
```

iizukak:
@oatawa1 Great.

oatawa1:
In our case, the `tensorflow-datasets==1.3.0` required absl-py but it did not specify the version so it installed the latest version.
Then the tensorflow-meradata is installed with tensorflow which required < 0.9, so it got error.

PR #1183, opened 2020-08-26
Add explanations about serial communication

lm-kasuga:
Add an explanation about the permission of the device file of the FPGA board

### DIFF
--- a/docs/install/install.md
+++ b/docs/install/install.md
@@ -110,7 +110,7 @@ First, change the permission of your board to `666`.
 ```
 $ sudo chmod 666 /dev/[your_fpga_board]
 ```
-If you don't want to change it every time, you can also do the following.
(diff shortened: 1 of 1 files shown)


docs/install/install.md
```diff
@@ -105,6 +105,20 @@ Login to the board via serial.
     - The path name of your FPGA board should look similar to `/dev/ttyUSB0`.
 - In macOS
     - The path name of your FPGA board should look similar to `/dev/tty.usbserial-A106I1IY`.
+
+First, change the permission of your board to `666`.
+```
+$ sudo chmod 666 /dev/[your_fpga_board]
+```
+If you don't want to change it every time, you can also do the following.
```
iizukak, line 113: This operation is for Ubuntu. Please not that. We also support macOS.

iizukak (approved):
OA

PR #1191, opened 2020-08-31
fix : Change time.clock to time.perf_counter

lm-kasuga:


iizukak (approved):
OA

tfujiwar (approved):
RA
nice refactoring 👍

PR #1193, opened 2020-09-02
Fix : Replace multiporcess to multithread

lm-kasuga:
The communication between some processes is considered slow.
So replace multi processes to multi threads, and use shared memory

### DIFF
--- a/output_template/python/usb_camera_demo.py
+++ b/output_template/python/usb_camera_demo.py
@@ -21,7 +21,6 @@ from __future__ import unicode_literals
 import time
 import os
 import sys
-from multiprocessing import Process, Queue
 from time import sleep
 from collections import deque
 import queue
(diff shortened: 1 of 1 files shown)


lm-kasuga:
I thought that the interprocess communication makes `usb_camera_demo.py` slowly, but that was not the reason for the delay.
A spin lock caused by `q_output.empty()` was a major cause.
https://github.com/blue-oil/blueoil/blob/ad162bdccc1b5f133b1be98014d34bd1ebfc2134/output_template/python/usb_camera_demo.py#L190
`multiprocessing.Queue.get` blocks its thread until a element appended, so this spin lock isn't necessary.
I removed this line and run on local, which was as fast as multithread's one.

lm-kasuga:
Please ignore this PR

PR #1197, opened 2020-09-04
Enable to save pretrain_vars.txt to GCS

hadusam:
Related #1205 
* Enable to save pretrain_vars.txt to GCS
* Refactor to use file_io module


iizukak (approved):
Thanks
OA

tfujiwar (approved):
RA

PR #1198, opened 2020-09-07
Remove xenial from test pipelines

kchygoe:
- Remove xenial env from dlk-pipeline
- Fix blueoil-pipeline to use fixed bionic env


iizukak (approved):
OA

Issue #937, opened 2020-03-19
The contents of keep_me.md should be placed at right place


tk26eng:
`blueoil/converter/templates/debug/keep_me.md` includes information about `-dbg` flag for `generate_project.py`. It should be placed at right place.

patrick-su:
@tk26eng do you mean we should be the content in `keep_me.md` to somewhere... like here?
https://github.com/blue-oil/blueoil/blob/master/docs/converter/

tk26eng:
@patrick-su 
Yes. That's right.
But maybe simply moving document is not enough because `debug` directory will be empty after moving `keep_me.md`.
Empty directory (`debug` directory) will be removed automatically by git command and no `debug` directory might occur problem for `generate_project.py`.
We need to consider also about this.

PR #1200, opened 2020-09-10
move keep_me.md(#937)

lm-kajihara:
Solve #937

Moved(renamed) keep_me.md and keep exist `debug` directory by creating dummy file(.file).
I referred to the following stackoverflow.
https://stackoverflow.com/questions/18773598/creating-folders-inside-a-github-repository-without-using-git#answer-18791455


iizukak:
@lm-kajihara 
Thank you for the PR.
Why do you want to move this file?

IMO: `.gitkeep` is better than `.file`.

iizukak (changes requested):
`keep_me.md` should be rename to `debug.md` or something.
and this documentation says `Debug data will be saved in this directory `, but it is not correct.
Debug directory is `blueoil/converter/templates/debug/` now.

iizukak (approved):
OA

Issue #1188, opened 2020-08-28
update for docs? (about python command)


joelN123:
previously, it was possible to run training using
```
docker run --runtime=nvidia -e CUDA_VISIBLE_DEVICES="0" -e PYTHONPATH=.:lmnet:dlk/python/dlk --rm -v `pwd`:/home/blueoil -w /home/blueoil -v <path to dataset dir>:/home/blueoil/dataset:ro --user=$(id -u):$(id -g) <docker image> python blueoil/cmd/main.py train -c blueoil/configs/core/classification/lmnet_quantize_cifar100.py -e testing123 --recreate
```
But in recent versions (e.g. 27) of Blueoil, it will cause an error `/docker-entrypoint.sh: line 13: exec: python: not found`. Instead, if replacing `python` in the above command with `python3` then it works fine.

This is because of recent changes where python2 is no longer supported?

If this is correct, maybe we should update docs, for example, in `https://github.com/blue-oil/blueoil/blob/master/docs/usage/train.md`, where it is written `$ python blueoil/cmd/main.py train -c config/test.py`

Another option is maybe to use a symlink or alias so that it's possible to use python command to run python3? ahh, no, this is bad practice? https://legacy.python.org/dev/peps/pep-0394/

joelN123:
oh, I forgot to close this issue (which was resolved by https://github.com/blue-oil/blueoil/pull/1202). Closing now.

PR #1202, opened 2020-09-11
change python to python3 in usage docs

joelN123:
This PR is for the following issue https://github.com/blue-oil/blueoil/issues/1188. The point is simply to change the documentation of Blueoil command from `python` (which causes error in the docker container), to `python3`.

There is also documentation that mentions `python` in [converter/overview.md](https://github.com/blue-oil/blueoil/blob/master/docs/converter/overview.md), [converter/shared_library.md](https://github.com/blue-oil/blueoil/blob/master/docs/converter/shared_library.md), and several files in docs/tutorial. But in each of those cases, it looks like it is being explained from the perspective of outside a container, where someone would likely be using a python environment etc. So I didn't change the documentation in those cases.

### testing it out

I'm not sure it was really necessary, but I ran (using python3, rather than simply python) each of : init, train, convert, predict, export, evaluate, measure_latency, profile_model (for cifar classification), in the style of
```
docker run --runtime=nvidia -e CUDA_VISIBLE_DEVICES="0" -e PYTHONPATH=.:lmnet:dlk/python/dlk --rm -it -v `pwd`:/home/blueoil -w /home/blueoil -v <path to dataset dir>:/home/blueoil/dataset:ro --user=$(id -u):$(id -g) <docker image> python3 blueoil/cmd/main.py <specific command>
```
(or in some cases, like blueoil/cmd/\<specific command\>.py). They all seemed to run fine.

### DIFF
--- a/docs/usage/convert.md
+++ b/docs/usage/convert.md
@@ -1,7 +1,7 @@
 # Convert your training result to FPGA ready format
 
 ```
-$ python blueoil/cmd/main.py convert -e test_20180101000000
+$ python3 blueoil/cmd/main.py convert -e test_20180101000000
 
 Usage:
   main.py convert [OPTIONS]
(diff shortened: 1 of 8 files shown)


joelN123:
run test

iizukak (approved):
OA
Thanks.

PR #1203, opened 2020-09-11
change the option name in export documentation to "image"

joelN123:
In the docs, this option is called "images", but when running the export code, an error will come up
```
Usage: export.py [OPTIONS]
Try 'export.py -h' for help.

Error: no such option: --images
```
It's because the actual option is "image". (which you can see by running 'export.py -h').
When using "image", it works fine. So this PR is just for correcting the documentation.


iizukak (approved):
OA
Thanks.

PR #1204, opened 2020-09-11
Add blueoil.io module for I/O with GCS support

hadusam:
A part of #1205

### DIFF
--- a/blueoil/io/file_io.py
+++ b/blueoil/io/file_io.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 The Blueoil Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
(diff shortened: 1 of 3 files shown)


iizukak (approved):
OA

blueoil/io/file_io.py
```diff
@@ -0,0 +1,81 @@
...
+import numpy as np
+from PIL import Image
+from tensorflow.io import gfile
+
+
+def _is_local_path(path: str) -> bool:
+    for p in {"gs://", "s3://", "hdfs://"}:
+        if path.startswith(p):
+            return False
+    return True
```
tfujiwar, line 28: You can pass a tuple to `startswith`.
https://docs.python.org/3/library/stdtypes.html#str.startswith

tests/unit/io_tests/test_file_io.py
```diff
@@ -0,0 +1,145 @@
...
+def test_mkdir():
+    with tempfile.TemporaryDirectory() as d:
+        path = os.path.join(d, "test")
+        file_io.mkdir(path)
+        assert os.path.isdir(path)
+
+
+def test_makedirs():
+    with tempfile.TemporaryDirectory() as d:
+        path = os.path.join(os.path.join(d, "test"), "test")
```
tfujiwar, line 61: Is this enough?
```suggestion
        path = os.path.join(d, "test", "test")
```

tfujiwar (commented):
Just small suggestions.

tests/unit/io_tests/test_file_io.py
hadusam: Fixed! 👍

blueoil/io/file_io.py
hadusam: Oh, thanks! I simplify checking the local path and remove this function. Please check it! 👍

tfujiwar (approved):
RA
Thank you~~

PR #1207, opened 2020-09-15
Fix confitions not to use tfds_pre_processor and tfds_augumentor

hadusam:
`TFDSMixin` always has attr of `tfds_pre_processor` and `tfds_augumentor`, so the conditions are always true and we cannot use normal pre-process and augmentation when we train with TFDS. I fixed these conditions to be able to run train with normal pre-process and augmentation.
https://github.com/blue-oil/blueoil/blob/ba646b908e7036f7ef5249d1523e35123873697f/blueoil/datasets/tfds.py#L117-L118


iizukak (approved):
OA

tfujiwar (approved):
RA
Thank you!!

PR #1209, opened 2020-09-15
Avoid ValueError for TFDS with BBox on the last pixel

### LINKED ISSUE #1208: Failed to build TFDS with openimages dataset
There is a subset of openimages (face). See: https://docs.blue-oil.org/tutorial/image_det.html#preparation
I used this dataset and run build_tfds command to build TFDS, but I got the following error.

```
    'BBox coordinates should be between 0 and 1. Got {}.'.format(bbox))
ValueError: BBox coordinates should be between 0 and 1. Got BBox(ymin=0.0, xmin=0.108385, ymax=0.885862, xmax=1.0000000000000002).
```

hadusam:
Fix #1208

### DIFF
--- a/blueoil/utils/tfds_builders/object_detection.py
+++ b/blueoil/utils/tfds_builders/object_detection.py
@@ -77,8 +77,8 @@ class ObjectDetectionBuilder(tfds.core.GeneratorBasedBuilder):
                     "bbox": tfds.features.BBox(
                         ymin / height,
                         xmin / width,
-                        (ymin + h) / height,
-                        (xmin + w) / width,
+                        min((ymin + h) / height, 1.0),
+                        min((xmin + w) / width, 1.0),
                     )
                 }
                 for xmin, ymin, w, h, label in annotations


blueoil/utils/tfds_builders/object_detection.py
```diff
@@ -77,8 +81,8 @@ def _generate_examples(self, dataset):
                     "bbox": tfds.features.BBox(
                         ymin / height,
                         xmin / width,
-                        (ymin + h) / height,
-                        (xmin + w) / width,
+                        _clip_max((ymin + h) / height),
```
iizukak, line 84: @hadusam 
`min((ymin + h) / height, 1.0)` is not enough for you?

hadusam: @iizukak 
Yeah, it's enough! Thanks!!!
Fixed.

iizukak (approved):
OA

tfujiwar (approved):
RA

PR #1212, opened 2020-09-16
Adding `--gpu={gpu index}` flag to docker option in tutorials.

ananno:
Adding gpu flag in docs for docker runtime to use GPU. Without this the instructions in tutorials are incomplete and will not use GPU.


iizukak:
@ananno 
Sorry, I'm not understand perfectly. `CUDA_VISIBLE_DEVICES` is not enough for now?

ananno:
> @ananno
> Sorry, I don't understand perfect. `CUDA_VISIBLE_DEVICES` is not enough for now?
> If `CUDA_VISIBLE_DEVICES` is not working and `--gpus` is available, please remove description about `CUDA_VISIBLE_DEVICES`.

The `CUDA_VISIBLE_DEVICES` is an **environment** variable to restrict TensorFlow to use particular GPU device(s). But `--gpu` flag is needed for **docker** to be use particular GPU device(s). But only `CUDA_VISIBLE_DEVICES` is not enough and docker does not initiate **CUDA**.

I've tested with different combination of these two parameter. The filter first use `CUDA_VISIBLE_DEVICES` to select allowed devices. Then docker will select the GPU(s) provided by `--gpu=index(s)`. Check the following combinations of GPU index used in `lmsrv05` with 2 `TitanX` gpus:

CUDA_VISIBLE_DEVICES | --gpus | Training on GPU
--|--|--
0 | 0 | Success
1 | 0 | Success
0 | 1 | Failed
1 | 1 | Failed

So, it seems like the GPU list is first filtered by `CUDA_VISIBLE_DEVICES` than docker use GPU with provided index from that filtered list.

suttang:
@hadusam As I am unable to judge the validity of this content, could you please confirm it when you have time?

joelN123:
oh, it's interesting. So, it seems like `--gpus=0` and not providing `CUDA_VISIBLE_DEVICES` will mean that gpu1 will be left free? So, I think that's what Iizuka-san means, that the `CUDA_VISIBLE_DEVICES` option is actually not necessary.
p.s. the `--gpus` flag is the newer version of `--runtime=nvidia` maybe?

iizukak:
@ananno 
I'm confusing...
Why `Training on GPU` successed with `--gpus=0`? `0` dose not mean don't use GPU?
And why failed with `--gpus=1`?

ananno:
> oh, it's interesting. So, it seems like `--gpus=0` and not providing `CUDA_VISIBLE_DEVICES` will mean that gpu1 will be left free? So, I think that's what @iizukak means, that maybe the `CUDA_VISIBLE_DEVICES` option is actually not necessary?
> p.s. the `--gpus` flag is the newer version of `--runtime=nvidia` maybe?
> edit: oh, maybe `--gpus device=0,1` is more similar to the `CUDA_VISIBLE_DEVICES="0,1"` usage

@joelN123 San:
I've tested `--gpus=0` means all available GPUs. In this test I didn't use `CUDA_VISIBLE_DEVICES` which by default exposes all available GPUs to tensorflow. And with `--gpus=0` docker will run training with all available GPUs. The effect is same as not using `CUDA_VISIBLE_DEVICES`.

And to answer to your question, If you do not use `CUDA_VISIBLE_DEVICES` in a multi-gpu server it will create tensorflow process on all the GPUs with ~150MB gpu memory. To avoid this and restrict the tensorflow process attach to one or specific gpu you have to use `CUDA_VISIBLE_DEVICES`.

joelN123:
understood ! thank you. In that case, it might be good to have `--gpus device=<gpu indices>` in the documentation, and then can remove the `CUDA_VISIBLE_DEVICES` environment variable from the documentation? Maybe it's down to preference. Anyway, I think that is what I will be doing from now on.

ananno:
> understood ! thank you. In that case, it might be good to have `--gpus device=<gpu indices>` in the documentation, and then can remove the `CUDA_VISIBLE_DEVICES` environment variable from the documentation? Maybe it's down to preference. Anyway, I think that is what I will be doing from now on, when running docker on gpu.

@joelN123 San:
I'm doing it also. I agree to remove `CUDA_VISIBLE_DEVICES` all together and use `--gpus device=<gpu indices>` option.

docs/tutorial/image_cls.md
```diff
@@ -106,15 +106,18 @@ If configuration finishes, the configuration file is generated in the `cifar10_t
...
         -v $(pwd)/saved:/home/blueoil/saved \
         blueoil_$(id -un):{TAG} \
         blueoil train -c config/cifar10_test.py
 
 Just like init, set the value of `{TAG}` to the value obtained by `docker images`.
-Change the value of `CUDA_VISIBLE_DEVICES` according to your environment.
+Change the value of `--gpu device=<gpu indices>` according to your environment. 
+For detail check the official [documentation](https://docs.docker.com/config/containers/resource_constraints/#gpu).
+> For old docker versions use `--runtime=nvidia` with `-e CUDA_VISIBLE_DEVICES=<gpu indices>` environment variable
+> instead of `--gpu device=<gpu indices>`.
```
hadusam, line 120: Thanks, but I think we don't need these lines because our supported docker version is `Docker >= 19.03`.
https://docs.blue-oil.org/install/install.html#prerequisites

docs/tutorial/image_cls.md
```diff
@@ -106,15 +106,18 @@ If configuration finishes, the configuration file is generated in the `cifar10_t
...
+        --gpus device=0 \
         -v $(pwd)/cifar:/home/blueoil/cifar \
         -v $(pwd)/config:/home/blueoil/config \
         -v $(pwd)/saved:/home/blueoil/saved \
         blueoil_$(id -un):{TAG} \
         blueoil train -c config/cifar10_test.py
 
 Just like init, set the value of `{TAG}` to the value obtained by `docker images`.
-Change the value of `CUDA_VISIBLE_DEVICES` according to your environment.
+Change the value of `--gpu device=<gpu indices>` according to your environment.
```
hadusam, line 117: there are some typo: `--gpu` -> `--gpus`, please check them. 🙇

hadusam, line 117: there are some typo: `--gpu` -> `--gpus`, please check them. 🙇

docs/tutorial/image_cls.md
```diff
@@ -106,15 +106,18 @@ If configuration finishes, the configuration file is generated in the `cifar10_t
 Train your model by running `blueoil train` with model configuration.
 
     $ docker run --rm \
-        -e CUDA_VISIBLE_DEVICES=0 \
+        --gpus device=0 \
```
hadusam, line 109: In my environment (Docker v19.03.7), this doesn't work. 
I think we need to use `--gpus '"device=0"'`. It's not beautiful though... 😭 
See: https://github.com/NVIDIA/nvidia-docker/issues/1257

hadusam (commented):
@ananno 
Thanks!
It looks good for me to use `--gpus` with `device=` instead of `CUDA_VISIBVLE_DEVICES`.
I left some comments, please check them!

docs/tutorial/image_cls.md
ananno: OK. I'll remove it. Thanks for the review.

docs/tutorial/image_cls.md
ananno: It works without the `'` or `"` quotation mark. I've checked. Should I include them anyway?

docs/tutorial/image_cls.md
ananno: Done

ananno: Done

docs/tutorial/image_cls.md
ananno: Done

hadusam (approved):
@ananno 
LGTM!

Could you please fix the title of this PR with the correct option? Because the title will be used to commit message of this PR.
Thanks for the nice PR!

iizukak (approved):
@ananno @joelN123 
Thanks. I understand just now.
OA

PR #1217, opened 2020-09-25
Consider that 'Reshape' won't preserve quantization

### LINKED ISSUE #1216: Failed to convert when reshape quantized tensor
Since `pass_propagate_format` doesn't consider `Reshape` operator, convert will fail.

Insert `Cast` operator before `Reshape` operator when input is quantized.

primenumber:
Fix #1216 .


iizukak (approved):
OA

a-hanamoto (approved):
RA

PR #1220, opened 2020-09-28
Replace EasyDict to SmartDict

iizukak:
Solve #1092 

Odoku-san implemnetd EasyDict. We can replace EasyDict to SmartDict.


blueoil/utils/config.py
```diff
@@ -133,19 +133,19 @@ def _load_py(config_file):
...
     # use only upper key.
-    return EasyDict({
+    return SmartDict({
         key: value
         for key, value in config.items()
         if key.isupper()
     })
 
 
 def _easy_dict_to_dict(config):
```
oatawa1, line 143: How about changing the function name?
```suggestion
def _smart_dict_to_dict(config):
```

oatawa1 (commented):
I leave one comment.

blueoil/utils/config.py
iizukak: @oatawa1 Thanks. Changed.

oatawa1 (approved):
LGTM. 

Such a big PR. Thank you.

tfujiwar (approved):
RA

Issue #1224, opened 2020-10-15
Stop to save config.yaml


iizukak:
Currently, we are saving `config.yaml` the contents is same as Python config.
IMO, nobody are using `config.yaml`.

https://github.com/blue-oil/blueoil/blob/master/blueoil/utils/config.py#L214-L236

oatawa1:
I think we may still need `config.yaml`.

To generate config.py, it duplicated the input python config file, while `config.yaml` is generated from the config dictionary.

It's possible that when we overwrite some variables, the `config.py` will not get update, but `config.yaml` will give us the correct configuration variables.

joelN123:
That's a good point, thank you for bringing it up.
It looks like there are two places where `config.yaml` file is saved
https://github.com/blue-oil/blueoil/blob/002f5408d11fd2a93457c5e7bc8dc876ace13ce0/blueoil/cmd/train.py#L299
and
https://github.com/blue-oil/blueoil/blob/002f5408d11fd2a93457c5e7bc8dc876ace13ce0/blueoil/cmd/export.py#L146
In `train.py`, it happens immediately after loading the config file. So I think there is no problem to not save the `config.yaml` there.
In `export.py`, it loads the config file, then overrides some things like `BATCH_SIZE` and `PRE_PROCESSOR`, before it then saves as `config.yaml`. But, after that, is the `config.yaml` used for anything? I understand that `meta.yaml` could be used for application such as demo. But I'm not sure about `config.yaml`.
The problem with not removing the `config.yaml` is that currently it can cause error when loading https://github.com/blue-oil/blueoil/pull/1110

PR #1225, opened 2020-10-16
Remove config yaml

joelN123:
The `config.yaml` file is not a useful part of the pipeline, as far as I know. There is already the `config.py` file, which contains the same information. So, code for saving and loading a `config.yaml` is not necessary. This is stated in https://github.com/blue-oil/blueoil/issues/1224. A second point is that loading the `config.yaml` can cause error https://github.com/blue-oil/blueoil/pull/1110. So this PR would also resolve that problem.

After modifying the code, I tried training, exporting, and converting. All seemed to run fine. No `config.yaml` appears.

p.s. in the `save_yaml` function, I just remove the `config.yaml` part, not the `meta.yaml` part.

### DIFF
--- a/blueoil/utils/config.py
+++ b/blueoil/utils/config.py
@@ -19,7 +19,6 @@ from abc import ABCMeta
 
 import yaml
 from blueoil.utils.smartdict import SmartDict
-from yaml.representer import Representer
 
 from blueoil.data_processor import Processor, Sequence
 from blueoil import environment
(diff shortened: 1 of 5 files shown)


tests/unit/test_configs.py
```diff
@@ -19,7 +19,7 @@
 import pytest
 
 from blueoil import environment
```
kchygoe, line 21: No longer required

blueoil/utils/config.py
```diff
@@ -19,7 +19,6 @@
```
kchygoe, line 19: Remove this

kchygoe (changes requested):
plz update

kchygoe (approved):
LGTM for the changes
(plz check the specification change with the owner)

joelN123:
@iizukak how do you think about these changes?
One possible contentious point is that the `save_yaml` function remains, for saving the `meta.yaml`. I think that meta file may be important for running certain kinds of demo https://github.com/blue-oil/blueoil/blob/master/docs/usage/export.md so it's better to keep it.

PR #1228, opened 2020-10-22
Fix PascalVOC dataset loader to accept float bndbox

### LINKED ISSUE #1226: PascalVOC dataset loader doesn't accept float type bounding box
PascalVOC format data accepts float data type for bounding box parameters, but the PascalVOC dataset loader only allows the int type of parameters. The PascalVOC dataset loader should be able to accept the float type of parameters.

https://github.com/blue-oil/blueoil/blob/002f5408d11fd2a93457c5e7bc8dc876ace13ce0/blueoil/datasets/pascalvoc_base.py#L132-#L135

oatawa1:
- Fix PascalVOC dataset loader to accept float bounding box
- Modify PascalVOC dataset unit test to coverage floating point and other class (not just person class)
Closes #1226

### DIFF
--- a/blueoil/datasets/pascalvoc_base.py
+++ b/blueoil/datasets/pascalvoc_base.py
@@ -129,10 +129,10 @@ class PascalvocBase(ObjectDetectionBase):
             for e in list(obj):
                 if e.tag == "bndbox":
                     # subtract 1 to make boxes 0-based coordinate.
-                    ymin = int(e.find("ymin").text) - 1
-                    xmin = int(e.find("xmin").text) - 1
-                    ymax = int(e.find("ymax").text) - 1
-                    xmax = int(e.find("xmax").text) - 1
+                    ymin = float(e.find("ymin").text) - 1
+                    xmin = float(e.find("xmin").text) - 1
+                    ymax = float(e.find("ymax").text) - 1
+                    xmax = float(e.find("xmax").text) - 1
 
                     w = xmax - xmin
                     h = ymax - ymin
--- a/tests/unit/datasets_tests/test_pascalvoc.py
+++ b/tests/unit/datasets_tests/test_pascalvoc.py
@@ -29,7 +29,7 @@ def test_pascalvoc_2007():
     assert dataset.available_subsets == ['train', 'validation', 'test', 'train_validation']
     assert dataset._files_and_annotations() == (
         ['unit/fixtures/datasets/PASCALVOC_2007/VOCdevkit/VOC2007/JPEGImages/000085.jpg'],
-        [[[21, 69, 232, 257, 14], [337, 11, 162, 187, 14]]]
+        [[[5, 103, 212, 174, 8], [21, 69, 232, 257, 14]]]
     )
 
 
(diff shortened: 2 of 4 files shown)


iizukak (approved):
OA

oatawa1:
@iizukak Thank you.

blueoil/datasets/pascalvoc_base.py
```diff
@@ -129,10 +129,10 @@ def _get_boxes_from_annotation(self, xml_file):
...
                 if e.tag == "bndbox":
                     # subtract 1 to make boxes 0-based coordinate.
-                    ymin = int(e.find("ymin").text) - 1
-                    xmin = int(e.find("xmin").text) - 1
-                    ymax = int(e.find("ymax").text) - 1
-                    xmax = int(e.find("xmax").text) - 1
+                    ymin = float(e.find("ymin").text) - 1
+                    xmin = float(e.find("xmin").text) - 1
+                    ymax = float(e.find("ymax").text) - 1
+                    xmax = float(e.find("xmax").text) - 1
```
tfujiwar, line 135: I'm not sure if the coordinates are 1-based also for float values. I feel it's more natural to use 0-based coordinates.

Do we really need to support float values? Is the tool that generates PascalVOC-formatted data aware of 1-based coordinates?

tfujiwar (commented):
Readability is good but I have a question about the functionality. Please discuss with the code owner and the issue author.

blueoil/datasets/pascalvoc_base.py
oatawa1: > I'm not sure if the coordinates are 1-based also for float values. I feel it's more natural to use 0-based coordinates in that case.

This should be the same as an Integer, as this is a standard for PascalVOC annotation.
I don't have much knowledge about this annotation.
But I think PascalVOC format annotation is  1-based float values. 
Because this PascalVOC dataset loader code in Chainer also process data as 1-based for float values 
https://github.com/chainer/chainercv/blob/7159616642e0be7c5b3ef380b848e16b7e99355b/chainercv/datasets/voc/voc_bbox_dataset.py#L105
In this case, it's better to say that we supported 1-based coordinates for PascalVOC.

> Do we really need to support float values? 

Yes, there is no exact evidence show that the bndbox can be float though, in many PascalVOC document the bndbox value can be float.
- http://host.robots.ox.ac.uk/pascal/VOC/voc2007/guidelines.html (The official website is currently down...)
- https://pjreddie.com/media/files/VOC2007_doc.pdf (Page 13)

> Is the tool that generates PascalVOC-formatted data aware of 1-based coordinates?

Yes, it seems possible to set the offset of the bndbox.

lm-lily (approved):
LGTM

blueoil/datasets/pascalvoc_base.py
tfujiwar: Thanks for the evidence. I've understood.

tfujiwar (approved):
RA

oatawa1:
@lm-lily Thank you.

@tfujiwar Thank you

PR #1229, opened 2020-10-28
Improve performance of dataset_itertator in prepetch mode

lm-takumi-nakaso:
Improve performance of dataset_itertator in prepetch mode

Change the number of threads from fixed 8 to cpu cores.
Share a single process pool with multiple dataset_iterators.
Use a single thread per process.

Use [loky](https://loky.readthedocs.io/en/stable/) to achieve this.


iizukak:
@lm-takumi-nakaso Thank you for the contribution!!
Could you add the speed comparison before and after of this PR?

PR #1235, opened 2021-02-08
Fix sign function to return only 1 and -1

### LINKED ISSUE #1201: use modified tf.sign that can only output two possible values (for binary quantizer)
At the high-level code for binary quantizer, `tf.sign` is used to binarize the weights (in the [binary.py](https://github.com/blue-oil/blueoil/blob/b647203f5249a1b026760edb5cc17307f29589c5/blueoil/quantizations/binary.py#L163) quantization definition). However, if given input of `0.0`, the output of TensorFlow sign function is `0.0`. So there is some very small chance of getting a third output value (not +/- 1). Although it doesn't practically cause problems now, it could be good to change this.

### detail

In the lower-level code there are only two possible weight values (it's a single bit), so a value of `0.0` would get converted to one of the other values. Maybe by [binarizer](https://github.com/blue-oil/blueoil/blob/b647203f5249a1b026760edb5cc17307f29589c5/blueoil/converter/core/operators.py#L762) method? Anyway, the point is the slight difference between higher-level code and lower-level code.

In practice, this is not a problem because of almost no `0.0` weight values. Below I plot the count of kernel vars equal to `0.0`, over a few different layers. It looks rare at initialization (only happened once in this example network), and after that, it becomes even more rare (essentially never happens).

image

However, there are a couple of reasons why it could be good to strictly enforce only two possible quantized weight values at the higher-level (TensorFlow) code
1. There might possibly be some unusual network or situation where zero-valued weights have non-negligible likelihood, then the inference behaviour of the model would be different between non-converted and converted code.
2. When someone new to the code reads `tf.sign`, it might be confusing to them, because it has 3 possible output values, strictly speaking.

### possible solution

One way is to modify the forward function, so that `0.0` gets mapped to `-1.0` (or, whichever way it is in the lower level code). In this way, the quantized weight values at the Tensorflow level could only have two possible values.

Still, even after making this change, there could be a problem in an unusual situation where some of the unquantized weights have value `0.0`, because in that case the binary quantizer is not perfectly symmetric. But I think that is a better situation to be in.

Maybe something simple like changing [binary quantizer line 163](https://github.com/blue-oil/blueoil/blob/b647203f5249a1b026760edb5cc17307f29589c5/blueoil/quantizations/binary.py#L163) from
`return tf.sign(x) * expectation`
to
```
not_zero = x - tf.cast(tf.equal(x, 0.0), tf.float32)
return tf.sign(not_zero) * expectation
```
could implement this solution. (and keeping the definition of gradient same as before).

### LINKED ISSUE #1234: BinaryMeanScalingQuantizer could return zero wrongly
https://github.com/blue-oil/blueoil/blob/fa870947c7c3705c378307d1266612821267d933/blueoil/quantizations/binary.py#L163

`tf.sign(x)` returns zero when x is zero.

But we assumes 1 or -1, it can cause inference result mismatch.

oatawa1:
We assume the value from quantization can be only `1` or `-1`. If not, it can cause inference result mismatch.
This PR implemented a custom sign function. The only difference from `tf.sign` and `np.sign` is `sign(0) = 1` instead of 0. 

Closes #1201 #1234

### DIFF
--- a/blueoil/converter/core/operators.py
+++ b/blueoil/converter/core/operators.py
@@ -743,14 +743,14 @@ class BinaryMeanScalingQuantizer(Quantizer):
     def run_forward(self) -> np.ndarray:
         in_data = self.input_ops['input'].data
         self._scaling_factor = np.mean(np.abs(in_data))
-        self._data = np.sign(in_data)
+        self._data = np.where((in_data >=0), 1, -1)
 
         return self._data * self._scaling_factor
 
     def run_forward_no_scaling_factor(self) -> np.ndarray:
         in_data = self.input_ops['input'].data
         self._scaling_factor = np.mean(np.abs(in_data))
-        self._data = np.sign(in_data)
+        self._data = np.where((in_data >=0), 1, -1)
 
         return self._data
 
@@ -2442,7 +2442,7 @@ class BinaryChannelWiseMeanScalingQuantizer(Quantizer):
     def run_forward(self) -> np.ndarray:
         in_data = self.input_ops['input'].data
         self._scaling_factor = np.mean(np.abs(in_data), axis=(1, 2, 3)).astype(np.float32)
-        self._data = np.sign(in_data)
+        self._data = np.where((in_data >=0), 1, -1)
 
         scaling = copy.deepcopy(self._scaling_factor)
         extra_dims = tuple(np.ones((len(self._data.shape) - len(scaling.shape)), dtype=np.int32))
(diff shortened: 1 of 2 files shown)


iizukak (approved):
OA

oatawa1:
@iizukak Thanks.

tfujiwar (approved):
RA

oatawa1:
@tfujiwar Thanks.

Issue #1238, opened 2021-03-02
install.md need to be updated


MasaYan24:
There is a section explaining how to expand root partition with a tool.
But the explanation is old.
It is needed to be updated.

https://docs.blue-oil.org/install/install.html
section 1.3.2

https://leapmind.kibe.la/notes/6637#comment_3532

PR #1239, opened 2021-03-02
fix install.md

MasaYan24:
The section explaining expanding root partition needs to be updated as presented here.

#1238


iizukak (approved):
OA

